PR #7, opened 2022-01-04
Added mintTo and emailTo parameters to support minting to external wallets

angel-paella:


alfonso-paella:
@mPaella should review this one

alfonso-paella:
Also we should add validations (maybe in a subsequent PR):
* That emailTo is an email address
* That mintTo is a valid solana address
* That if emailTo is passed, mintTo is passed too, and viceversa

angel-paella:
> Also we should add validations (maybe in a subsequent PR):
> 
> * That emailTo is an email address
> * That mintTo is a valid solana address
> * That if emailTo is passed, mintTo is passed too, and viceversa

Makes sense, I'll add it as a git task for a follow-up

mPaella (approved):

Issue #14, opened 2022-01-18
[Optimization] Do quick client side validation of clientId formats


alfonso-paella:
Once we've settled on a format, we can quickly check and error out if a clientId is malformed, as opposed to having to do a network request to check

pgarciaegido:
@alfonso-paella how does the client ID look like?

PR #15, opened 2022-01-19
Stop making API requests with unset client ID

alfonso-paella:
Addresses #14 

We're getting hammered with requests from folks integrating crossmint who haven't yet added their own client ID

This just checks for the most common cases where it's not yet set, and throws a warning instead of trying to call the API

### DIFF
--- a/packages/ui/react-ui/src/CrossMintStatusProvider.tsx
+++ b/packages/ui/react-ui/src/CrossMintStatusProvider.tsx
@@ -19,6 +19,11 @@ export const CrossMintStatusProvider: FC<CrossMintStatusProviderProps> = ({
     );
 
     async function fetchClientIntegration() {
+        if (!clientId || clientId === "" || clientId === "<YOUR_CLIENT_ID>") {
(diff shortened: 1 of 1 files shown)


packages/ui/react-ui/src/CrossMintStatusProvider.tsx
```diff
@@ -19,6 +19,11 @@ export const CrossMintStatusProvider: FC<CrossMintStatusProviderProps> = ({
     );
 
     async function fetchClientIntegration() {
+        if (!clientId || clientId === "" || clientId === "<YOUR_CLIENT_ID>") {
```
angel-paella, line 22: !clientId already covers the ```clientId === ""``` use case

angel-paella (approved):
Nice 😂😂

Soon enough we should probably stop getting alarms on client errors or we'll stop believing (and therefore checking) the client alarms

packages/ui/react-ui/src/CrossMintStatusProvider.tsx
alfonso-paella: Crap, I missed this comment. I'll fix on a new PR

PR #38, opened 2022-02-20
#14 Client side validation of client ID

pgarciaegido:


packages/ui/react-ui/src/hooks/useCrossMintStatus.ts
```diff
@@ -19,14 +19,24 @@ interface IProps {
...
     async function fetchClientIntegration() {
         if (!clientId || clientId === "" || clientId === "<YOUR_CLIENT_ID>") {
-            console.warn("You must enter your own CrossMint client ID in <CrossMintProvider clientId=XXX>");
+            console.error("You must enter your own Crossmint client ID in <CrossMintButton clientId=XXX>");
+            return;
+        }
+
+        if (!isValidUUID(clientId)) {
+            console.error(
+                "Entered `clientId` is not valid. It must be a valid Crossmint client ID. It will have the following format: `00000000-0000-0000-0000-00000000`."
```
pgarciaegido, line 38: Thoughts on the error message? Do you think we should be even more explicit? Maybe with a code snippet of the button?

packages/ui/react-ui/src/hooks/useCrossMintStatus.ts
```diff
@@ -19,14 +19,24 @@ interface IProps {
...
+
 export default function useCrossMintStatus({ clientId }: IProps) {
     const [status, setStatus] = useState<OnboardingRequestStatusResponse>(
         OnboardingRequestStatusResponse.WAITING_SUBMISSION
     );
 
     async function fetchClientIntegration() {
         if (!clientId || clientId === "" || clientId === "<YOUR_CLIENT_ID>") {
-            console.warn("You must enter your own CrossMint client ID in <CrossMintProvider clientId=XXX>");
+            console.error("You must enter your own Crossmint client ID in <CrossMintButton clientId=XXX>");
```
pgarciaegido, line 32: Replaced `console.warn` for `console.error`

alfonso-paella:
So sorry, I should have shared this code snippet before: https://github.com/Paella-Labs/crossbit-main/blob/c95582013e301faec370e9c45c6fffe7481f2cce/storage/ClientIntegrationRepository.ts#L93

That's the right client id validation

packages/ui/react-ui/src/hooks/useCrossMintStatus.ts
alfonso-paella: For now I'd say "The clientId passed to <CrossMintButton> is invalid. Make sure to pass the clientId obtained from the crossmint team, with format `XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXX`"

I think we can improve this later to pointing to a URL within our documentation site that explains how client IDs work. But we don't have a great place yet

alfonso-paella (changes requested):

alfonso-paella (approved):

PR #41, opened 2022-02-21
Take back Crossmint Button Status

pgarciaegido:


packages/ui/react-ui/src/CrossMintButtonStatus.tsx
```diff
@@ -0,0 +1,79 @@
...
+import useCrossMintStatus, { OnboardingRequestStatusResponse } from "./hooks/useCrossMintStatus";
+import { Button, Img, Paragraph } from "./styles/index";
+
+export interface StatusButtonProps {
+    className?: string;
+    disabled?: boolean;
+    onClick?: (e: MouseEvent<HTMLButtonElement>) => void;
+    style?: CSSProperties;
+    tabIndex?: number;
+    clientId: string;
```
pgarciaegido, line 11: Same as with `CrossmintButton`, now both `clientId` and `auctionId` props need to be provided in the button.

packages/ui/react-ui/src/CrossMintButtonStatus.tsx
```diff
@@ -0,0 +1,79 @@
...
+}) => {
+    const status = useCrossMintStatus({ clientId });
+
+    const handleClick: MouseEventHandler<HTMLButtonElement> = useCallback(
+        (event) => {
+            if (onClick) onClick(event);
+
+            if (status === OnboardingRequestStatusResponse.WAITING_SUBMISSION) {
+                window.open(
+                    `https://crossmint.io/developers/onboarding${clientId ? `?clientId=${clientId}` : ""}${
```
alfonso-paella, line 35: Now that we're on it can we change this url to start with www. ?

crossmint.io redirects to www.crossmint.io

packages/ui/react-ui/src/CrossMintButtonStatus.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { Button, Img, Paragraph } from "./styles/index";
+
+export interface StatusButtonProps {
+    className?: string;
+    disabled?: boolean;
+    onClick?: (e: MouseEvent<HTMLButtonElement>) => void;
+    style?: CSSProperties;
+    tabIndex?: number;
+    clientId: string;
+    auctionId: string;
```
alfonso-paella, line 12: Auction ID should be optional, and only passed back to crossmint API if set

alfonso-paella (changes requested):

packages/ui/react-ui/src/CrossMintButtonStatus.tsx
pgarciaegido: How about I change them all in https://github.com/CrossMint/crossmint-client-sdk/pull/43 since Im using an enum there?

alfonso-paella: ok

mPaella (approved):

PR #43, opened 2022-02-21
Feature/#18 add headers in requests

pgarciaegido:
Adds headers in the `/status` request, and query params `clientName` and `clientVersion` to pass to the checkout url.

### DIFF
--- a/packages/ui/react-ui/package.json
+++ b/packages/ui/react-ui/package.json
@@ -24,6 +24,7 @@
         "access": "public"
     },
     "scripts": {
+        "prebuild": "node -p \"'export const LIB_VERSION = ' + JSON.stringify(require('./package.json').version) + ';'\" > src/version.ts",
         "clean": "shx rm -rf lib/*",
         "build": "yarn clean && tsc -p tsconfig.json && tsc -p tsconfig-cjs.json",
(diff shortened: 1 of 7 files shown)


packages/ui/react-ui/package.json
```diff
@@ -24,6 +24,7 @@
         "access": "public"
     },
     "scripts": {
+        "prebuild": "node -p \"'export const LIB_VERSION = ' + JSON.stringify(require('./package.json').version) + ';'\" > src/version.ts",
```
pgarciaegido, line 27: creates a `version.ts` file where to get the package.json version

alfonso-paella (approved):
LGTM. 

One thing I dont understand is how does the version number update with lerna. Make sure to test that that process doesn't break what you put together here

pgarciaegido:
@alfonso-paella you mean with the `publish` command? I guess that when you bump a package version, you would need to build and then publish. If the `postbuild` hook works as expected on the release process, I would assume that the `prebuild` would work as well.

How do you normally handle the release process? Maybe @mPaella knows too.

Thanks!

alfonso-paella (approved):
Thanks for addressing that

PR #44, opened 2022-02-22
Various fixes

pgarciaegido:
In this PR I tackle multiple problems:

1. Remove `@solana/web3.js` dependency, since it seems to be thought for server enviroments (it was requiring `fs` for example)
2. Replace `styled-components` for `jss`. For some reason, Lerna wasn't happy with tagged templates... I worked in every other enviroment (referencing the package locally, using a npm link outside the monorepo...) but for some reason linking with lerna wasn't working. But thats alright, with jss we have exactly the same deal.
3. Move `types.js` to inside of the react-ui repo, since the output from typescript wasn't looking good. We need to rething where to place all these common files/utilities, but does not need to be right now, we can start thinking about it when we start creating the buttons adaptors for different frameworks.


pgarciaegido:
Unfortunately I will need to revisit the usage of styled-components. Something is not working correctly. Will do that tomorrow

packages/ui/react-ui/src/hooks/useCrossMintStatus.ts
```diff
@@ -26,7 +25,7 @@ interface IProps {
 
 const validateClientId = (clientId: string): boolean => {
     try {
-        const isValid = validate(clientId) || new PublicKey(clientId);
+        const isValid = validate(clientId);
```
pgarciaegido, line 28: Since this is just a check for saving calls to the status endpoint, do you guys think we can live without the `PublicKey` from `@solana/web3.js`? @mPaella

alfonso-paella: Well the thing is if we remove it, we'll fail to send valid IDs (in the form of public keys) to the server

alfonso-paella: Actually, it should be fine: 

all the clients who used the old "public key" format for client IDs don't need to check the onboarding status. Only the new ones do. So I think you can remove this safely

pgarciaegido:
All problems have been tackled (see description). Now you can easily link and use starters for testing the button.

mPaella:
LGTM

mPaella (approved):

packages/ui/react-ui/src/hooks/useCrossMintStatus.ts
pgarciaegido: @mPaella will do that 👍

PR #46, opened 2022-02-23
Ensure that our buttons only render client-side

pgarciaegido:
Hey,

So this fixes a funny situation happening in next apps:

- Next can render both server side and client side
- CSS-in-JS libraries would create random classnames in order to attach the styling provided.
- There are times where dom is rendered on the server-side, and then re-hydrated in the client. That confuses  `JSS`, since it need to generate the class-name twice. That results into wrong styling.

This is a bit of a hacky solution to make sure that our buttons only render on client side, regarding of wether they are running on a react or a nextjs app.

I could create a parent component that would integrate the hook, something like:

```
<OnlyRenderOnClientSide>
  <button {props}>
     .....
</OnlyRenderOnClientSide>
```

Whatever you guys prefer. I think this solution is simple enough.

Thanks!


alfonso-paella:
Did you consider something like this? https://stackoverflow.com/questions/49411796/how-do-i-detect-whether-i-am-on-server-on-client-in-next-js

Seems simpler

pgarciaegido:
Your solution seems to be working alright @alfonso-paella !! Thanks!

packages/ui/react-ui/src/CrossMintButtonStatus.tsx
```diff
@@ -61,21 +62,27 @@ export const CrossMintStatusButton: FC<StatusButtonProps> = ({
         }
     }, [status]);
 
+    const isClientSide = getIsClientSide();
```
alfonso-paella, line 65: I am not 100% confident, but I think you can simplify further and 

1. Rename the exported property from getIsClientSide to isClientSide
2. export the boolean directly and not a function, as it will always return the same value given an operating environment
3. Remove this intermediate variable

pgarciaegido: yep, it works! I hesitated a bit at first, but it definitely makes sense that it does: It will instanciate the utils file in the server, and then in the client again.

alfonso-paella (approved):

PR #49, opened 2022-02-24
Feature/#33 rename button

pgarciaegido:
Renamed:

`CrossMintButton` to `CrossmintPayButton`
`CrossMintButtonStatus` to `CrossmintButtonStatus`

I think I have changed them everywhere. Also changed the migration google docs: https://docs.google.com/document/d/14IKpjrij7kU7Dr0I7rZkf0PyDNbXiklx2v4GuzUrFbw/edit?usp=sharing


packages/starter/example/pages/index.tsx
```diff
@@ -11,10 +11,11 @@ export default function Index() {
                 alignItems: "center",
             }}
         >
-            <CrossMintButton
+            <CrossmintPayButton
                 collectionTitle="Test Collection"
                 collectionDescription="This is a CrossMint test collection!"
                 collectionPhoto="https://picsum.photos/300"
+                clientId="058eeac7-3751-4c66-afc3-f77270c9d636"
```
alfonso-paella, line 18: was it intended to add an id here?

alfonso-paella (approved):
approved with a comment

Also -- we shouldnt merge this until we're ready to publish, as devs may be relying on our example code and this is changing it to an unreleased version

packages/starter/example/pages/index.tsx
pgarciaegido: This is what we had in the `<CrossMintProvider>`. I just moved it to the button.

mPaella (approved):

PR #50, opened 2022-02-26
Adding config prop to CrossmintPayButton

pgarciaegido:
Here I add the config type to the CrossmintPayButton, and refactor a bit, incuding interfaces to the `types.ts` file.


packages/ui/react-ui/src/types.ts
```diff
@@ -0,0 +1,52 @@
...
+    clientName = "X-Client-Name",
+}
+
+enum theme {
+    light = "light",
+    dark = "dark",
+}
+
+export enum payButtonTypes {
+    solanaCandyMachine = "solana-candy-machine",
```
mPaella, line 23: In general we should follow [this](https://github.com/Paella-Labs/crossbit-main/blob/dc6f1bf0b25d6a819f542ad630027f49d1be4b70/models/ClientIntegration.ts#L7) enum

packages/ui/react-ui/src/types.ts
```diff
@@ -0,0 +1,52 @@
...
+export interface BaseButtonProps {
+    className?: string;
+    disabled?: boolean;
+    onClick?: (e: MouseEvent<HTMLButtonElement>) => void;
+    style?: CSSProperties;
+    tabIndex?: number;
+    clientId: string;
+    auctionId?: string;
+    theme?: theme;
+    development: boolean;
```
mPaella, line 39: should `development` be optional? default to false

packages/ui/react-ui/src/types.ts
```diff
@@ -0,0 +1,52 @@
...
+export interface CrossmintPayButtonProps extends BaseButtonProps {
+    collectionTitle?: string;
+    collectionDescription?: string;
+    collectionPhoto?: string;
+    mintTo?: string;
+    emailTo?: string;
+    listingId?: string;
+    showOverlay?: boolean;
+    hideMintOnInactiveClient?: boolean;
+    config?: PayButtonConfig;
```
mPaella, line 51: Maybe a different name to be more specific that this is related to their blockchain contract? `mintConfig` or something similar

pgarciaegido: That makes sense. wdyt @alfonso-paella ?

alfonso-paella: Sure

packages/ui/react-ui/src/types.ts
```diff
@@ -0,0 +1,54 @@
...
+    dark = "dark",
+}
+
+export enum mintingContractTypes {
+    CANDY_MACHINE = "candy-machine",
+    SOLANA_AUCTION = "solana-auction",
+    ERC_721 = "erc-721",
+}
+
+export interface PayButtonConfig {
```
alfonso-paella, line 28: On discord we decided to not have any client side validation / enforcement of what goes on the mintConfig property, so we can release new contracts without doing a new SDK release each time

As a result,, this type is not needed and should be replaced with an "any"

alfonso-paella (changes requested):

alfonso-paella (approved):

PR #64, opened 2022-03-04
Feature/create agnostic services

pgarciaegido:
The main idea of this PR is to start taking out logics from React to agnostic services that could be used regarding of the framework in use.

In this PR I have created three services:
- `crossmintModalService`. This replaces the existing hook `useCrossmintModal`.
- `crossmintPayButtonService`. This would take care of all the logics that happen on the `CrossmintPayButton` view.
- `crossmintStatusService`. It would take care of the logics regarding fetching the status and handling the button status view.

This was actually pretty straight forward since our components are not really too heavy on reactivity, so it is fairly easy to take the logics out.

I would like to get an initial feedback on this approach, and if I get the green light Ill move all the logics into common services and start creating Vue, Angular and Svelte wrappers this weekend.

Screenshot 2022-03-04 at 18 38 24

Thank you!


packages/ui/react-ui/__tests__/CrossmintPayButton.spec.tsx
```diff
@@ -14,31 +14,83 @@ global.fetch = jest.fn(() => fetchReturns);
 // TODO(#61): make this automatically mocked in every test suite
```
pgarciaegido, line 14: IGNORE THIS FILE.

This has its own PR. https://github.com/Crossmint/crossmint-client-sdk/pull/63

pgarciaegido:
Tests are failing because now its depending on the base package... Will see how to fix later

mPaella:
I don't think that the hooks should go inside the base package? 

If u checkout [the example this repo is based on](https://github.com/solana-labs/wallet-adapter/tree/master/packages/core), they have separate folders for react, and base is a general package containing only non-framework specific types and functions.

pgarciaegido:
There are no hooks inside the base package. 

Maybe I should change the way the services are used. Instead of:

`const { connect } = useCrossmintModalService();`

Im changing it for:

`crossmintModalService.connect()`;

So its clear that there is no reactivity in the base logics, just pure typescript.

packages/ui/react-ui/src/CrossmintPayButton.tsx
```diff
@@ -30,34 +33,34 @@ export const CrossmintPayButton: FC<CrossmintPayButtonProps> = ({
...
-            "No collection description specified. Please add a collection description to your <CrossmintPayButton />"
-        );
-        collectionDescription = "";
-    }
-
-    if (collectionPhoto === "<OPT_URL_TO_PHOTO_COVER>") {
-        console.warn("No collection photo specified. Please add a collection photo to your <CrossmintPayButton />");
-        collectionPhoto = "";
-    }
+    const [newCollectionTitle, newCollectionDescription, newCollectionPhoto] = checkProps({
```
mPaella, line 49:
```
[collectionTitle, collectionDescription, collectionPhoto] = checkProps({
        collectionTitle,
        collectionPhoto,
        collectionDescription,
    });
```

works?

mPaella:
> There are no hooks inside the base package.
> 
> Maybe I should change the way the services are used. Instead of:
> 
> `const { connect } = useCrossmintModalService();`
> 
> Im changing it for:
> 
> `const { connect } = crossmintModalService();`
> 
> So its clear that there is no reactivity in the base logics, just pure typescript. Reactivity should be handled at the component level.
> 
> Im [using functions to encapsulate information](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures) instead of classes (as in the example you provided). Do classes make more sense to you?
> 
> @mPaella

Functions are good. Only other thing I can see is possibly the useCrossmintStatus hook should go inside the base package as well? wdyt?

pgarciaegido:
cc @angel-paella 

@mPaella could you please test locally to ensure that everything is working correctly.

For the tests to work, we would need to publish the new base package. Locally, they work.

Thanks!

mPaella:
LGTM. We should also add the tsup build to the base package, then deploy both.

PR #71, opened 2022-03-09
Fix esm import issues

mPaella:
- Changes source code imports to have file extensions. Ex: `./CrossmintStatusButton` => `./CrossmintStatusButton.js`
- Use tsc-esm for better esm compatibility
- Add type declarations in lib output

Allows for the package to be used in react-scripts v5, and no longer require nextjs users to use `next-transpile-modules`

TODO: Change postbuild script to rename all .js files in /lib/esm to .mjs


packages/ui/react-ui/package.json
```diff
@@ -28,8 +27,8 @@
         "version": "yarn run create-version-file && git add .",
         "prebuild": "yarn run create-version-file",
         "clean": "shx rm -rf lib/*",
-        "build": "yarn clean && tsc -p tsconfig.json && tsc -p tsconfig-cjs.json",
-        "postbuild": "echo '{\"type\":\"commonjs\"}' | npx json > lib/cjs/package.json && echo '{\"type\":\"module\"}' | npx json > lib/esm/package.json",
+        "build": "yarn clean && tsc -p tsconfig.json && tsc-esm -p tsconfig.json && tsc -p tsconfig.cjs.json",
+        "postbuild": "echo '{\"type\":\"commonjs\"}' | npx json > lib/cjs/package.json && echo '{\"type\":\"module\"}' | npx json > lib/esm/package.json && shx mv lib/esm/index.js lib/esm/index.mjs",
```
pgarciaegido, line 31: Do we actually need the `shx mv lib/esm/index.js lib/esm/index.mjs` part? Its failing on my end:

`mv: no such file or directory: lib/esm/index.js`

Looks like index.mjs is already being generated.

pgarciaegido:
Tests:

React env with react-scripts v-5: ✅
Next: ✅
React env with react-scripts v-4: 
Screenshot 2022-03-09 at 22 38 11
Do you see this aswell @mPaella ? I have just downgraded `react-scripts` in outr react starter from  `5.0.0` to `4.0.3`.

packages/ui/react-ui/package.json
mPaella: uhh its 100% not being generated on my machine, meaning the mv command is required

PR #73, opened 2022-03-20
Bundling button with tsup

pgarciaegido:
Hey,

So, as an alternative to @mPaella PR https://github.com/Crossmint/crossmint-client-sdk/pull/71 and after banging my head against the wall for a few hours, I thought: why don't we just change the whole approach and deliver an already fully bundled solution to the client?

So, using [tsup](https://github.com/egoist/tsup) for bundling the button creates 2 output files (one .cjs and another .js) and a types declaration file. It also delivers the code minified.

Tested in:
- React scripts 5
- React scripts 4
- Nextjs

@mPaella let me know what you think.

After approving, we could publish an alpha version to make sure everything is still working

Thanks!


alfonso-paella (approved):
Maxwell should review as well if possible! 

But the test plan sounds good

mPaella:
`tsup` should probably be in the package.json at the root, then used in all the packages we deploy

mPaella:
Dopeee working in all starters for me. And works on windows.

PR #90, opened 2022-04-01
New bundle to Vanilla UI button

pgarciaegido:
Now, with this new bundle, Vanilla UI `crossmint-pay-button` will work in ES6 environments, Common JS environments, but also in the browser via CDN! See the new `vanilla-ui-starter/index.html` 😄

### DIFF
--- a/packages/core/base/src/services/crossmintStatusService.ts
+++ b/packages/core/base/src/services/crossmintStatusService.ts
@@ -1,6 +1,6 @@
-import { clientNames, customHeaders, onboardingRequestStatusResponse, OnboardingQueryParams } from "../models/types";
-import { validate } from "uuid";
+import { OnboardingQueryParams, clientNames, customHeaders, onboardingRequestStatusResponse } from "../models/types";
 import { getEnvironmentBaseUrl } from "../utils/ui";
+import validateUUID from "../utils/validateUUID";
 
 interface CrossmintStatusServiceParams {
     libVersion: string;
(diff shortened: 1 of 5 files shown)


packages/core/base/src/services/crossmintStatusService.ts
```diff
@@ -1,6 +1,6 @@
-import { clientNames, customHeaders, onboardingRequestStatusResponse, OnboardingQueryParams } from "../models/types";
-import { validate } from "uuid";
+import { OnboardingQueryParams, clientNames, customHeaders, onboardingRequestStatusResponse } from "../models/types";
 import { getEnvironmentBaseUrl } from "../utils/ui";
+import validateUUID from "../utils/validateUUID";
```
pgarciaegido, line 3: This is actually the very same logics that `uuid` package uses 😄 

See https://github.com/uuidjs/uuid/blob/main/src/validate.js

packages/starter/vanilla-ui-starter/index.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>Crossmint pay button</title>
+    <script src="../../ui/vanilla-ui/lib/index.global.js"></script>
```
alfonso-paella, line 5: If a customer copy pastes the starter this import wont work, no?

Shouldnt we import the package via npm and  import it here somehow differently?

pgarciaegido: Forgot to mention!! Probably here we will have an CDN link like https://unpkg.com/browse/@crossmint/client-sdk-vanilla-ui@0.0.1-alpha.0/lib/CrossmintPayButton.js

But since this is not deployed yet, we cant place it there 😄

alfonso-paella (approved):
test are failing

PR #92, opened 2022-04-04
Enhacement/update docs

pgarciaegido:


mPaella:
We probably should forward people to the docs website, instead of having to maintain this? 

https://docs.crossmint.io/

pgarciaegido:
I think that makes sense, altough is true that its nice when you enter in a repo and you have a good readme that get you on track super fast. Maybe we could simplify it so we have the very basic installation and setup process and then the link to the docs.

alfonso-paella (approved):
Sounds fine but as Maxwell was suggesting, I would put a very visibile link to docs.crossmint.io somewhere

pgarciaegido:
Added a link to docs.crossmint.io on top of the 3 readme.

alfonso-paella (approved):

PR #100, opened 2022-04-29
Pay with ETH

pgarciaegido:
Changes requested in https://docs.google.com/document/d/1vOlaBfIeaPA1TwstwGH4ey9F2bgQ8UT1bjpvrKFSt6g/edit#

### DIFF
--- a/packages/core/base/src/models/types.ts
+++ b/packages/core/base/src/models/types.ts
@@ -16,7 +16,7 @@ export enum customHeaders {
 
 type theme = "light" | "dark";
 
-export type paymentMethods = "ETH";
+export type paymentMethods = "FIAT" | "ETH";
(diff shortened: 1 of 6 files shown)


packages/core/base/src/models/types.ts
```diff
@@ -16,6 +16,9 @@ export enum customHeaders {
 
 type theme = "light" | "dark";
 
+export type paymentMethods = "ETH";
```
alfonso-paella, line 19: I think we should add "fiat" (category for Credit card, debit card, and whatever we add in the future) as an option, and make it the default if omitted

alfonso-paella (approved):
Approved with a comment. In the interest of moving fast, it may be ok if we just copy the comment into a GH issue and do it as a fast follow post launch

PR #103, opened 2022-05-05
Enforce that the button always opens a popup

pgarciaegido:
For some reason, (maybe chrome update?) our button was no longer opening a popup but a tab. 

https://user-images.githubusercontent.com/20989060/166885399-c22423ba-fe11-49f1-8170-ef7f07a33877.mov

Now with this change it does. Tested in Chrome and Firefox.

### DIFF
--- a/packages/core/base/src/services/crossmintModalService.ts
+++ b/packages/core/base/src/services/crossmintModalService.ts
@@ -20,9 +20,9 @@ type MintQueryParams = {
 const overlayId = "__crossmint-overlay__";
 
 function createPopupString() {
-    return `height=750,width=400,left=${window.innerWidth / 2 - 200},top=${
-        window.innerHeight / 2 - 375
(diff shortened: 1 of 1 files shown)


angel-paella (approved):

alfonso-paella:
Have we tested ad and pop-up blockers? 

The prior code had the side effect that whenever there was an ad blocker or pop up blocker installed, it opened in another tab. Wanted to ensure that continues to work like that

pgarciaegido:
@alfonso-paella Great point!!

So testing in my chrome profile with popups disabled, two ad-blockers and an extra extension for blocking popups:

Screenshot 2022-05-06 at 10 04 12

Screenshot 2022-05-06 at 10 15 48

It has always opened correctly.

In any case, Im adding an extra attempt to manually adding a new tab in case popover does not open here: https://github.com/Crossmint/crossmint-client-sdk/pull/104

PR #104, opened 2022-05-06
Extra safety on popup not opening

pgarciaegido:
See https://github.com/Crossmint/crossmint-client-sdk/pull/103#issuecomment-1119380838

### DIFF
--- a/packages/core/base/src/services/crossmintModalService.ts
+++ b/packages/core/base/src/services/crossmintModalService.ts
@@ -118,9 +118,12 @@ export function crossmintModalService({
             if (showOverlay) {
                 addLoadingOverlay();
             }
-        } else {
-            setConnecting(false);
-            console.error("Failed to open popup window");
+            return;
+        }
+
+        const newTab = window.open(url, "_blank");
+        if (!newTab) {
+            console.error("Failed to open popup window and new tab");
         }
     };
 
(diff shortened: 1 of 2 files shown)


packages/core/base/src/services/crossmintModalService.ts
```diff
@@ -118,9 +118,12 @@ export function crossmintModalService({
             if (showOverlay) {
                 addLoadingOverlay();
             }
-        } else {
-            setConnecting(false);
-            console.error("Failed to open popup window");
+            return;
+        }
+        setConnecting(false);
+        const newTab = window.open(url, "_blank");
```
pgarciaegido, line 124: I think when a new tab opens, we dont need for the button to be `Connecting...` nor the overlay to be set IMO, since you are in a different context.

 Its a bit weird that you have to close the recently opened tab for the overlay to disappear...

packages/core/base/src/services/crossmintModalService.ts
```diff
@@ -118,9 +118,12 @@ export function crossmintModalService({
...
+        if (!newTab) {
+            console.error("Failed to open popup window and new tab");
```
pgarciaegido, line 126: It would be excellent if we could monitor we are reaching this point.

PR #107, opened 2022-05-06
Send ETH query param in lower case

pgarciaegido:


packages/core/base/src/services/crossmintModalService.ts
```diff
@@ -14,7 +14,7 @@ type MintQueryParams = {
     clientVersion: string;
     mintConfig: string;
     whPassThroughArgs?: string;
-    paymentMethod?: paymentMethods;
+    paymentMethod?: string;
```
angel-paella, line 17: why this change?

pgarciaegido: Becuase the type is `"fiat" | "ETH"` and we are technically passing `eth`

PR #117, opened 2022-05-20
Research react 18

pgarciaegido:


packages/ui/react-ui/src/useEnvironment.ts
```diff
@@ -0,0 +1,10 @@
+import { useEffect, useState } from "react";
+
+export default function useEnvironment() {
+    const [isSSR, setIsSSR] = useState(true);
```
afeight, line 4: is SSR a common enough term that it can be easily interpreted or should we expand it out into `isServerSideRendering`?

packages/ui/react-ui/src/useEnvironment.ts
```diff
@@ -0,0 +1,10 @@
...
+    useEffect(() => {
+        setIsSSR(false);
+    }, []);
```
afeight, line 7: more just curious here, what's this do since we initialize the value to `true` but then set it to `false` when the component loads?

pgarciaegido: - useEffect is only executed on the client side. 
- Adding a `[]` as the second parameter means to get executed only once. This is the same as the old `didComponentMount` lifecycle hook.
- In this hook we assume that, if running in a nextjs app, the first render will happen on the server. Then, on hydration time (once client side code is being executed), we set this value to false. We are in the client now. 
- We want to render the button only after the SSR has happened. 
- Why? Because it was messing up with the styling of then button on subsequent interactions. 
- Why? Because how `react-jss` works: it was creating a unique class per element on SSR and another one on CSR. This ended up in inconsistent styling behaviour. So its better to use `react-jss` on the client side

Hope this makes sense 😄

afeight: Makes a lot of sense! Thanks for explaining it so well 🙂

mPaella (approved):

PR #119, opened 2022-05-26
Remove light theme

pgarciaegido:
So we have a situation at the moment where our latest release is not correctly working on some React 17 environments. 

The reason for this is beacuse incompatibilities with the newest alpha version from `react-jss`. I have opened an issue here: https://github.com/cssinjs/jss/issues/1618

Now the reason why I had to update to the new alpha version is because the latest stable version was having trouble with dynamic styles in React 18.

So... in order to fix all of that, and as an emergency measure, I have disabled the theming in the button, that was the only place where we were using dynamic styles with `react-jss`.

This is an emergency measure until I hear back from the `react-jss` team.

Users that want to use `theme="light"` with this version will need to do it manually on their end. They can follow our styling guide from the docs. I can also assist with that if needed.

### DIFF
--- a/packages/ui/react-ui/package.json
+++ b/packages/ui/react-ui/package.json
@@ -33,7 +33,7 @@
     },
     "dependencies": {
         "@crossmint/client-sdk-base": "^0.1.3-alpha.1",
-        "react-jss": "10.9.1-alpha.1",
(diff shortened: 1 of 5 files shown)


pgarciaegido:
No longer relevant

PR #123, opened 2022-08-02
Add payment method SOL

pgarciaegido:
Adding payment method SOL

### DIFF
--- a/packages/core/base/src/services/crossmintPayButtonService.ts
+++ b/packages/core/base/src/services/crossmintPayButtonService.ts
@@ -38,13 +38,14 @@ export function crossmintPayButtonService({ onClick, connecting, paymentMethod }
         if (connecting) {
             return "Connecting...";
         }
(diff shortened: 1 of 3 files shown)


packages/core/base/src/services/crossmintPayButtonService.ts
```diff
@@ -30,8 +35,15 @@ export function crossmintPayButtonService({ onClick, connecting, paymentMethod }
...
+        if (connecting) {
+            return "Connecting...";
+        }
+        if (paymentMethodIsEth(paymentMethod)) {
+            return "Buy with ETH";
+        }
+        if (paymentMethodIsSol(paymentMethod)) {
+            return "Buy with SOL";
+        }
         return "Buy with credit card";
```
angel-paella, line 47: Maybe just a switch? And then you don't need the paymentMethodIs*

angel-paella (approved):

PR #124, opened 2022-08-02
Preferred signin method solana

pgarciaegido:
Adding preferred signin method -> Solana


angel-paella (approved):
Unit tests are failing

PR #149, opened 2023-03-01
Adds AUD support

guillermodoghel:
Adds AUD as valid currency type


chrism-paella (approved):
Should we also be adding it as a Locale?

guillermodoghel:
Nope, just the currency (there wont be any special translations for en-AU)

(maybe for a next update, I might change what I called `locale`, to `language` to avoid confusing those terms)

PR #151, opened 2023-03-06
PAY-766 Adding successCallbackURL and failureCallbackURL params to the query params

mganopolsky:
This PR adds the PAY-766 Adding `successCallbackURL` and `failureCallbackURL` params to the query params.
It was tested with the test UI starter project.

### DIFF
--- a/packages/core/base/src/services/crossmintModalService.ts
+++ b/packages/core/base/src/services/crossmintModalService.ts
@@ -15,6 +15,9 @@ type MintQueryParams = {
     prepay?: string;
     locale: Locale;
     currency: Currency;
+    layout?: string;
+    successCallbackURL? : string;
+    failureCallbackURL?: string;
 };
 
 const overlayId = "__crossmint-overlay__";
@@ -85,6 +88,9 @@ interface CrossmintModalServiceParams {
     clientName: clientNames;
     locale: Locale;
     currency: Currency;
+    layout?: string;
+    successCallbackURL? : string;
+    failureCallbackURL?: string;
 }
 
 export interface CrossmintModalServiceReturn {
@@ -110,6 +116,9 @@ export function crossmintModalService({
                                           clientName,
                                           locale,
                                           currency,
+                                          layout,
+                                          successCallbackURL,
+                                          failureCallbackURL
                                       }: CrossmintModalServiceParams): CrossmintModalServiceReturn {
     const createPopup = (
         mintConfig: PayButtonConfig,
@@ -129,7 +138,7 @@ export function crossmintModalService({
                 clientVersion: libVersion,
                 mintConfig: JSON.stringify(mintConfig),
                 locale,
-                currency,
+                currency
             };
 
             if (mintTo) mintQueryParams.mintTo = mintTo;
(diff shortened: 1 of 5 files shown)


packages/core/base/src/services/crossmintModalService.ts
```diff
@@ -85,6 +88,9 @@ interface CrossmintModalServiceParams {
     clientName: clientNames;
     locale: Locale;
     currency: Currency;
+    layout?: string;
```
guillermodoghel, line 91: Maybe make layout a type? 

ie `Currency` on `packages/core/base/src/models/types.ts`

packages/ui/react-ui/src/CrossmintPayButton.tsx
```diff
@@ -42,6 +42,9 @@ export const CrossmintPayButton: FC<CrossmintPayButtonReactProps> = ({
     prepay,
     locale = "en-US",
     currency = "USD",
+    layout = "popUpWindow",
+    successCallbackURL = "https://www.crossmint.com/console/",
```
guillermodoghel, line 46: After the payment succeeded or failed, the user would be redirected here directly, or would see the congrats with some action button?

packages/core/base/src/services/crossmintModalService.ts
```diff
@@ -139,6 +148,9 @@ export function crossmintModalService({
             if (paymentMethod) mintQueryParams.paymentMethod = paymentMethod.toLowerCase() as paymentMethods;
             if (preferredSigninMethod) mintQueryParams.preferredSigninMethod = preferredSigninMethod;
             if (prepay) mintQueryParams.prepay = "true";
+            if (layout) mintQueryParams.layout = layout ?? "popUpWindow";
```
mPaella, line 151: we don't need to pass the `layout`, the layout behavior is all within this SDK

packages/ui/react-ui/src/CrossmintPayButton.tsx
mPaella, line 46: do we need fallbacks? we obviously dont want to redirect to the console after a user has checkout out

the flow should remain the same as it was before these changes, if no callbackURLs are passed

packages/core/base/src/services/crossmintModalService.ts
mPaella: When we call `window.open` we should look at what layout was passed, and open the window using that layout

mganopolsky: @mPaella so one of the reasons we are doing this is because some clients are having the following issue:
`In their testing they found that in metamask browser the checkout popup actually takes over the current window. So once payment with crossmint is completed there is no way to get back to the original app/marketplace/minting site. `

would what you are proposing (sounds like you're saying we shouldn't have `layout` at all?) fix the issue? keep it the same?

pgarciaegido (approved):
Makes sense to me

packages/ui/react-ui/src/CrossmintPayButton.tsx
guillermodoghel: I think we could redirect for the initial version, custom buttons could be defined later

PR #157, opened 2023-03-15
Adding new NFT interface

pgarciaegido:
Adding new types for NFT interface as discussed in https://crossmint.slack.com/archives/C04TY75BDA7/p1678867834767079

This makes the experience pretty delightful IMO:

https://user-images.githubusercontent.com/20989060/225292813-6e4e0157-682b-4f11-95d5-a64009d317cf.mov

Notice how much code just good typing removes.

### DIFF
--- a/packages/core/base/src/models/types.ts
+++ b/packages/core/base/src/models/types.ts
@@ -1,3 +1,6 @@
+export const EVM_CHAINS = ["ethereum", "polygon", "bsc"] as const;
+export type EVMChain = (typeof EVM_CHAINS)[number];
+
 export enum clientNames {
     reactUi = "client-sdk-react-ui",
     vanillaUi = "client-sdk-vanilla-ui",
@@ -13,6 +16,15 @@ export enum customHeaders {
     clientVersion = "X-Client-Version",
     clientName = "X-Client-Name",
 }
+
+export enum ChainLocators {
+    solana = "sol",
+    ethereum = "eth",
+    polygon = "poly",
+    cardano = "ada",
+    bsc = "bsc",
+}
+export type EVMChainLocators = ChainLocators.bsc | ChainLocators.ethereum | ChainLocators.polygon;
 
 type theme = "light" | "dark";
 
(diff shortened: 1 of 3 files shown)


packages/core/base/src/models/types.ts
```diff
@@ -113,17 +125,34 @@ export interface NFTCollectionViewProps extends CommonProps {
     wallets: Wallet[];
 }
 
-export interface NFT {
-    address: string;
-    chain: string;
-    tokenId?: string;
+export interface CardanoNFT {
```
Qazalin, line 128: interface AbstractNFT = {
    chain: BlockchainType
} 

EVMNFT extends AbstractNFT?

packages/core/base/src/utils/ui.ts
```diff
@@ -7,41 +7,22 @@ export const getEnvironmentBaseUrl = (environment = ""): string => {
...
-            return "eth";
-        case "polygon":
-            return "poly";
-        case "cardano":
-            return "ada";
-        case "bsc":
-            return "bsc";
-        default:
-            throw new Error(`Invalid chain type ${chain}`);
+function getNFTLocator(nft: NFT): string {
```
Qazalin, line 10: remove :string, doesn't typescript infer that?

packages/core/base/src/utils/ui.ts
```diff
@@ -7,41 +7,22 @@ export const getEnvironmentBaseUrl = (environment = ""): string => {
...
-            return `${chainLocator}:${address}:${tokenId}`;
-        // address in this context is actually an AssetId
+        case "bsc":
+            return `${ChainLocators[nft.chain]}:${nft.contractAddress}:${nft.tokenId}`;
         case "cardano":
-            return `${chainLocator}:${address}`;
+            return `${ChainLocators[nft.chain]}:${nft.assetId}`;
         default:
-            throw new Error(`Invalid chain type ${chain}`);
+            throw new Error(`Invalid chain type ${JSON.stringify(nft)}`);
```
Qazalin, line 25: if chain is wrong, we should only have nft.chain in the error, the other info is irrelevant for this

packages/ui/react-ui/__tests__/CrossmintNFTDetail.spec.tsx
```diff
@@ -2,9 +2,11 @@ import "@testing-library/jest-dom";
 import { render, screen } from "@testing-library/react";
 import React from "react";
 
+import { NFT } from "@crossmint/client-sdk-base";
+
 import { CrossmintNFTDetail } from "../src/CrossmintNFTDetail";
 
-const nft = { chain: "ethereum", address: "0x12345", tokenId: "12" };
+const nft: NFT = { chain: "ethereum", contractAddress: "0x12345", tokenId: "12" };
```
Qazalin, line 9: in the future would be great to have some sorta constraint either within typescript or with zod or something that contractAddress should be 42 chars long

packages/core/base/src/models/types.ts
pgarciaegido: that woudn't fit here, since I want dedicated chain types for each...

packages/ui/react-ui/__tests__/CrossmintNFTDetail.spec.tsx
pgarciaegido: true, typings can evolve from here.

packages/core/base/src/utils/ui.ts
pgarciaegido: done

Qazalin (approved):
your call on the chain thing, we can always improve these later

packages/core/base/src/models/types.ts
pgarciaegido: hhmmm...

Ok agreed that we might need some extra validation for javascript users. Lets do it in next iteration.

PR #163, opened 2023-03-16
Add on event prop

pgarciaegido:
This PR does two things:

- Adds an events type with all checkout events.
- Watches from events sent from the popover (crossbit-main).

That way, users will be able to subscribe to events that we will sent in different stages of the mint.

See demo:
https://www.loom.com/share/ee069a333abc4659a70bc82e7dfa2d50

### DIFF
--- a/packages/core/base/src/models/events.ts
+++ b/packages/core/base/src/models/events.ts
@@ -1,5 +1,7 @@
 export enum CheckoutEvents {
     PAYMENT_READY = "payment:ready",
+    PAYMENT_STARTED = "payment:started",
+    PAYMENT_FAILED = "payment:failed",
     PAYMENT_COMPLETED = "payment:completed",
     PAYMENT_CANCELED = "payment:canceled",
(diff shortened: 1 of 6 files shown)


packages/core/base/src/models/events.ts
```diff
@@ -0,0 +1,11 @@
+export enum CheckoutEvents {
```
pgarciaegido, line 1: This type can be imported in crossbit-main and used so there is a single source of truth for both customers and us

packages/core/base/src/services/crossmintModalService.ts
```diff
@@ -202,6 +205,17 @@ export function crossmintModalService({
                 }
             }
         }, 500);
+
+        window.addEventListener("message", (message) => {
```
mPaella, line 209: we should `window.removeEventListener` when the component unmounts

mPaella (approved):

PR #170, opened 2023-03-17
Adding dts vue UI

pgarciaegido:
OK that was tough...

Added .d.ts files to the bundle so the typing works as expected:

Screenshot 2023-03-17 at 20 38 47

Also started to add types to the button itself... that was a shit show aswell. See the code for more context

### DIFF
--- a/packages/core/base/src/models/paymentElement.ts
+++ b/packages/core/base/src/models/paymentElement.ts
@@ -4,7 +4,7 @@ import { Currency, Locale, PaymentMethod } from "./types";
 // packages/ui/vue-ui/src/components/CrossmintPaymentElement.vue
 export interface PaymentElement {
     clientId: string;
-    mintArgs: Record<string, any>;
+    mintArgs?: Record<string, any>;
     recipient?: string;
     paymentMethod?: PaymentMethod;
     currency?: Currency;
--- a/packages/core/base/src/models/paymentElement.ts
+++ b/packages/core/base/src/models/paymentElement.ts
@@ -1,3 +1,4 @@
+import { CheckoutEventMap } from "./events";
 import { Currency, Locale, PaymentMethod } from "./types";
 
 // TODO: Remmeber to update this same interface in the Vue component aswell.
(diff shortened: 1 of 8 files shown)


packages/ui/vue-ui/src/components/CrossmintPaymentElement.vue
```diff
@@ -1,17 +1,24 @@
 <script setup lang="ts">
-console.log("hola");
-interface hola {
-    adios: string;
+import type { Currency, Locale, PaymentMethod } from "@crossmint/client-sdk-base";
+
+// TODO: Looks like you cannot import the interface directly from the package
+// https://github.com/vuejs/core/issues/4294#issuecomment-970861525
+export interface PaymentElement {
```
pgarciaegido, line 6: Really bad... but it is what it is...

packages/core/base/src/models/paymentElement.ts
```diff
@@ -1,12 +1,14 @@
 import { Currency, Locale, PaymentMethod } from "./types";
 
+// TODO: Remmeber to update this same interface in the Vue component aswell.
+// packages/ui/vue-ui/src/components/CrossmintPaymentElement.vue
 export interface PaymentElement {
     clientId: string;
     mintArgs: Record<string, any>;
```
mPaella, line 7: mint args can be undefined as well

packages/core/base/src/models/paymentElement.ts
```diff
@@ -1,12 +1,14 @@
...
     recipient?: string;
-    paymentMethod: PaymentMethod;
-    currency: Currency;
-    locale: Locale;
+    paymentMethod?: PaymentMethod;
+    currency?: Currency;
+    locale?: Locale;
     // TODO: Use checkout event type when it's ready;
-    onEvent: (event: string, metadata: Record<string, any>) => void;
+    onEvent?: (event: string, metadata: Record<string, any>) => void;
```
mPaella, line 13: lets make this something like this in the future

`onEvent<K extends keyof CrossmintEventMap>(event: K, callback: CrossmintEventMap[K]): this`

pgarciaegido: Dope! Added

mPaella (approved):

PR #171, opened 2023-03-19
Events and service

pgarciaegido:
OK so here it comes. 3 things this does on CrossmintPaymentElement component:
1. Rendering Iframe with /sdk/page
2. Listening to events from iframe, and emit them to `onEvent`
3. Emit event when recipient changes

### DIFF
--- a/packages/core/base/src/models/paymentElement.ts
+++ b/packages/core/base/src/models/paymentElement.ts
@@ -1,14 +1,10 @@
 import { CheckoutEventMap } from "./events";
 import { Currency, Locale, PaymentMethod, UIConfig } from "./types";
 
-export type Recipient =
-    | {
-          email: string;
-          mintTo?: string;
-      }
-    | {
-          mintTo: string;
-      };
+export type Recipient = {
+    email?: string;
+    mintTo?: string;
+};
 
 // TODO: Remmeber to update this same interface in the Vue component aswell.
 // packages/ui/vue-ui/src/components/CrossmintPaymentElement.vue
@@ -23,5 +19,3 @@ export interface PaymentElement {
     environment?: string;
     onEvent?<K extends keyof CheckoutEventMap>(event: K, payload: CheckoutEventMap[K]): this;
 }
-
-export const RecipientEvent = "recipient:updated";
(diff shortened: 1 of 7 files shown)


packages/ui/vue-ui/src/components/CrossmintPaymentElement.vue
```diff
@@ -18,19 +21,35 @@ export interface PaymentElement {
     currency?: Currency;
     locale?: Locale;
     uiConfig?: UIConfig;
-    onEvent?<K extends keyof CheckoutEventMap>(event: K, payload: CheckoutEventMap[K]): this;
+    environment?: string;
+    onEvent?<K extends keyof CheckoutEventMap>(event: K, payload: CheckoutEventMap[K]): any;
```
pgarciaegido, line 25: it was complaining when :this on the return 🤦

packages/core/base/src/services/crossmintPaymentService.ts
```diff
@@ -0,0 +1,47 @@
...
+        });
+    }
+
+    function emitRecipient(recipient?: Recipient) {
+        if (recipient == null) {
+            return;
+        }
+
+        const iframe = document.getElementById("iframe-crossmint-payment-element") as HTMLIFrameElement;
+        iframe?.contentWindow?.postMessage({ type: "queryParamsUpdate", payload: recipient }, baseUrl);
```
pgarciaegido, line 39: We need to define this event name

packages/core/base/src/models/paymentElement.ts
```diff
@@ -1,14 +1,10 @@
...
-export type Recipient =
-    | {
-          email: string;
-          mintTo?: string;
-      }
-    | {
-          mintTo: string;
-      };
+export type Recipient = {
+    email?: string;
```
mPaella, line 5: `email` or `emailTo`?

i had `emailTo` in crossmint-main, but can change

packages/core/base/src/services/crossmintPaymentService.ts
mPaella: yeah, dont need to document but should have a better name

packages/core/base/src/models/paymentElement.ts
pgarciaegido: yeah `emailTo` works

mPaella (approved):

PR #172, opened 2023-03-21
Api changes requested

pgarciaegido:
Improving types and doing some changes requested in https://docs.google.com/document/d/13u0jGRGX7MXXMUQ8-spQITCzfVDUiORQcw1FuYue-2Y/edit#

### DIFF
--- a/packages/core/base/src/models/events.ts
+++ b/packages/core/base/src/models/events.ts
@@ -1,8 +1,3 @@
-export interface CrossmintEvent {
-    type: string;
-    payload: Record<string, any>;
-}
-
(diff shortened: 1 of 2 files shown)


packages/core/base/src/models/events.ts
```diff
@@ -1,23 +1,34 @@
-export enum CheckoutEvents {
+export interface CrossmintEvent {
+    type: string;
+    payload: Record<string, any>;
```
mPaella, line 3: we have the better type in the file below

mPaella (approved):

PR #173, opened 2023-03-21
Ingore ts in 2 rect tests

pgarciaegido:
I was about to do a release, when all of a sudden, 2 of the react tests started failing with this error:
https://stackoverflow.com/questions/71831601/ts2786-component-cannot-be-used-as-a-jsx-component

I tried aligning types and react versioning, but I did not make it work. Tried also to add the resolution, but did not work either...

This is an awful workaround, but Im just avoiding to check the types in these 2 files...

After that, tests pass normally and the starters work as expected:
Screenshot 2023-03-21 at 19 08 28
React starter
Screenshot 2023-03-21 at 19 08 48
Next starter
Screenshot 2023-03-21 at 19 09 01


pgarciaegido:
Found a better fix

PR #187, opened 2023-03-23
Crossmint events

pgarciaegido:
Handling order events:

See demo:
https://www.loom.com/share/a0054d47383d437eac1053266bc399c4


packages/core/base/src/services/useCrossmintEvents.ts
```diff
@@ -1,23 +1,62 @@
 import { CheckoutEvents, ListenToMintingEventsProps, ListenerType } from "../models/events";
-import { fetchOrderStatus } from "../utils/payments";
+import { CrossmintCheckoutEvent } from "../models/paymentElement";
+import { getEnvironmentBaseUrl } from "../utils";
 
-export default function useCrossmintEvents() {
+export function useCrossmintEvents({ environment }: { environment?: string } = {}) {
```
mPaella, line 5: `CrossmintEnvironment`

packages/ui/vue-ui/src/main.ts
```diff
@@ -8,4 +11,5 @@ export default {
     },
 };
 
-export { CrossmintPaymentElement };
+export { CrossmintPaymentElement, useCrossmintEvents, CheckoutEvents };
```
mPaella, line 14: we shouldnt export stuff thats already exported from other packages?

packages/core/base/src/services/useCrossmintEvents.ts
```diff
@@ -1,23 +1,62 @@
...
-                return;
-            }
-            cb(event);
-            emittedEvents.push(event.type);
-            if (event.type === CheckoutEvents.MINTING_COMPLETED) {
-                clearInterval(timer);
+            const events = await fetchOrderStatus({ orderIdentifier });
+
+            for (const event of events) {
+                if (emittedEvents.includes(event.type)) {
```
mPaella, line 15: this feels a bit messy

can we break this out a bit, maybe like:

if hasSeenEventAlready => handleDuplicateEvent
else => handleNewEvent

packages/core/base/src/services/useCrossmintEvents.ts
```diff
@@ -1,23 +1,62 @@
...
+            const res = await fetch(`${getEnvironmentBaseUrl(environment)}/api/sdk/orders/${orderIdentifier}/status`, {
+                method: "GET",
+                headers: {},
+            });
+            const response = await res.json();
+            if (response.error) {
+                return [];
+            }
+            return response;
+        } catch (e) {
```
mPaella, line 55: log

mPaella (commented):
merging for now...

packages/core/base/src/services/useCrossmintEvents.ts
pgarciaegido: hmmm, left environment for consistency, since this is the name of the prop everywhere else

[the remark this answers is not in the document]
pgarciaegido: Passing the object I think its easier to add new stuff if we need it

packages/ui/vue-ui/src/main.ts
pgarciaegido: hmmm why not? I want to make it easy to the user to get all these packages from vue sdk, and don't have to also import the sdk base package.

Similarly when you work with react, that you import everything from one place

packages/core/base/src/services/useCrossmintEvents.ts
pgarciaegido: makes sense. Will improve it today

packages/core/base/src/services/useCrossmintEvents.ts
pgarciaegido: yep

PR #196, opened 2023-03-28
New events payload types

pgarciaegido:
Based on the new models we decided

### DIFF
--- a/packages/core/base/src/models/paymentElement.ts
+++ b/packages/core/base/src/models/paymentElement.ts
@@ -32,19 +32,23 @@ export interface FiatPrice {
     currency: Currency;
 }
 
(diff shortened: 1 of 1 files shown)


packages/core/base/src/models/paymentElement.ts
```diff
@@ -32,19 +32,23 @@ export interface FiatPrice {
     currency: Currency;
 }
 
-interface QuoteBreakdown {
-    unitPrice: FiatPrice;
-    totalGasFees?: FiatPrice;
-    totalCrossmintFees: FiatPrice;
+export interface CollectionMetadata {
```
mPaella, line 35:
```suggestion
export interface ItemMetadata {
```

PR #215, opened 2023-04-05
Add error execution reverted generic

pgarciaegido:
Adding contract reverted generic error. This will be used to handle some contract reverted errors.


packages/core/base/src/models/errors.ts
```diff
@@ -22,6 +22,7 @@ export const PaymentErrors = {
     PAYMENT_REJECTED_PROCESSING_ERROR: "payments:payment-rejected.processing-error",
     PAYMENT_REJECTED_CARD_INCORRECT_NUMBER: "payments:payment-rejected.card-incorrect-number",
     TRANSACTION_ERROR_GENERIC: "payments:transaction-error.generic",
+    CONTRACT_EXECUTION_REVERTED_GENERIC: "payments:contract-execution-reverted.generic",
```
mPaella, line 25: does this need its own event?

to me makes more sense to be a code in an error, instead of its own event

mPaella (approved):

PR #221, opened 2023-04-17
Add listener remove functions

jhubbardsf:
Added a removeEventListeners function to crossmintPaymentService and crossmintUIService. 

I had noticed an issue where events were firing off multiple times if the embedded payment component was removed and recreated. I added this in so I could remove the listeners on the component destroy callback.

Edit: I just realized #219 covers this same bug in a different way.


emalikterzi:
@jhubbardsf Hello , #219 is merged and released fyi

Issue #224, opened 2023-04-19
Vue and Lit CrossmintPaymentElement components don't remove the event listener on unmount


jhubbardsf:
The React UI package was updated to take advantage of the returned removeEventListener from the `listenToEvents` function in `crossmintUiService.ts` and `crossmintPaymentService.ts`. The Vue and Lit web components should do the same thing so that multiple events don't fire if the component is removed from the dom and readded.

pgarciaegido:
Resolved with https://github.com/Crossmint/crossmint-client-sdk/pull/225 and released `alpha.26`

Thanks again @jhubbardsf !!

PR #225, opened 2023-04-19
Remove event listeners when unmounting the Lit and Vue component.

jhubbardsf:
Updated Vue and Lit component to remove event listeners on unmount. This is essentially doing the same thing #219 did for the React component. Covers bug #224.


pgarciaegido (approved):
Looks good! Thanks a lot!

PR #228, opened 2023-04-19
Adding in paymentMethod prop to embedded ui

jhubbardsf:
paymentMethod wasn't being used in `crossmintPaymentService`. Adding the prop according to the docs and updating the type to match the docs.

### DIFF
--- a/packages/core/base/src/models/paymentElement.ts
+++ b/packages/core/base/src/models/paymentElement.ts
@@ -19,7 +19,7 @@ export interface PaymentElement {
     clientId: string;
     mintConfig?: MintConfig;
     recipient?: Recipient;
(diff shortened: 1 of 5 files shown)


pgarciaegido:
Hey @jhubbardsf ,

First of all, thanks a lot for chiming in and contribute! ❤️ 

I agree there is a misalignment between the docs and the code, but in this case is the docs that are not correct! Google pay and Apple pay are enabled by default. No need to set this field for anything. 

We will remove that field from the docs shortly, cause it has a different intention.

Thanks again and sorry for the confusion!

jhubbardsf:
@pgarciaegido Thanks for the clarification!

PR #229, opened 2023-04-21
Add support for new languages

guillermodoghel:
Adds support for the new 12 languages
Orders the dictionary structure
Make currency types lowercase for consistency


packages/core/base/src/utils/i18n/dictionaries/Klingon.ts
pgarciaegido, line 1: xD

pgarciaegido (approved):

PR #231, opened 2023-04-27
Extend UiConfig to include fonts

pgarciaegido:
Add an option to pass fonts to the uiConfig

### DIFF
--- a/packages/core/base/src/models/fonts.ts
+++ b/packages/core/base/src/models/fonts.ts
@@ -9,6 +9,5 @@ export interface CustomFontSource {
     src: string;
     display?: string;
     style?: "normal" | "italic" | "oblique";
-    unicodeRange?: string;
     weight?: string;
 }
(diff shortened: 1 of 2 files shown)


packages/core/base/src/models/types.ts
```diff
@@ -52,7 +54,7 @@ interface StatusButtonConfig {
 }
 
 export type Locale =
-    "en-US"
+    | "en-US"
```
mPaella, line 57: this is valid?

packages/core/base/src/models/fonts.ts
```diff
@@ -0,0 +1,14 @@
...
+    family: string;
+}
+
+export interface CustomFontSource {
+    family: string;
+    // This will be guarded and only Crossmint domain will be allowed
+    src: string;
+    display?: string;
+    style?: "normal" | "italic" | "oblique";
+    unicodeRange?: string;
```
mPaella, line 12:
```suggestion
```

seems like overkill?

mPaella (approved):

packages/core/base/src/models/types.ts
pgarciaegido: Yeah, its what prettier recommends

PR #247, opened 2023-07-03
embed: Adds emailInputOptions and change event

mPaella:


pgarciaegido (commented):
Should we add that prop into React and Vanilla components aswell?

packages/ui/vanilla-ui/src/CrossmintPaymentElement.ts
```diff
@@ -79,14 +85,21 @@ export class CrossmintPaymentElement extends LitElement {
 
     const onEvent = getOnEventFunction(this.onEvent);
 
-    const { listenToEvents } = crossmintPaymentService({ 
-      clientId: this.clientId || this.collectionId,
+    const { listenToEvents } = crossmintPaymentService({
+      clientId: this.clientId,
+      collectionId: this.collectionId,
```
mPaella, line 90: `crossmintPaymentService` handles selecting the correct one, so we dont need to do it here

pgarciaegido (approved):

PR #250, opened 2023-07-07
pay button: update useMemo on getButtonText change

mPaella:
Fixes bug, where button text would update, due to not re-rendering on change


packages/ui/react-ui/src/CrossmintPayButton.tsx
```diff
@@ -110,7 +110,6 @@ export function CrossmintPayButton(buttonProps: CrossmintPayButtonReactProps) {
                     onClick={_handleClick}
                     style={{ ...style }}
                     tabIndex={tabIndex}
-                    {...props}
```
mPaella, line 113: this was incorrectly left here, removing

PR #251, opened 2023-07-19
fix: onEvent type union

mPaella:
Fixes bug for Sound.xyz

OLD:
image

NEW: payload is correctly inferred
image


guillermodoghel (approved):
LGTM!

PR #255, opened 2023-08-03
Adds a checkoutProps to enable the new experimental checkout experience

guillermodoghel:
Default values

```
      checkoutProps = {
            experimental: false,
            display: "same-tab",
            delivery: "all",
            paymentMethods: ["fiat", "ETH", "SOL"],
        }
```

Values

```
type CheckoutProps = {
    experimental?: boolean;
    display: "same-tab" | "new-tab" | "popup";
    delivery: "custodial" | "non-custodial" | "all";
    paymentMethods?: PaymentMethod[];
};

export type PaymentMethod = "fiat" | "ETH" | "SOL";
```


packages/core/base/src/services/hosted/crossmintPayButtonService.ts
```diff
@@ -1,18 +1,22 @@
...
+    checkoutProps?: CheckoutProps;
 }
 
-export function crossmintPayButtonService({ onClick, connecting, paymentMethod, locale }: IProps) {
+export function crossmintPayButtonService({ onClick, connecting, paymentMethod, locale, checkoutProps }: IProps) {
     const getButtonText = (connecting: boolean) => {
         if (connecting) {
             return t("crossmintPayButtonService.CONNECTING", locale);
         }
+        if(checkoutProps?.experimental === true) {
```
mPaella, line 17: if a paymentMethod is passed, we probably want to include that in the button text

or if `paymentMethods` on has 1 item

mPaella (approved):

PR #258, opened 2023-08-09
Bug fix for Safari: new tab cannot be opened

chesovnik:
Open link in current tab if new tab failed


mPaella:
thanks for the pr @chesovnik, and sorry for the delay

could u please post a reproduction of the issue this solves?

PR #260, opened 2023-08-22
Add support for Vietnamese

guillermodoghel:
Adds translation and types for Vietnamese language and currency

### DIFF
--- a/packages/core/base/src/types/index.ts
+++ b/packages/core/base/src/types/index.ts
@@ -35,9 +35,10 @@ export type Locale =
     | "tr-TR"
     | "uk-UA"
     | "th-TH"
+    | "vi-VN"
     | "Klingon";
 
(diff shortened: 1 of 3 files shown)


packages/core/base/src/types/index.ts
```diff
@@ -35,9 +35,10 @@ export type Locale =
     | "tr-TR"
     | "uk-UA"
     | "th-TH"
+    | "vi-VN"
     | "Klingon";
 
-export type Currency = "usd" | "eur" | "gbp" | "aud" | "sgd" | "hkd" | "krw" | "inr";
+export type Currency = "usd" | "eur" | "gbp" | "aud" | "sgd" | "hkd" | "krw" | "inr" | "vnd"
```
pgarciaegido, line 41: why removing the `;`?

pgarciaegido (approved):

packages/core/base/src/types/index.ts
guillermodoghel: that was my IDE 😄

PR #262, opened 2023-08-23
Set the default mintConfig to an empty object

VeniaminC:
Set the default mintConfig to be an empty object in case the mintConfig property was never passed.

Was able to test the vanillajs change locally, unsure how to test the react change


mPaella:
this should be handled on the crossmint-main side

PR #264, opened 2023-09-05
V1.1(pt1-3): Start of types + vue tuneup + organization

mPaella:
Start fixing up types:

1.) Prepares for crypto pay by separating fiat/crypto types
image

2.) Trying to keep old/bad stuff separated, so we can fix in follow up prs

For now the signer interfaces are very simple, next prs will improve with more specific types/chain

image


packages/core/base/src/services/crossmintPaymentService.ts
```diff
@@ -1,9 +1,10 @@
+import { FiatPaymentElementProps, PaymentElementProps } from "../types";
 import { CrossmintCheckoutEventUnion, ParamsUpdatePayload, PaymentElementSDKEvents } from "../types/events";
 import { CheckoutEvents } from "../types/events";
-import { PaymentElement } from "../types/paymentElement";
 import { getEnvironmentBaseUrl } from "../utils";
 
-export function crossmintPaymentService(props: PaymentElement) {
+// TODO: Accept both fiat and crypto
+export function crossmintPaymentService(props: FiatPaymentElementProps) {
```
guillermodoghel, line 7: The reference for the new component is going to be  determined by this function, or an internal redirect on 

`/sdk/paymentElement`?

packages/core/base/src/types/index.ts
```diff
@@ -17,7 +17,18 @@ export enum baseUrls {
     dev = "http://localhost:3001",
 }
 
-export type PaymentMethod = "fiat" | "ETH" | "SOL";
+export const CryptoPaymentMethod = {
+    ETH: "ETH",
+    SOL: "SOL",
```
guillermodoghel, line 22: woud use lowercase to be on sync with `src/types/Currencies.ts`

packages/ui/vanilla-ui/src/CrossmintPaymentElement.ts
```diff
@@ -28,158 +27,170 @@ const propertyDefaults: PaymentElement = {
...
-  height = 0;
-  // eslint-disable-next-line @typescript-eslint/no-empty-function
-  removeEventListener = () => {};
-  // eslint-disable-next-line @typescript-eslint/no-empty-function
-  removeUIEventListener = () => {};
+    height = 0;
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    removeEventListener = () => {};
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    removeUIEventListener = () => {};
```
guillermodoghel, line 85: check linter

packages/core/base/src/types/PaymentElement/crypto.ts
```diff
@@ -0,0 +1,25 @@
...
+};
+
+// Signers
+export type ETHPaymentElementSigner = CommonPaymentElementSignerProps & {
+    signAndSendTransaction: (transaction: any) => Promise<any>;
+};
+
+export type SOLPaymentElementSigner = CommonPaymentElementSignerProps & {
+    signTransaction: (transaction: any) => Promise<any>;
+};
```
guillermodoghel, line 25: Just to confirm, we are going to do so sign on front, broadcast on back?

packages/ui/vanilla-ui/src/CrossmintPaymentElement.ts
mPaella: this is linted correctly

packages/core/base/src/services/crossmintPaymentService.ts
mPaella: sort of both

this function will return the correct iframe url (so, the new route only for crypto pay currently), but then we will also have different components

packages/core/base/src/types/PaymentElement/crypto.ts
mPaella: actually, to start i think its fine if we make this `signAndSendTransaction`

will change

packages/core/base/src/types/PaymentElement/crypto.ts
```diff
@@ -0,0 +1,25 @@
...
+    address: string;
+};
+
+// Signers
+export type ETHPaymentElementSigner = CommonPaymentElementSignerProps & {
+    signAndSendTransaction: (transaction: any) => Promise<any>;
+};
+
+export type SOLPaymentElementSigner = CommonPaymentElementSignerProps & {
+    signTransaction: (transaction: any) => Promise<any>;
```
mPaella, line 24:
```suggestion
    signAndSendTransaction: (transaction: any) => Promise<string>;
};

export type SOLPaymentElementSigner = CommonPaymentElementSignerProps & {
    signAndSendTransaction: (transaction: any) => Promise<string>;
```

PR #266, opened 2023-09-06
V1.1(pt-3): Start rename to CrossmintEmbeddedCheckout + organize

mPaella:
Discussed with Alfonso. We will change naming to `CrossmintEmbeddedCheckout`, as well as likely change the hosted SDK to `CrossmintHostedCheckout`

Since this is still v1.1, the main element `CrossmintPaymentElement`'s name has _not changed_, as that would be a breaking change. So, for now all the related + new types should use the new name. Once v2 is ready, we will make the breaking change to `CrossmintEmbeddedCheckout`


packages/core/base/src/types/EmbeddedCheckout/crypto.ts
```diff
@@ -0,0 +1,25 @@
+import { CommonEmbeddedCheckoutProps } from ".";
```
mPaella, line 1: no new code

guillermodoghel (approved):

PR #273, opened 2023-09-20
Embed/minor adjustments

guillermodoghel:
Better looking demo for the pay with crypto embedded checkout


mPaella:
merged the full connect example with this code into main

PR #279, opened 2023-09-29
Vmc/credentials presentation

vmc-crossmint:
# Added credential presentation SDK
- Now is possible to retrieve all the nfts with some credential associated.
  - Get all nfts in a wallet.
  - Get contract level metadata.
  - Return only nfts that are part of a contract with credentials info in the metadata.
- Filter the returned nfts based on credential type and issuer.
- Added v0.8 contract.
- Extracted rpc provider retrieval.


packages/verifiable-credentials/src/ABI/upgradeable721-v0.8.json
```diff
@@ -0,0 +1,3585 @@
+{
```
mPaella, line 1: lets either make all the ABIs json or ts, but not mix and match

packages/verifiable-credentials/package.json
```diff
@@ -31,6 +31,7 @@
         "@ethersproject/wallet": "5.7.0",
         "@krebitdao/eip712-vc": "1.0.3",
         "date-fns": "2.30.0",
+        "node-fetch": "^3.3.2",
```
mPaella, line 34: u shouldnt need node-fetch

node 18 includes fetch, node 16 does not (but it is end of life and should not be used anymore). so we should be using node 18

packages/verifiable-credentials/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "@crossmint/verify-credentials",
     "private": true,
-    "version": "1.1.1",
+    "version": "1.2.0-alpha.1",
```
mPaella, line 4: u dont need to edit this manually, just follow https://github.com/Crossmint/crossmint-client-sdk/blob/main/README.md

packages/verifiable-credentials/src/presentation/getNfts.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export async function getWalletNfts(chain: string, wallet: string, environment: string) {
+    let page = 1;
+    let hasMore = true;
+    let allData: VC_EVMNFT[] = [];
+    const perPage = 20;
+
+    if (environment === "test") {
+        environment = "http://crossmint-main-git-vmc-newprojectid-crossmint.vercel.app";
+    }
```
mPaella, line 14: probably shouldnt leave this in the release

mPaella (approved):

PR #291, opened 2023-12-01
AA: Error control, order code, clean logs

lluiserena:
- Error control on the RPC messages response to capture 'physicalDeviceId' error.
- Handle earlier the 'Passphrase is required' check if is a new account.
- Clean code

### DIFF
--- a/packages/client/wallets/aa/src/CrossmintAASDK.ts
+++ b/packages/client/wallets/aa/src/CrossmintAASDK.ts
@@ -67,11 +67,7 @@ export class CrossmintAASDK {
 
             return evmAAWallet;
         } catch (e) {
-            if (e instanceof WalletSdkError) {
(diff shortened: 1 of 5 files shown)


packages/client/wallets/aa/src/CrossmintAASDK.ts
```diff
@@ -1,92 +1,91 @@
-import type { SmartAccountSigner } from "@alchemy/aa-core";
-import { ZeroDevEthersProvider, convertEthersSignerToAccountSigner } from "@zerodev/sdk";
-import { Signer } from "ethers";
+import type { SmartAccountSigner } from '@alchemy/aa-core';
```
mPaella, line 1: i think your editor must be using a different prettier config, we use 4 spaces and double quotes

PR #292, opened 2023-12-02
add: generic window comms package

mPaella:
api surface:

```typescript
// common
const FROM_CHILD_EVENTS = {
    shardInitialized: z.object({
       someData: z.string()
    })
}
const FROM_PARENT_EVENTS = {
    initializeShard: z.object({
        deviceId: z.string(),
        walletId: z.string()
    })
}


// parent
const iframe = await IFrameWindow.init("https://crossmint.com/aa/frame", { outgoingEvents: FROM_PARENT_EVENTS, incomingEvents: FROM_CHILD_EVENTS })
// we register listeners before handshaking, but u could do it after as well
iframe.on("shardInitialized", ({ someData }) => ...) // this is fully type + runtime safe bc of Zod

await iframe.handshakeWithChild()

iframe.send("initializeShard", { walletId, deviceId })


// child
const child = await ChildWindow.init(window.parent, "https://harrypotter.com", { outgoingEvents: FROM_CHILD_EVENTS, incomingEvents: FROM_PARENT_EVENTS })
child.on("initializeShard", { deviceId, walletId } => {  // this is fully type + runtime safe bc of Zod
     // do whatever logic    
     child.send("shardInitialized", { someData }) // this is fully type + runtime safe bc of Zod
})

await child.handshakeWithParent()
```

also added an async send and waiting api

```typescript
// sendAction and recieveAction
// this abstracts away the complication of sending/waiting for events, timeouts, etc

// sendAction
// this takes in an event name, and a response event name
// it send the event, then waits for the corresponding response event, allows for conditions to keep waiting for specific data from event

// onAction
// the counterpart for sendAction - wait for an incoming event, and send the corresponding response event
```

image
image


packages/client/window/src/windows/IFrame.ts
```diff
@@ -0,0 +1,54 @@
...
+        }
+        super(contentWindow, targetOrigin, options);
+
+        this.iframe = iframe;
+        this.targetOrigin = targetOrigin;
+    }
+
+    static async init<IncomingEvents extends EventMap, OutgoingEvents extends EventMap>(
+        url: string,
+        options?: EventEmitterWithHandshakeOptions<IncomingEvents, OutgoingEvents>
```
MatiNavCrossmint, line 29: From the example shared in the PR description and from my understanding, I see that the init method always should receive the `outgoingEvents` and the `incomingEvents`.

If so, should this be required?

packages/client/window/src/EventEmitter.ts
```diff
@@ -0,0 +1,177 @@
...
+        this.otherWindow = otherWindow;
+        this.targetOrigin = targetOrigin;
+    }
+
+    send<K extends keyof OutgoingEvents>(event: K, data: z.infer<OutgoingEvents[K]>) {
+        const result = this.outgoingEvents[event].safeParse(data);
+        if (result.success) {
+            this.otherWindow?.postMessage({ event, data }, this.targetOrigin);
+        } else {
+            console.error("Invalid data for event", event, result.error);
```
MatiNavCrossmint, line 70: This comment is about the error handling strategy, so it applies to the other cases as well. I see that we do:
```typescript
try {
    ...do stuff
} catch (error) {
    console.log(error)
}
```

My question is, should we just log or should we throw the error to be handled by the one implementing this package?

MatiNavCrossmint (approved):
This looks amazing. 

The only part that I'm missing is how to properly test it. Should I create two entirely different apps, download this in both, and implement it doing what's in the description?

packages/client/window/src/windows/IFrame.ts
mPaella: i think its fine, sometimes you dont want to send any events, it can be used to just open windows

packages/client/window/src/EventEmitter.ts
mPaella: hmm for now leaving as is, but you are right that the error handling can be improved

PR #293, opened 2023-12-05
embed(pay w crypto): collectionId => clientId

mPaella:
coerces collectionId to clientId properly

image
image

### DIFF
--- a/packages/client/base/src/services/embed/crossmintIFrameService.ts
+++ b/packages/client/base/src/services/embed/crossmintIFrameService.ts
@@ -16,6 +16,8 @@ export function crossmintIFrameService(props: CrossmintEmbeddedCheckoutProps) {
     const targetOrigin = getEnvironmentBaseUrl(props.environment);
 
     function getUrl(props: CrossmintEmbeddedCheckoutProps) {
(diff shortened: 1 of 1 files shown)


packages/client/base/src/services/embed/crossmintIFrameService.ts
```diff
@@ -91,3 +93,10 @@ export function crossmintIFrameService(props: CrossmintEmbeddedCheckoutProps) {
         emitInternalEvent,
     };
 }
+
+function coerceCollectionIdToClientId(props: CrossmintEmbeddedCheckoutProps) {
+    if ("collectionId" in props && props.collectionId) {
```
alfonso-paella, line 98: Can you explain this check? not sure how the && isnt redundant

alfonso-paella (approved):

packages/client/base/src/services/embed/crossmintIFrameService.ts
mPaella: its mainly a ts thing, but empty string, null would satisfy `"collectionId" in props` so the && check prevents those

PR #296, opened 2023-12-05
Depending on the network defined, connect to staging or prod

jordiamez:
If the chain is a testnet we return the staging URL, if the chain is a mainnet we return the prod URL.

### DIFF
--- a/packages/client/wallets/aa/src/CrossmintAASDK.ts
+++ b/packages/client/wallets/aa/src/CrossmintAASDK.ts
@@ -1,14 +1,10 @@
+import { CrossmintService } from "@/api";
+import { Blockchain, EVMAAWallet, EVMBlockchain, FireblocksNCWallet } from "@/blockchain";
+import type { CrossmintAASDKInitParams, FireblocksNCWSigner, UserIdentifier, WalletConfig } from "@/types";
+import { CURRENT_VERSION, WalletSdkError, ZERO_DEV_TYPE, ZERO_PROJECT_ID } from "@/utils";
 import type { SmartAccountSigner } from "@alchemy/aa-core";
 import { ZeroDevEthersProvider, convertEthersSignerToAccountSigner } from "@zerodev/sdk";
 import { Signer } from "ethers";
-
-import { CrossmintService } from "./api/CrossmintService";
-import { Blockchain, EVMBlockchain } from "./blockchain/BlockchainNetworks";
-import { EVMAAWallet } from "./blockchain/wallets/EVMAAWallet";
-import { FireblocksNCWallet } from "./blockchain/wallets/FireblocksNCWallet";
-import type { CrossmintAASDKInitParams, FireblocksNCWSigner, UserIdentifier, WalletConfig } from "./types/Config";
-import { CURRENT_VERSION, ZERO_DEV_TYPE, ZERO_PROJECT_ID } from "./utils/constants";
-import { WalletSdkError } from "./utils/error";
 
 export class CrossmintAASDK {
     crossmintService: CrossmintService;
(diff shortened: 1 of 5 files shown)


packages/client/wallets/aa/src/api/CrossmintService.ts
```diff
@@ -78,15 +86,26 @@ export class CrossmintService {
...
+        const jsonResponse = await response.json();
+
+        this.handleResponse(jsonResponse);
+
+        return jsonResponse;
+    }
+
+    private handleResponse(response: any) {
+        if (response && response.error) {
+            if (response.message === "API key not found") {
```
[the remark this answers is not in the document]
MatiNavCrossmint: All this file has changed, this PR needs to sync with main

packages/client/wallets/aa/src/api/CrossmintService.ts
```diff
@@ -10,6 +13,11 @@ export class CrossmintService {
             "x-client-secret": clientSecret,
             "x-project-id": projectId,
         };
+        this.crossmintBaseUrl = "https://staging.crossmint.com/api";
+    }
+
+    setBlockchainType(blockchain: Blockchain) {
+        this.crossmintBaseUrl = getApiUrlByBlockchainType(blockchain);
     }
```
MatiNavCrossmint, line 21: I think this should be set in the constructor

packages/client/wallets/aa/src/CrossmintAASDK.ts
```diff
@@ -67,11 +64,9 @@ export default class CrossmintAASDK {
 
             return evmAAWallet;
         } catch (e) {
-            if (e instanceof WalletSdkError) {
-                console.log(e.code, e.message);
-            } else {
-                console.error("Unexpected error:", e);
-            }
+            throw new WalletSdkError(
```
MatiNavCrossmint, line 67: Nice

jordiamez:
@MatiNavCrossmint - I've made the changes, thanks for the feedback! 

The line: 
`  this.crossmintService.setCrossmintUrl(chain);`
   
 can't be at the constructor, because we create the crossmintService on the init function and we do not have the CHAIN param at that point so we need to set it on the getOrCreateWallet: 
 
```
    private constructor(config: CrossmintAASDKInitParams) {
        this.crossmintService = new CrossmintService(config.clientSecret, config.projectId);
    }

    static init(params: CrossmintAASDKInitParams): CrossmintAASDK {
        return new CrossmintAASDK(params);
    }
```

maxisch:
> @MatiNavCrossmint - I've made the changes, thanks for the feedback!
> 
> The line: ` this.crossmintService.setCrossmintUrl(chain);`
> 
> can't be at the constructor, because we create the crossmintService on the init function and we do not have the CHAIN param at that point so we need to set it on the getOrCreateWallet:
> 
> ```
>     private constructor(config: CrossmintAASDKInitParams) {
>         this.crossmintService = new CrossmintService(config.clientSecret, config.projectId);
>     }
> 
>     static init(params: CrossmintAASDKInitParams): CrossmintAASDK {
>         return new CrossmintAASDK(params);
>     }
> ```

Question here, why isn't the blockchain part of the constructor? Can (or should) the chain change during the lifecycle?

jordiamez:
> > @MatiNavCrossmint - I've made the changes, thanks for the feedback!
> > The line: ` this.crossmintService.setCrossmintUrl(chain);`
> > can't be at the constructor, because we create the crossmintService on the init function and we do not have the CHAIN param at that point so we need to set it on the getOrCreateWallet:
> > ```
> >     private constructor(config: CrossmintAASDKInitParams) {
> >         this.crossmintService = new CrossmintService(config.clientSecret, config.projectId);
> >     }
> > 
> >     static init(params: CrossmintAASDKInitParams): CrossmintAASDK {
> >         return new CrossmintAASDK(params);
> >     }
> > ```
> 
> Question here, why isn't the blockchain part of the constructor? Can (or should) the chain change during the lifecycle?

@maxisch We have the init function on AAWallet that works as a constructor to make the connection with crossbitmain

We could add the blockchain on that point but on the document design we have only the CrossmintAASDKInitParams that includes only clientSecret projectId so we can't add it there. That's why we have the setCrossmintUrl(chain);

maxisch (approved):

PR #297, opened 2023-12-06
WAL-1447: Crossmint Service error handling

maxisch:


MatiNavCrossmint (commented):
It looks good, and I liked the strategy to keep the error handling clean 👏

There are some minor comments about some of the errors. And I don't have a strong opinion about whether we should add "Please contact support" in all error occurrences. I mean, these errors will be seen by external users who don't have knowledge ( and shouldn't care ) about what happens inside the SDK. So including this message could be a good ally for them when something happens. But up to you.

I won't approve, as it's in draft. So lmk when it's 100% done

kikoncuo (approved):
LGTM

PR #298, opened 2023-12-07
WAL-1258: Added Web3AuthSigner

jordiamez:
## Description

Added new signer type. Web3AuthSigner.  Using the following definition: 

export type Web3AuthSigner =  {
    type: "WEB3_AUTH";
    clientId: string; //Web3Auth Client ID
    verifierId: string; //Web3Auth Verifier ID
    jwt: string;
};
Attached screenshots to better understanding of the clientId and verifierId

Captura de pantalla 2023-12-12 a las 11 46 25
Captura de pantalla 2023-12-12 a las 11 46 20
Captura de pantalla 2023-12-12 a las 11 46 15

## Test

Tested on scw_demo: 

https://github.com/Crossmint/crossmint-sdk/assets/131860331/21456187-00b4-4d8d-ba0c-18be1d553336

### DIFF
--- a/packages/client/wallets/aa/src/blockchain/BlockchainNetworks.ts
+++ b/packages/client/wallets/aa/src/blockchain/BlockchainNetworks.ts
@@ -1,4 +1,5 @@
-import { CROSSMINT_PROD_URL, CROSSMINT_STG_URL } from "@/utils";
+import { CROSSMINT_PROD_URL, CROSSMINT_STG_URL, WEB3_AUTH_MAINNET, WEB3_AUTH_TESTNET } from "@/utils";
+import { TORUS_LEGACY_NETWORK_TYPE } from "@web3auth/single-factor-auth";
 
 export const EVMBlockchain = {
     ETHEREUM: "ethereum",
@@ -21,7 +22,6 @@ export const BlockchainTestNet = {
     GOERLI: "goerli",
     SEPOLIA: "sepolia",
     MUMBAI: "mumbai",
-    TESTNET: "testnet",
 } as const; //testnet as a placeholder for non-EVM chains
 export type BlockchainTestNet = (typeof BlockchainTestNet)[keyof typeof BlockchainTestNet];
 
@@ -46,7 +46,6 @@ export function getAssetIdByBlockchain(chain: Blockchain) {
         [Blockchain.GOERLI, "ETH_TEST3"],
         [Blockchain.SEPOLIA, ""], // TODO
         [Blockchain.MUMBAI, "MATIC_POLYGON_MUMBAI"],
-        [Blockchain.TESTNET, ""], // TODO
     ]).get(chain)!;
 }
 
@@ -64,10 +63,26 @@ export function getBlockchainByChainId(chain: number) {
         [5, Blockchain.GOERLI],
         //[0, Blockchain.SEPOLIA], // TODO
         [80001, Blockchain.MUMBAI],
-        [0, Blockchain.TESTNET], // TODO
     ]);
 
     return chainIdMap.get(chain);
+}
+
+export function getChainIdByBlockchain(chain: Blockchain) {
+    return new Map([
+        [Blockchain.ETHEREUM, 1],
+        [Blockchain.POLYGON, 137],
+        [Blockchain.BSC, 56],
+        [Blockchain.OPTIMISM, 0], // TODO
+        [Blockchain.ARBITRUM, 42161],
+        [Blockchain.ARBITRUM_NOVA, 42170],
+        [Blockchain.ZORA, 7777777],
+        [Blockchain.SOLANA, 0], // TODO
+        [Blockchain.CARDANO, 0], // TODO
+        [Blockchain.GOERLI, 5],
+        [Blockchain.SEPOLIA, 0], // TODO
+        [Blockchain.MUMBAI, 80001],
+    ]).get(chain)!;
 }
 
 export function getUrlProviderByBlockchain(chain: Blockchain) {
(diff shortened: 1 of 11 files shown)


packages/client/wallets/aa/src/utils/auth.ts
```diff
@@ -0,0 +1,10 @@
+export const parseToken = (token: any) => {
+    try {
+        const base64Url = token.split(".")[1];
+        const base64 = base64Url.replace("-", "+").replace("_", "/");
+        return JSON.parse(window.atob(base64 || ""));
+    } catch (err) {
+        console.error(err);
+        return null;
```
maxisch, line 8: let's throw an Exception here

packages/client/wallets/aa/src/utils/signer.ts
```diff
@@ -0,0 +1,64 @@
...
+import { EthereumPrivateKeyProvider } from "@web3auth/ethereum-provider";
+import { Web3Auth } from "@web3auth/single-factor-auth";
+import { convertEthersSignerToAccountSigner, getRPCProviderOwner } from "@zerodev/sdk";
+import { Signer } from "ethers";
+
+export async function createOwnerSigner(
+    user: UserIdentifier,
+    chain: Blockchain,
+    walletConfig: WalletConfig,
+    crossmintService: CrossmintService
```
maxisch, line 25: let's make it optional and check not null only for fireblocks ncw case

packages/client/wallets/aa/src/utils/signer.ts
```diff
@@ -0,0 +1,64 @@
...
+            rpcTarget: getUrlProviderByBlockchain(chain),
+            displayName: getDisplayNameByBlockchain(chain),
+            blockExplorer: getBlockExplorerByBlockchain(chain),
+            ticker: getTickerByBlockchain(chain),
+            tickerName: getTickerNameByBlockchain(chain),
+        };
+
+        const web3auth = new Web3Auth({
+            clientId: signer.clientId, // Get your Client ID from Web3Auth Dashboard
+            web3AuthNetwork: "testnet", // ["cyan", "testnet"]
```
maxisch, line 46: planning to do something similiar as done here to define the network? https://github.com/Crossmint/crossmint-sdk/pull/296/files#diff-44a8de940941124bdd920a8721c56d3a0a9262939880fe56f3a457a96c301562R91

packages/client/wallets/aa/src/utils/signer.ts
maxisch: What I wanted to say is: the `crossmintService` is only necessary for the `FIREBLOCKS_NCW`
When we use `WEB3_AUTH` we don't need it. 
So I'm suggesting to make the parameter optional in the `createOwnerSigner` signature, and check that it is not null when we are in a `FIREBLOCKS_NCW` scenario. (throw excpetion if null I guess)

packages/client/wallets/aa/src/blockchain/BlockchainNetworks.ts
```diff
@@ -68,6 +68,24 @@ export function getBlockchainByChainId(chain: number) {
...
+        [Blockchain.OPTIMISM, 0], // TODO
+        [Blockchain.ARBITRUM, 42161],
+        [Blockchain.ARBITRUM_NOVA, 42170],
+        [Blockchain.ZORA, 7777777],
+        [Blockchain.SOLANA, 0], // TODO
+        [Blockchain.CARDANO, 0], // TODO
+        [Blockchain.GOERLI, 5],
+        [Blockchain.SEPOLIA, 0], // TODO
+        [Blockchain.MUMBAI, 80001],
+        [Blockchain.TESTNET, 0], // TODO
```
MatiNavCrossmint, line 85: Is this a typo?

packages/client/wallets/aa/src/blockchain/BlockchainNetworks.ts
```diff
@@ -85,3 +103,75 @@ export function getUrlProviderByBlockchain(chain: Blockchain) {
         [Blockchain.TESTNET, ""],
```
MatiNavCrossmint, line 103: Same comment as [above](https://github.com/Crossmint/crossmint-sdk/pull/298#discussion_r1424586009)

packages/client/wallets/aa/src/blockchain/BlockchainNetworks.ts
MatiNavCrossmint, line 85: Is this `Blockchain.TESTNET` a typo?

packages/client/wallets/aa/src/utils/signer.ts
MatiNavCrossmint: What you say Maxi makes sense, however, it's already marked as required in the proposal so I would leave this as an improvement to make it in the future ( we need to sync it with the proposal).

Also, what I realized is that if a AA Wallet is created without Fireblocks, then we don't save any data about this wallet and we can not use it at all. Even on Wallet Connect. I believe that it's clear but just realizing that, right?

packages/client/wallets/aa/src/blockchain/BlockchainNetworks.ts
jordiamez: I think it's the future localtesnet, but I can delete it and add it later

packages/client/wallets/aa/src/utils/signer.ts
jordiamez: I'm not sure if I understand your question @MatiNavCrossmint. We store always the abstract wallet info:


await this.crossmintService.storeAbstractWallet({
                userEmail: user.email!,
                type: ZERO_DEV_TYPE,
                smartContractWalletAddress: abstractAddress,
                eoaAddress: address,
                sessionKeySignerAddress,
                version: CURRENT_VERSION,
                baseLayer: "evm",
            });

packages/client/wallets/aa/src/utils/signer.ts
jordiamez: I was waiting for the other PR merged, I'm doing it right now

jordiamez: @maxisch - Done!

packages/client/wallets/aa/src/blockchain/BlockchainNetworks.ts
```diff
@@ -99,3 +181,15 @@ export function getApiUrlByBlockchainType(chain: Blockchain): string {
         return CROSSMINT_PROD_URL;
     }
 }
+
+export function getWeb3AuthBlockchain(chain: Blockchain): TORUS_LEGACY_NETWORK_TYPE {
```
maxisch, line 185: one last change, this method is identical to `getApiUrlByBlockchainType`. I suggest a small refactor.
a private method isTestNet(chain) with this logic returning a bool and later
```
getWeb3AuthBlockchain(chain: Blockchain) {
    return isTestNet(chain) ? WEB3_AUTH_TESTNET : WEB3_AUTH_MAINNET
}
```
and:
```
getApiUrlByBlockchainType(chain: Blockchain):
    return isTestNet(chain) ? CROSSMINT_STG_URL : CROSSMINT_PROD_URL
```

jordiamez: Done

maxisch (approved):

PR #300, opened 2023-12-07
WAL-1445: big scope error handling

maxisch:
### Description
We are bundling errors regarding RPC-Bundlers-Blockchain-ZeroDev into a big try-catch flow, and returning a user friendly error

### Goal
Client developer does not get any cryptic exceptions that were not catched.

### Test plan

Ensure workflow is still working, view friendlier errors in case of an error.


MatiNavCrossmint (approved):
LGTM!

PR #304, opened 2023-12-11
WAL-1456: Add native token and ERC20 support for the transfer function in our SDK

lluiserena:
Added ERC20 ABI.

Added a new param 'amount' on the 'Transfer' method on the BaseWallet. If this param has info it will perform a ERC20 tx.


maxisch (approved):
LGTM

PR #305, opened 2023-12-11
WAL-1445: Purge wallet data

lluiserena:
Deletes all the LocalStorage info which key starts with 'NCW-'


maxisch:
@jordiamez I would remove the function `purgeUserWalletData` so the developers don't see this option at all, until we figure out the localstorage handling.
@kikoncuo agree?

jordiamez:
> @jordiamez I would remove the function `purgeUserWalletData` so the developers don't see this option at all, until we figure out the localstorage handling. @kikoncuo agree?

Done!

maxisch (approved):

PR #310, opened 2023-12-12
add support for prod wallets

afeight:
## Description

Add support for prod wallets

## Test plan

Tested pointing to my local server w/ prod keys


packages/client/wallets/aa/src/blockchain/wallets/FireblocksNCWallet.ts
```diff
@@ -61,7 +61,10 @@ export const FireblocksNCWallet = async (
         return crossmintService.getNCWIdentifier(deviceId);
     });
 
+    const testnets: Blockchain[] = [BlockchainTestNet.GOERLI, BlockchainTestNet.MUMBAI, BlockchainTestNet.SEPOLIA, BlockchainTestNet.TESTNET]
+    const testnet = testnets.includes(chain)
```
afeight, line 65: @mPaella I assume there's a better TS way to do this, but I was unable to treat `BlockchainTestNet` as a dict to do something like `.values().includes(...)`, I was only able to treat it as a type

MatiNavCrossmint (approved):
Amazing !!

PR #313, opened 2023-12-13
Bug: creating signer always on staging.

maxisch:
## Description

Before we were creating a signer ALWAYS on staging, and later deciding the environment based on the chain (could be prod)

## Test plan

Running everything in localhost, the signer was being created in stagging.


maxisch:
Included in: https://github.com/Crossmint/crossmint-sdk/pull/314

PR #321, opened 2023-12-18
Fixes Using “new tab” display doesn’t open on second attempt

guillermodoghel:
## Description
When `CrossmintPaymentElement` is configured with `new-tab`  the button keeps the "connecting" state blocking to allow re-clik if the tab is closed

## Test plan
Tested locally with this specific config

### DIFF
--- a/packages/core/base/src/services/hosted/crossmintModalService.ts
+++ b/packages/core/base/src/services/hosted/crossmintModalService.ts
@@ -198,6 +198,7 @@ export function crossmintModalService({
                 }
                 case "new-tab": {
                     window.open(url, "_blank");
+                    setConnecting(false);
                     return;
                 }
                 case "same-tab":


packages/client/base/src/services/hosted/crossmintModalService.ts
```diff
@@ -198,6 +198,7 @@ export function crossmintModalService({
                 }
                 case "new-tab": {
                     window.open(url, "_blank");
+                    setConnecting(false);
```
mPaella, line 201: i think we should do the same thing as the pop, like:

```
const newTab = window.open(url, "_blank");
if (newTab ) {
    registerListeners(pop);
    if (showOverlay) {
        addLoadingOverlay(dismissOverlayOnClick);
    }
}
```

mPaella (approved):

PR #326, opened 2023-12-22
WAL-1529: add zkatana

maxisch:
## Description
Add support for zkatana testnet 

## Test plan

Manual testing


packages/client/wallets/aa/src/blockchain/BlockchainNetworks.ts
```diff
@@ -69,19 +72,21 @@ export function getAssetIdByBlockchain(chain: Blockchain) {
...
-        [80001, Blockchain.MUMBAI],
+        [ETHEREUM_CHAIN_ID, Blockchain.ETHEREUM],
+        [POLYGON_CHAIN_ID, Blockchain.POLYGON],
+        [BSC_CHAIN_ID, Blockchain.BSC],
+        [OPTIMISM_CHAIN_ID, Blockchain.OPTIMISM],
+        [ARBITRUM_CHAIN_ID, Blockchain.ARBITRUM],
+        [GOERLI_CHAIN_ID, Blockchain.GOERLI],
+        [SEPOLIA_CHAIN_ID, Blockchain.SEPOLIA],
+        [MUMBAI_CHAIN_ID, Blockchain.MUMBAI],
+        [ZKATANA_CHAIN_ID, Blockchain.ZKATANA],
```
afeight, line 89: ooh very nice, thanks!

afeight (approved):
lgtm

just as a separate thought–it seems like we had to add this value into a bunch of different places, none of which are compiler-enforced. Ideally it'd look like: you add new `Blockchain` value, then the app doesn't build until you fill in all of the proper slots. Any thoughts on what we could do to achieve this?

maxisch:
> lgtm
> 
> just as a separate thought, not to block this PR–it seems like we had to add this value into a bunch of different places, none of which are compiler-enforced. Ideally it'd look like: you add new `Blockchain` value, then the app doesn't build until you fill in all of the proper slots. Any thoughts on what we could do to achieve this?

I'll see if we can do something similar to what's been done in main

PR #330, opened 2023-12-29
NCW: Remove datadog

MatiNavCrossmint:
## Description

As datadog requires to run in the browser, Crossmint is failing when importing the SDK and. the bug isn't straightforward to resolve. To unblock Crossmint, we remove Datadog and then dedicate a special time and approach to fix it

## Test plan

Tested locally creating an AA wallet and checking that it works


kikoncuo (approved):
LGTM

PR #337, opened 2024-01-05
SCW-81: Accept userid and phonenumber as user identifiers

lluiserena:
## Description

Accept userid and phonenumber as user identifiers on the creation of the SCW

This changes will work with the following changes done in Crossbit-main (PR in review process) https://github.com/Paella-Labs/crossbit-main/pull/10729 

## Test plan

Create different SCW using different parameters (email, projectId, phoneNumber)


packages/client/wallets/aa/src/utils/user.ts
```diff
@@ -0,0 +1,75 @@
...
+        );
+    }
+
+    if (!startsWithPlus) {
+        input = `+${input.trimStart()}`;
+    }
+
+    const phoneNumber = parsePhoneNumber(input);
+    if (!phoneNumber.isValid()) {
+        console.log(`PDWARN - invalid phone number reported (${input}).`);
```
MatiNavCrossmint, line 60: Can we remove the `PDWARN`?

MatiNavCrossmint (commented):
I'm a bit skeptical to see all this user logic here. I'll approved but we should refactor this to consume a Crossmint endpoint ( which doesn't exists currently). 

Can we create a ticket to handle to do a second look to this with low priority?

packages/client/wallets/aa/src/types/Config.ts
```diff
@@ -6,12 +6,17 @@ export type CrossmintAASDKInitParams = {
...
-export type UserIdentifier = {
-    email: string;
+export type UserIdentifierParams = {
+    email?: string;
     userId?: string;
     phoneNumber?: string;
 };
 
+export type UserIdentifier =
+    | { type: "white-label"; userId: string }
```
afeight, line 16: nit: consistency w/ `phoneNumber`
```suggestion
    | { type: "whiteLabel"; userId: string }
```

packages/client/wallets/aa/src/utils/signer.ts
```diff
@@ -10,8 +10,9 @@ import {
     getUrlProviderByBlockchain,
     getWeb3AuthBlockchain,
 } from "@/blockchain";
-import { FireblocksNCWSigner, UserIdentifier, WalletConfig, Web3AuthSigner } from "@/types";
+import { FireblocksNCWSigner, UserIdentifier, UserIdentifierParams, WalletConfig, Web3AuthSigner } from "@/types";
```
afeight, line 13: I don't see `UserIdentifierParams` being used anywhere, can we remove?
```suggestion
import { FireblocksNCWSigner, UserIdentifier, WalletConfig, Web3AuthSigner } from "@/types";
```

packages/client/wallets/aa/src/CrossmintAASDK.ts
```diff
@@ -17,17 +25,19 @@ export class CrossmintAASDK {
...
         walletConfig: WalletConfig
     ) {
         try {
             logInfo("[GET_OR_CREATE_WALLET] - INIT", {
-                userEmail: user.email!,
+                user,
                 chain,
             });
+            const userIdentifier = parseUserIdentifier(user);
+
```
afeight, line 38: nit: we use this on line 40, so let's put this line on the line above
```suggestion
```

packages/client/wallets/aa/src/CrossmintAASDK.ts
```diff
@@ -17,17 +25,19 @@ export class CrossmintAASDK {
...
             this.crossmintService.setCrossmintUrl(chain);
-            const owner = await createOwnerSigner(user, chain, walletConfig, this.crossmintService);
+            const owner = await createOwnerSigner(userIdentifier, chain, walletConfig, this.crossmintService);
```
afeight, line 40:
```suggestion
            const userIdentifier = parseUserIdentifier(user);
            const owner = await createOwnerSigner(userIdentifier, chain, walletConfig, this.crossmintService);
```

packages/client/wallets/aa/src/utils/user.ts
```diff
@@ -0,0 +1,75 @@
+import { UserIdentifier, UserIdentifierParams } from "@/types/Config";
+import { isEmpty } from "@/utils/helpers";
+import EmailValidator from "email-validator";
+import { parsePhoneNumber } from "libphonenumber-js";
+
+export function narrowUserIdentifier(userIdentifier: UserIdentifier) {
+    switch (userIdentifier.type) {
+        case "white-label":
```
afeight, line 8: ^
```suggestion
        case "whiteLabel":
```

packages/client/wallets/aa/src/utils/user.ts
```diff
@@ -0,0 +1,75 @@
+import { UserIdentifier, UserIdentifierParams } from "@/types/Config";
```
afeight, line 1: are these copied from somewhere?

afeight: I think we're copying these from crossbit-main. If that's the case, I'd love to put these into `@crossmint/client-sdk-base` and let both the AA sdk and Crossbit-main pull them in as dependencies.
Is that possible?

packages/client/wallets/aa/src/CrossmintAASDK.ts
```diff
@@ -1,9 +1,17 @@
...
+    getChainIdByBlockchain,
+    getZeroDevProjectIdByBlockchain,
+    isEVMBlockchain,
+} from "@/blockchain";
+import type { CrossmintAASDKInitParams, UserIdentifierParams, WalletConfig } from "@/types";
 import { CURRENT_VERSION, WalletSdkError, ZERO_DEV_TYPE, createOwnerSigner, errorToJSON } from "@/utils";
 import { ZeroDevEthersProvider } from "@zerodev/sdk";
+
 import { logError, logInfo } from "./services/logging";
+import { narrowUserIdentifier, parseUserIdentifier } from "./utils/user";
```
afeight, line 14: unused import
```suggestion
import { parseUserIdentifier } from "./utils/user";
```

packages/client/wallets/aa/src/utils/user.ts
```diff
@@ -0,0 +1,75 @@
...
+import { isEmpty } from "@/utils/helpers";
+import EmailValidator from "email-validator";
+import { parsePhoneNumber } from "libphonenumber-js";
+
+export function narrowUserIdentifier(userIdentifier: UserIdentifier) {
```
afeight, line 6: is this needed? I think it's ok to have extra params in either case, whether it's a log or an API call, I'm not worried about including the `type`

afeight (commented):
For the most part, this looks good
I'd like for us to take a different approach to reduce code duplication though between crossbit main and the SDK [comment here]()

packages/client/wallets/aa/src/utils/user.ts
jordiamez: This change should be at another PR, if not we cannot use it because it is not deployed and will break everything. I'm going to create a new PR for that and after that we can switch to use `@crossmint/client-sdk-base`

packages/client/wallets/aa/src/utils/user.ts
jordiamez: Yes it's needed, but we can discuss it.

packages/client/wallets/aa/src/api/CrossmintService.ts
```diff
@@ -51,22 +52,22 @@ export class CrossmintService {
...
     }
 
-    async getOrAssignWallet(userEmail: string) {
+    async getOrAssignWallet(userIdentifier: UserIdentifier) {
         return this.fetchCrossmintAPI(
             "unstable/wallets/aa/ncw",
-            { method: "POST", body: JSON.stringify({ userEmail }) },
-            `Error getting or assigning wallet for user: ${userEmail}`
+            { method: "POST", body: JSON.stringify({ userIdentifier }) },
+            `Error getting or assigning wallet for user: ${narrowUserIdentifier(userIdentifier)}`
```
afeight, line 59: I still think we should remove `narrowUserIdentifier` entirely and just do 
```suggestion
            `Error getting or assigning wallet for user: ${JSON.stringify(userIdentifier)}`
```

packages/client/wallets/aa/src/api/CrossmintService.ts
```diff
@@ -51,22 +52,22 @@ export class CrossmintService {
...
+    async unassignWallet(userIdentifier: UserIdentifier) {
         return this.fetchCrossmintAPI(
             "unstable/wallets/aa/ncw/unassign",
             {
                 method: "POST",
-                body: JSON.stringify({ userEmail }),
+                body: JSON.stringify({ userIdentifier }),
             },
-            `Error unassigning wallet for user: ${userEmail}`
+            `Error unassigning wallet for user: ${narrowUserIdentifier(userIdentifier)}`
```
afeight, line 70: ^

afeight (approved):
approve barring comments

packages/client/verifiable-credentials/package.json
```diff
@@ -6,10 +6,11 @@
         "@krebitdao/eip712-vc": "1.0.3",
         "@lit-protocol/lit-node-client": "3.0.24",
         "date-fns": "2.30.0",
-        "ethers": "5.7.2"
+        "ethers": "5.7.2",
+        "@crossmint/common-sdk-base": "0.0.3"
     },
     "devDependencies": {
-        "@crossmint/client-sdk-base": "1.1.3"
+        "@crossmint/client-sdk-base": "1.1.4"
```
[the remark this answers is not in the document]
jordiamez: All these changes are in this commit: 

https://github.com/Crossmint/crossmint-sdk/pull/347

PR #340, opened 2024-01-10
SCW-300 Check status tx

jordiamez:
## Description

**Improved Error Handling in Token Transfers:**

In the `transfer` function, we now check the `status` of the transaction receipt. If the transaction is successful (i.e., `status === 1`), the function returns the transaction hash as before. However, if the transaction fails (`status !== 1`), it now throws a TransferError with a descriptive message including the token ID and the transaction hash. This makes it easier to debug issues with token transfers as it provides more context about the failure.

**Refined Transaction Error Handling:**

In the `sendTransaction` method, we've implemented a try-catch block. If an error occurs while sending the transaction, instead of allowing it to propagate up as a generic error, the method now catches the error and throws a `TransactionError`. This encapsulates the error context better and makes the error handling more consistent across our blockchain-related methods.
These changes aim to provide clearer and more consistent error handling mechanisms, making it easier to debug and handle transaction failures.

## Test plan

SendTransaction function and transfer

### DIFF
--- a/packages/client/wallets/aa/src/blockchain/wallets/BaseWallet.ts
+++ b/packages/client/wallets/aa/src/blockchain/wallets/BaseWallet.ts
@@ -98,7 +98,9 @@ class BaseWallet extends ZeroDevAccountSigner<"ECDSA"> {
                 return transaction!.hash;
             } else {
                 throw new TransferError(
-                    `Error transferring token ${evmToken.tokenId} with transaction hash ${transaction!.hash}`
+                    `Error transferring token ${evmToken.tokenId}${
+                        !transaction || !transaction.hash ? "" : ` with transaction hash ${transaction.hash}`
+                    }`
                 );
             }
         } catch (error) {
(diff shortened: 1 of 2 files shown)


packages/client/wallets/aa/src/blockchain/wallets/BaseWallet.ts
```diff
@@ -92,13 +94,25 @@ class BaseWallet extends ZeroDevAccountSigner<"ECDSA"> {
...
 
             const receipt = await transaction!.wait();
-            console.log("Transaction receipt:", receipt);
-
-            return transaction!.hash;
+            if (receipt.status === 1) {
+                return transaction!.hash;
+            } else {
+                throw new TransferError(
+                    `Error transferring token ${evmToken.tokenId} with transaction hash ${transaction!.hash}`
```
maxisch, line 101: is there always a txHash if the tx failed?

maxisch (approved):

PR #346, opened 2024-01-12
update test command

afeight:
## Description

`vitest` on its own runs in 'watch' mode, so when I tried to run the release script, testing never finished to proceed

## Test plan

ran `yarn test` locally on the root folder and see that the 3 packages all test and complete

### DIFF
--- a/packages/common/base/package.json
+++ b/packages/common/base/package.json
@@ -1,14 +1,12 @@
 {
-    "name": "@crossmint/common-sdk-base",
-    "version": "0.0.2",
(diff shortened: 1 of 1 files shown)


packages/common/base/package.json
```diff
@@ -1,14 +1,12 @@
 {
-    "name": "@crossmint/common-sdk-base",
```
afeight, line 2: prettier ran, only actual change is `"test": "vitest run"`

maxisch (approved):

PR #357, opened 2024-02-07
SCW-361: Updated Fireblocks v11.0.1

jordiamez:
## Description

Using the latest version of fireblocks ncw

## Test plan

Create AAWallet and mint


afeight (approved):
nice

PR #359, opened 2024-02-07
update aa packages

afeight:
## Description

just ran `pnpm i`, not sure why there are updates not previously applied

## Test plan

n/a


afeight:
we use yarn not pnpm lol

PR #361, opened 2024-02-08
PayWithL2: Added 'supportedNetworks' parameter to pay with ETH

angel-paella:
## Description
Similar to https://github.com/Crossmint/crossmint-sdk/pull/358

## Test plan
Tested end to end with https://github.com/Paella-Labs/crossbit-main/pull/11375

### DIFF
--- a/packages/client/base/src/types/embed/crypto.ts
+++ b/packages/client/base/src/types/embed/crypto.ts
@@ -47,6 +47,7 @@ type SolanaTransaction = _SolanaTransaction;
 export type ETHEmbeddedCheckoutSigner = CommonEmbeddedCheckoutSignerProps & {
     signAndSendTransaction: (transaction: EthersTransaction) => Promise<string>;
     network?: Blockchain;
+    supportedNetworks?: Blockchain[];
 };
 
 export type SOLEmbeddedCheckoutSigner = CommonEmbeddedCheckoutSignerProps & {


packages/client/base/src/types/embed/crypto.ts
```diff
@@ -47,6 +47,7 @@ type SolanaTransaction = _SolanaTransaction;
 export type ETHEmbeddedCheckoutSigner = CommonEmbeddedCheckoutSignerProps & {
     signAndSendTransaction: (transaction: EthersTransaction) => Promise<string>;
     network?: Blockchain;
+    supportedNetworks?: Blockchain[];
```
mPaella, line 50: i think you need to update `updatableParams` file also

angel-paella: Mmmmm I think this is not supposed to change. Otherwise we open a bunch of other scenarios, like what happens if they change supportedNetworks and the selected network is not on that list anymore. So it was intentionally made "not updatable". thoughts?

mPaella: oops i meant whatever place the params get serialized into the url, im pretty sure the signer has some custom logic, so not sure if this would get passed through or not

mPaella (approved):

packages/client/base/src/types/embed/crypto.ts
angel-paella: ah! no, the signer is transformed into a JSON.stringify so it gets it directly as it is 🙏

PR #362, opened 2024-02-08
PayWithL2: Added 'switchNetwork' parameter to pay with ETH

angel-paella:
## Description
Following from https://github.com/Crossmint/crossmint-sdk/pull/361 but now for the switchNetwork one

Discussion and context: https://crossmint.slack.com/archives/C06CV64AS6M/p1707339734618149

## Test plan
tested e2e with super crappy UI -> https://www.loom.com/share/019bd8fdd44347349bad24da1bf96b56


packages/client/ui/react-ui/src/components/embed/crypto/CryptoEmbeddedCheckoutIFrame.tsx
```diff
@@ -27,6 +27,12 @@ export default function CryptoEmbeddedCheckoutIFrame(props: CryptoEmbeddedChecko
             console.log("[Crossmint] Received incoming transaction", serializedTransaction);
             handleIncomingTransaction(serializedTransaction);
         }
+
+        if (type === IncomingInternalEvents.CRYPTO_SWITCH_NETWORK) {
+            const { network } = payload;
+            console.log("[Crossmint] Received change of network", network);
+            (signer as ETHEmbeddedCheckoutSigner).switchNetwork!(network);
```
mPaella, line 34:
```suggestion
            (signer as ETHEmbeddedCheckoutSigner).switchNetwork?(network);
```

mPaella: u might want to add a log/error if switch network isnt defined

angel-paella: I meaaaan... Not really. We should only display the Dropdown experience if they send us the switchNetwork function. So if we made it to this point, it should be because the switchNetwork is not undefined. If it is, we wanna crash and know.

If we use "?" instead, if there is a bug the dev will not be receiving the notification, and we will not know. I'd rather have it crash. I can add a explicit check and throw an Error to make it explicit but I generally see those as too verbose...

mPaella: i mean a specific error is much better imo than `cannot read properties of undefined reading blah blah"`

angel-paella: I have changed it, but I'm not really sure it's much more clear. To me, handling the error explicitly makes it seems like it **can happen**, and using the ``!`` makes it clear that it is expected to always exist.

I'll leave it with the error if you insist but I think it's more confusion now >_>

mPaella (approved):

PR #372, opened 2024-02-14
VC: unit tests

vmc-crossmint:
## Description
Added unit tests for VerifiableCredentials SDK

## Test plan
Unit tests


packages/client/verifiable-credentials/src/presentation/getCollections.test.ts
```diff
@@ -0,0 +1,52 @@
...
+});
+
+describe("getCredentialCollections", () => {
+    beforeEach(() => {
+        jest.spyOn(GetNfts, "getWalletNfts").mockResolvedValue([] as any);
+    });
+
+    it("should throw error if chain is not polygon", async () => {
+        await expect(getCredentialCollections("ethereum", "wallet", {}, "environment")).rejects.toThrow(
+            "Only polygon is supported"
```
begona-paella, line 37: can you add other chains?

packages/client/verifiable-credentials/src/presentation/getCollections.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    it("should filter collections by types if filter is provided", async () => {
+        const collections: CredentialsCollection[] = [
+            { metadata: { credentialMetadata: { type: ["type1"] } } } as any,
+            { metadata: { credentialMetadata: { type: ["type2"] } } } as any,
+        ];
+        jest.spyOn(ContactMetadataService.prototype, "getContractWithVCMetadata").mockResolvedValue(collections as any);
+
+        const result = await getCredentialCollections("polygon", "wallet", { types: ["type1"] }, "environment");
+
+        expect(result).toEqual([collections[0]]);
```
begona-paella, line 50: not sure if the test should be placed here, but can we add a test when type is not matching?

packages/client/verifiable-credentials/src/verification/services/nftStatus.test.ts
```diff
@@ -0,0 +1,81 @@
...
+            }
+            if (tokenId === "error") {
+                throw new Error("error");
+            }
+            throw new Error("ERC721: invalid token ID");
+        });
+    });
+
+    it("should return false if NFT is not burnt", async () => {
+        const mockNFT: EVMNFT = {
```
begona-paella, line 27: what happens if the NFT is still minting?

vmc-crossmint: is not possible, you receive a credential only after the nft has been minted

packages/client/verifiable-credentials/src/presentation/getCollections.test.ts
vmc-crossmint: done

kikoncuo (approved):

PR #373, opened 2024-02-15
blockchain: cleanup types

mPaella:
## Description

i need some things like chainIdToBlockchain for wallet connect sdk, which previously only existed in the aa sdk
on top of that, i went ahead and cleaned up basically all the blockchain types/utils in the common sdk

## Test plan
ci


packages/client/wallets/aa/src/CrossmintAASDK.test.ts
```diff
@@ -27,7 +27,7 @@ describe("CrossmintAASDK", () => {
             expect(sdk).toBeInstanceOf(CrossmintAASDK);
         });
     });
-    const allChains = [...EVM_CHAINS, ...BLOCKCHAIN_TEST_NET];
+    const allChains = [...EVM_CHAINS, ...BLOCKCHAIN_TESTNETS];
```
afeight, line 30: :+1:

packages/common/base/src/blockchain/types/index.ts
```diff
@@ -0,0 +1,30 @@
...
+    ...Blockchain,
+    ...BlockchainTestnet,
+} as const;
+export type BlockchainIncludingTestnet = ObjectValues<typeof BlockchainIncludingTestnet>;
+export const BLOCKCHAINS_INCLUDING_TESTNETS = objectValues(BlockchainIncludingTestnet);
+
+export * from "./baseLayers";
+export * from "./evm";
+export * from "./nft";
+export * from "./wallet";
```
afeight, line 30: nit: most `index.ts` files we have are just these, would it make sense to move the code above this into a new file and export them here to match the others?

packages/common/base/src/blockchain/types/nft.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export type NFT = SolanaNFT | EVMNFT | CardanoNFT;
+
+export type NFTOrNFTLocator =
+    | NFT
+    | NFTLocator<"solana">
+    | NFTLocator<"ethereum">
+    | NFTLocator<"polygon">
+    | NFTLocator<"bsc">
+    | NFTLocator<"cardano">;
```
afeight, line 30: is this part actually used? I'm not convinced that it's a good abstraction and wouldn't _yet_ be proud of making it public

packages/common/base/src/blockchain/utils/evm/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from "./blockchainToChainId";
+export * from "./chainIdToBlockchain";
+export * from "./isEVMBlockchain";
```
afeight, line 3: like this ^ re: export `index.ts` comment

packages/common/base/src/blockchain/utils/isBlockchainIncludingTestnets.ts
```diff
@@ -0,0 +1,10 @@
+import { BLOCKCHAINS_INCLUDING_TESTNETS, Blockchain, BlockchainIncludingTestnet } from "../types";
+
+export function isBlockchainIncludingTestnets<T extends BlockchainIncludingTestnet = BlockchainIncludingTestnet>(
+    value: unknown,
+    expectedBlockchain?: T
+): value is T {
```
afeight, line 6: are generics really necessary here? feels like this is the same but simpler
```suggestion
export function isBlockchainIncludingTestnets(value: unknown, expectedBlockchain?: BlockchainIncludingTestnet): value is BlockchainIncludingTestnet {
```

or, is this used anywhere in our SDKs? Might just not want to include this

packages/common/base/src/blockchain/utils/isBlockchainIncludingTestnets.ts
```diff
@@ -0,0 +1,10 @@
...
+    return expectedBlockchain
+        ? value === expectedBlockchain
```
afeight, line 8: how is this function used? we have logic that switches on a param to decide if the entire function amounts to `value === expectedBlockchain`? 
feels wack

packages/common/base/src/types/react/nftDisplay.ts
```diff
@@ -0,0 +1,14 @@
...
+interface CommonProps {
+    uiConfig?: UIConfig;
+    environment?: string;
+}
+export interface NFTCollectionViewProps extends CommonProps {
+    wallets: Wallet[];
+}
+export interface NFTDetailProps extends CommonProps {
+    nft: NFTOrNFTLocator;
+}
```
afeight, line 14: re: `NFTOrNFTLocator` – if this isn't used I'd prefer to omit

packages/common/base/src/blockchain/types/nft.ts
mPaella: it is used, required by the react sdk

packages/common/base/src/blockchain/utils/isBlockchainIncludingTestnets.ts
mPaella: this is taken from crossmint-main

its pretty good imo, bc it allows you to build every single `isEVM, isSolana, isSOMECHAIN` all from this single function

you can see how its used in `isEVMBlockchain`

packages/common/base/src/types/react/nftDisplay.ts
mPaella: unfortunately we cant remove it

packages/common/base/src/blockchain/types/index.ts
mPaella: i mostly agree, but what name, like `/blockchain/types/blockchain.ts` or `/blockchain/types/chains.ts` feels a bit weird

not a strong opinion tho

packages/common/base/src/blockchain/utils/isBlockchainIncludingTestnets.ts
mPaella: which i think is better than this

from aa sdk
image

packages/common/base/src/blockchain/utils/isBlockchainIncludingTestnets.ts
mPaella: see below, if we keep it as the building block for all the `isSOMECHAIN` checks, then it should keep the generic

packages/common/base/src/blockchain/types/nft.ts
mPaella: its been public for probably a year lol

afeight (approved):

PR #375, opened 2024-02-15
dev ex: add dev scripts + cross-env

mPaella:
## Description

adds dev (watch), scripts which auto rebuild when you save changes

adds cross-env so aa test script works on windows/all platforms

## Test plan
tested dev scripts work
ci


MatiNavCrossmint (commented):
Amazing, thanks for all this useful tools!

I have one small comment though, can we omit the `--minify` for dev builds? The reason is that using it will make impossible to debug it locally

MatiNavCrossmint (approved):
LGTM!

PR #376, opened 2024-02-15
version: common-sdk-base 0.0.4

mPaella:
## Description

release 0.0.4 of common base

## Test plan

ci

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1607,6 +1607,14 @@
     exponential-backoff "3.1.1"
     uuid "9.0.1"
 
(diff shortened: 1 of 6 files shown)


yarn.lock
```diff
@@ -1607,6 +1607,14 @@
     exponential-backoff "3.1.1"
     uuid "9.0.1"
 
+"@crossmint/common-sdk-base@0.0.3":
```
mPaella, line 1610: this is still here cuz i didnt actually release the other packages that depend on the common pkg, the change is just locally

PR #379, opened 2024-02-16
SCW-346 Migrate zerodev to 5.0.0

jordiamez:
## Description

Upgrade Zerodev to v5.1.0

## Test plan

Create AA


jordiamez:
This PR is ready for review. V1 using only web3auth. I'm working on another PR to allow the ethers signer

packages/client/wallets/aa/src/CrossmintAASDK.ts
```diff
@@ -68,24 +76,24 @@ export class CrossmintAASDK {
...
+            const account = await createKernelAccount(publicClient, {
+                plugins: {
+                    sudo: ecdsaValidator,
                 },
             });
 
-            const evmAAWallet = new EVMAAWallet(zDevProvider, this.crossmintService, chain);
+            if (!isEVMBlockchain(chain)) {
+                throw new WalletSdkError(`The blockchain ${chain} is still not supported`);
+            }
```
afeight, line 93: shouldn't we do this much earlier?

afeight: actually I think we already have this code here: https://github.com/Crossmint/crossmint-sdk/pull/379/files#diff-44d9925b3eab949d3dc6cce03789d9160cfa1d1a34305326e05e39be67f70920R60-R62

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -1,80 +1,65 @@
...
+    private crossmintService: CrossmintWalletService;
+    private publicClient: PublicClient;
+    private ecdsaValidator: KernelValidator<"ECDSAValidator">;
+    kernelClient: KernelAccountClient;
+    account: KernelSmartAccount;
     chain: B;
 
-    constructor(provider: ZeroDevEthersProvider<"ECDSA">, crossmintService: CrossmintWalletService, chain: B) {
-        super(provider, crossmintService);
+    constructor(
```
afeight, line 33: let's mark this as `private` – right now this is considered a breaking change but clients shouldn't be calling this directly, so let's make sure of that

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -1,80 +1,65 @@
...
-            signer: await this.getAddress(),
-            message,
-            signature,
-        });
+            },
+        }) as KernelAccountClient;
     }
 
-    setSessionKeySignerAddress(sessionKeySignerAddress: string) {
+    setSessionKeySignerAddress(sessionKeySignerAddress: Hex) {
```
afeight, line 62: same thought–should a client call this directly? Or should we mark as `private`?

afeight: same thought–should a client call this directly? Or should we mark as `private`?
```suggestion
    private setSessionKeySignerAddress(sessionKeySignerAddress: Hex) {
```

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -1,80 +1,65 @@
...
+        account: KernelSmartAccount,
+        crossmintService: CrossmintWalletService,
+        chain: B,
+        publicClient: PublicClient,
+        ecdsaValidator: KernelValidator<"ECDSAValidator">
+    ) {
         this.chain = chain;
-    }
-
-    async getSigner<Type extends SignerType = SignerType>(type: Type): Promise<SignerMap[Type]> {
```
afeight, line 41: we still need this function–did it move?

packages/client/wallets/aa/src/api/CrossmintWalletService.ts
```diff
@@ -30,71 +30,6 @@ export class CrossmintWalletService extends BaseCrossmintService {
...
-    }
-
-    async getNCWIdentifier(deviceId: string) {
-        return this.fetchCrossmintAPI(
-            `unstable/wallets/aa/ncw?deviceId=${deviceId}`,
-            { method: "GET" },
-            `Error getting NCW identifier for device: ${deviceId}`
-        );
-    }
-
```
afeight, line 97: we're removing this because we don't support fireblocks anymore? We can't just remove this functionality–if we need to add it back in a separate PR then let's make sure that PR exists first but can't merge this in removing Fireblocks

packages/client/wallets/aa/src/utils/signer.ts
```diff
@@ -27,70 +27,39 @@ type CreateOwnerSignerInput = {
 };
 
 export async function createOwnerSigner({
-    userIdentifier,
-    projectId,
     chain,
     walletConfig,
-    crossmintService,
```
afeight, line 34: if we only use `chain` and `walletConfig` now, then let's update the definition for `CreateOwnerSignerInput` to only those params

packages/client/wallets/aa/src/utils/signer.ts
```diff
@@ -27,70 +27,39 @@ type CreateOwnerSignerInput = {
...
-            ticker: getTickerByBlockchain(chain),
-            tickerName: getTickerNameByBlockchain(chain),
-        };
-        const web3auth = new Web3Auth({
-            clientId: signer.clientId,
-            web3AuthNetwork: signer.web3AuthNetwork,
-            usePnPKey: false,
-        });
+}: CreateOwnerSignerInput): Promise<SmartAccountSigner<"custom", Address>> {
+    const signer = walletConfig.signer as Web3AuthSigner;
```
afeight, line 33: still need to handle fireblocks

packages/client/wallets/aa/src/types/Config.ts
```diff
@@ -23,27 +22,15 @@ export type UserIdentifier =
     | { type: "email"; email: string }
     | { type: "phoneNumber"; phoneNumber: string };
 
-export type SignerType = "ethers" | "viem";
```
afeight, line 26: need to be added back, can be second PR

afeight (changes requested):
I see that in this PR we dropped support for:
1. fireblocks
2. provide ethers signer as EOA
3. `getSigner("ethers")`
maybe more? 

We will need these to remain, so we can make a follow-up PR that builds on top of this one, but we'll need to merge it all at once so we don't errantly make a release that breaks functionality

packages/client/wallets/aa/src/CrossmintAASDK.ts
```diff
@@ -60,32 +63,31 @@ export class CrossmintAASDK {
...
+                plugins: {
+                    sudo: ecdsaValidator,
                 },
             });
 
-            const evmAAWallet = new EVMAAWallet(zDevProvider, this.crossmintService, chain);
+            const evmAAWallet = new EVMAAWallet(account, this.crossmintService, chain, publicClient, ecdsaValidator);
 
-            const abstractAddress = await evmAAWallet.getAddress();
+            const abstractAddress = evmAAWallet.kernelClient.account!.address;
```
afeight, line 90: can we remove the `!`?
```suggestion
            const abstractAddress = evmAAWallet.kernelClient.account.address;
```

packages/client/wallets/aa/src/utils/signer.ts
```diff
@@ -81,16 +55,23 @@ export async function createOwnerSigner({
             idToken: signer.jwt,
         });
 
-        return getRPCProviderOwner(provider);
+        const web3 = new Web3(provider as any);
+        const [address] = await web3.eth.getAccounts();
+
+        return await providerToSmartAccountSigner(provider! as EIP1193Provider, address as Hex);
```
afeight, line 61: we're using `as any` and `!` to circumvent better `null` handling. 
Can we add
```suggestion
        if (provider == null) {
            throw new WalletSdkError("web3auth returned a null signer")
        }
        
        const web3 = new Web3(provider);
        const [address] = await web3.eth.getAccounts();
        return await providerToSmartAccountSigner(provider as EIP1193Provider, address as Hex);
```

packages/client/wallets/aa/src/utils/signer.ts
```diff
@@ -81,16 +55,23 @@ export async function createOwnerSigner({
...
+    } else if (isEIP1193Provider(walletConfig.signer)) {
+        const web3 = new Web3(walletConfig.signer as any);
```
afeight, line 63: not needed
```suggestion
        const web3 = new Web3(walletConfig.signer);
```

packages/client/wallets/aa/src/blockchain/BlockchainNetworks.ts
```diff
@@ -236,3 +238,11 @@ export function getViemNetwork(networkName: EVMBlockchainIncludingTestnet) {
             throw new Error(`Unsupported network: ${networkName}`);
```
MatiNavCrossmint, line 238: If we will use only `viem`, then will we lose support for `astar-zkevm`?

MatiNavCrossmint: If so, I propose:
- Refactor `getOrCreateWallet<B extends EVMBlockchainIncludingTestnet = EVMBlockchainIncludingTestnet>` to a class that can receive a type of actually supported chains
- Put a check at the top of the function, asserting that

jordiamez: Added

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -1,62 +1,218 @@
...
+        try {
+            return this.kernelClient.account.address;
+        } catch (error) {
+            logError("[GET_ADDRESS] - ERROR", {
+                service: SCW_SERVICE,
+                error: errorToJSON(error),
+                signer: this.kernelClient.account,
+            });
+            throw new Error(`Error getting address. If this error persists, please contact support.`);
+        }
```
afeight, line 85: how can this fail? it looks like just property access

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -1,62 +1,218 @@
...
+                    account: this.account,
+                    address: contractAddress,
+                    abi: erc721,
+                    functionName: "safeTransferFrom",
+                    args: [this.getAddress(), toAddress, evmToken.tokenId],
+                });
+                transaction = await publicClient.writeContract(request);
+            }
+
+            if (transaction !== undefined) {
```
afeight, line 186: slightly safer
```suggestion
            if (transaction != null) {
```

packages/client/wallets/aa/src/types/Config.ts
```diff
@@ -43,7 +47,7 @@ export type FireblocksNCWSigner =
           deviceId: string;
       });
 
-type Signer = FireblocksNCWSigner | ethers.Signer | Web3AuthSigner;
+type Signer = FireblocksNCWSigner | EIP1193Provider | Web3AuthSigner;
```
afeight, line 50: is fireblocks still supported? Looks like we dropped it [here](https://github.com/Crossmint/crossmint-sdk/pull/379/files#diff-d4de4d77c0ae39dbb040b4fa7d344b1d1522fdb13ca57007c6cf2b123081770aL36) which is totally fine, let's just remove this option for now if it's not supported

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -150,76 +283,15 @@ export class EVMAAWallet<B extends EVMBlockchainIncludingTestnet = EVMBlockchain
         }
     }
 
-    async setCustodianForKillswitch(custodian?: Custodian | undefined) {
```
afeight, line 153: just checking–doing killswitch is still supported, but in a different way?

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
afeight, line 85: is this possible?
```suggestion
        return this.kernelClient.account.address;
```

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
jordigiam: is not supported in this version

jordiamez: it's not supported in this version

PR #381, opened 2024-02-17
VC: override usage

vmc-crossmint:
## Description
Build process may fail if the usage origin is wrongly determined.

## Test plan
Tested manually


packages/client/verifiable-credentials/src/services/crossmintAPI.ts
```diff
@@ -22,11 +22,11 @@ export class CrossmintAPI {
...
             "https://nftstorage.link/ipfs/{cid}",
-        ]
+        ],
+        overrideUsageOrigin?: "client" | "server"
     ) {
-        const usageOrigin = getUsageOrigin();
-
-        if (usageOrigin != null) {
+        const usageOrigin = overrideUsageOrigin ?? getUsageOrigin();
+        if (usageOrigin) {
```
alfonso-paella, line 29:
```suggestion
        if (usageOrigin != null) {
```

PR #383, opened 2024-02-19
SCW-344: Revert "SCW-344: Fix 'wallet is not defined' error"

lluiserena:
## Description

Revert SDK modification. SDK, as a front package, should have the DatadogLogger provider. Solution implemented on Crossbit 

## Test plan

Test WC and CC using SDK current version


MatiNavCrossmint (approved):
LGTM!

PR #384, opened 2024-02-20
wc: reusable error + unsupported chains modal + #385 & #387

mPaella:
## Description

if we cant support the required chains, show this non-recoverable modal

## Test plan

locally with wc demo

image


packages/common/base/src/blockchain/utils/blockchainToCopyName.ts
```diff
@@ -0,0 +1,24 @@
+import { BlockchainIncludingTestnet } from "../types";
+
+export const BLOCKCHAIN_TO_COPY_NAME: Record<BlockchainIncludingTestnet, string> = {
```
mPaella, line 3: moved/improved from aa package, needs to be in common here

packages/client/wallets/walletconnect/src/utils/wallet/supportsRequiredMethods.ts
```diff
@@ -0,0 +1,26 @@
+import { BuildApprovedNamespacesParams } from "@walletconnect/utils";
+import { Web3WalletTypes } from "@walletconnect/web3wallet";
+
+export function supportsRequiredMethods(
```
mPaella, line 4: need this for later if an unsupported method is requested

PR #386, opened 2024-02-20
wc: mock required namespace methods

mPaella:
## Description

mock the required namespace methods, if the wallet doesnt support them

this improves compatibility, because we can "get in the door" (aka connect, maybe sign a message) with any website - and later if the site tries to use an unsupported method, we will show an "Unsupported method requested" modal

not implemented, modal coming later
image

## Test plan

mocks required methods correctly
image

### DIFF
--- a/packages/client/wallets/walletconnect/src/utils/walletconnect/mockRequiredNamespaceMethods.ts
+++ b/packages/client/wallets/walletconnect/src/utils/walletconnect/mockRequiredNamespaceMethods.ts
@@ -1,0 +1,29 @@
+import { ProposalTypes } from "@walletconnect/types";
+import { BuildApprovedNamespacesParams } from "@walletconnect/utils";
+
+export function mockRequiredNamespaceMethods(
+    requiredNamespaces: ProposalTypes.RequiredNamespaces,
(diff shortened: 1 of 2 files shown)


packages/client/wallets/walletconnect/src/utils/walletconnect/mockRequiredNamespaceMethods.ts
```diff
@@ -0,0 +1,29 @@
...
+    Object.entries(requiredNamespaces).forEach(([caipKey, requiredNamespace]) => {
+        if (!mockedNamespaces[caipKey]) {
+            return;
+        }
+
+        const unsupportedRequiredMethods = requiredNamespace.methods.filter(
+            (method) => !mockedNamespaces[caipKey].methods?.includes(method)
+        );
+        console.warn(
+            "[mockRequiredNamespaceMethods] Mocking unsupported required methods:",
```
alfonso-paella, line 19: As a dev could see this, I suggest rewording to be more clear to:

`Mocking unsupported required methods: ${methods}. Calling these methods will result in a no-op.`

alfonso-paella (approved):

PR #391, opened 2024-02-22
SCW-384: Create updated demo site for testing

lluiserena:
## Description

Created a next frontend inside the AA package to do the manual testing and start the integration of Playwright to test.

## Test plan

On `packages/client/wallets/aa/front` use the command `yarn dev` to open the frontend on `localhost:3001`


packages/client/wallets/aa/front/services/walletService.ts
```diff
@@ -0,0 +1,37 @@
...
+export const getOrCreateWallet = async () => {
+    const { CrossmintAASDK } = await import("@crossmint/client-sdk-aa");
+    const xm = CrossmintAASDK.init({
+        apiKey: process.env.NEXT_PUBLIC_API_KEY_STG || '',
+    });
+    const signer: ethers.Wallet = new ethers.Wallet("01f4f4a0efe13dbeaff3e2a030f0fa1754e2d90b1391a0120a6a11c1bf10ea6a");
+    const walletInitParams = {
+        signer
+    }
+    const userIdentifier = { email:"_test@test.com" };
```
MatiNavCrossmint, line 16: This should be kind of dynamic / or received by input. If not, we will execute the creation flow the first time, and then always the retrieve one

MatiNavCrossmint (commented):
Overall It looks good.

What we need to add, is having both buttons:
- "Execute getOrCreate with same email"
- "Execute getOrCreateWithNewEmail"

With that, the former should test the creation flow, the latter the retrieve, and also we will be sure that we are testing if the solution identifies wether to execute one or the other, successfully

packages/client/wallets/aa/front/services/walletService.ts
lluiserena: **Added two inputs** to add the `email` and the `private key` to create the wallet using `new ethers.Wallet(privateKey);`

packages/client/wallets/aa/front/pages/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+                <input
+                    className={styles.input}
+                    placeholder="Private key"
+                    value={privateKey}
+                    onChange={(e) => setPrivateKey(e.target.value)}
+                />
+                <button className={styles.button} onClick={handleCreateWallet}>
+                    Create wallet
+                </button>
+                <input className={styles.input} type="text" value={walletAddress} readOnly />
```
MatiNavCrossmint, line 88: Is the `readinly` attribute ok here or will it make impossible to change the input's value?

MatiNavCrossmint (commented):
Awesome, just 1 comment

packages/client/wallets/aa/front/pages/index.tsx
lluiserena: we won't want to change the walletAddress, we retrieve it from the getOrCreateWalletProcess and we show the user the address.

MatiNavCrossmint (approved):
LGTM!

PR #393, opened 2024-02-26
SCW-382: Set up monitoring approach

lluiserena:
## Description

Added Datadog logs on every catch. Also, added a 'service' property to ease the filtering on Datadog 

## Test plan

Filter and search logs on Datadog.


packages/client/wallets/aa/src/CrossmintAASDK.ts
```diff
@@ -34,6 +34,7 @@ export class CrossmintAASDK {
     ) {
         try {
             logInfo("[GET_OR_CREATE_WALLET] - INIT", {
+                service: SCW_SERVICE,
```
MatiNavCrossmint, line 37: cool

MatiNavCrossmint (approved):
This will help us a lot, thank you for handling it!

PR #396, opened 2024-02-27
UItests: adding first ui tests for local demo

begona-paella:

### DIFF
--- a/packages/client/wallets/aa/playwright/tests/functions/aa-wallets.ts
+++ b/packages/client/wallets/aa/playwright/tests/functions/aa-wallets.ts
@@ -1,0 +1,10 @@
+import { Page, expect, request } from "@playwright/test";
+
+export async function getWallet(page: Page,) {
+    await page.goto("/");
+    await page.getByPlaceholder('Email').fill('begona+testgetwallet@paella.dev');
+    await page.getByPlaceholder('Private key').fill('274c683c7aaa921afb22b49ddaf3395d9f61f72245fd3e1c4885b39caff12ec0');
+    await page.getByRole('button', { name: 'Create wallet' }).click();
+    const locator = page.getByTestId('createdOrGotWallet');
+    await expect(locator).toHaveValue("0xbA23099994403e1e1A4Fe569ab2A4B06f59d047d", { timeout: 10_000 });
+}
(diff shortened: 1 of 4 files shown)


lluiserena (approved):
Looks good to me.

packages/client/wallets/aa/playwright/tests/functions/aa-wallets.ts
```diff
@@ -0,0 +1,10 @@
+import { Page, expect } from "@playwright/test";
+
+export async function getWallet(page: Page,) {
+    await page.goto("/");
+    await page.getByPlaceholder('Email').fill('begona+testgetwallet@paella.dev');
+    await page.getByPlaceholder('Private key').fill('274c683c7aaa921afb22b49ddaf3395d9f61f72245fd3e1c4885b39caff12ec0');
```
MatiNavCrossmint, line 6: I know that the front has an input for introducing a `privateKey`, but that's not used by the developer when implementing AA. So I'm curious about it, why it's used, and its impact on the tests

packages/client/wallets/aa/playwright/tests/functions/aa-wallets.ts
```diff
@@ -0,0 +1,10 @@
+import { Page, expect } from "@playwright/test";
+
+export async function getWallet(page: Page,) {
+    await page.goto("/");
+    await page.getByPlaceholder('Email').fill('begona+testgetwallet@paella.dev');
```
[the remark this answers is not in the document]
begona-paella: yes, for this first round we are keeping simple and getting the wallet.
For the next iteration, we will include a test for creating a wallet each execution. We will use fireblocks and a new email each time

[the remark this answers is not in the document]
MatiNavCrossmint: Makes total sense!

packages/client/wallets/aa/playwright/tests/functions/aa-wallets.ts
lluiserena: For create an ethers Wallet using the ethers lib you need the pk. The users of course will use the signer itself, but in the demo we need to instantiate the signer.

MatiNavCrossmint: I see, thanks for the explanation. It leads me to this questions:
- I see us instantiating a custom signer, but to test all the paths, we will need to switch between all possible ones ( 
`type Signer = FireblocksNCWSigner | ethers.Signer | Web3AuthSigner;` ) and that isn't happening, right? 
  - Is that because this is a v0?

PR #398, opened 2024-02-28
Synced base with crossbit main blockchains

maxisch:
## Description

Sync chains with crossmint
The build fails because the AA package doesn't build. But I need this changes release to update the AA package, which I do in https://github.com/Crossmint/crossmint-sdk/pull/399
It's a circular dependency, that's why I need this merged as is.

## Test plan

Manual

### DIFF
--- a/packages/common/base/src/blockchain/types/evm.ts
+++ b/packages/common/base/src/blockchain/types/evm.ts
@@ -15,10 +15,17 @@ export type EVMBlockchain = ObjectValues<typeof EVMBlockchain>;
 export const EVM_CHAINS = objectValues(EVMBlockchain);
 
 export const EVMBlockchainTestnet = {
-    GOERLI: "goerli",
+    ARBITRUM_SEPOLIA: "arbitrum-sepolia",
+    BASE_GOERLI: "base-goerli",
+    BASE_SEPOLIA: "base-sepolia",
+    BSC_TESTNET: "bsc-testnet",
+    ETHEREUM_GOERLI: "ethereum-goerli",
     ETHEREUM_SEPOLIA: "ethereum-sepolia",
-    MUMBAI: "mumbai",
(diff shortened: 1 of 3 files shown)


packages/common/base/src/blockchain/types/evm.ts
```diff
@@ -15,10 +15,17 @@ export type EVMBlockchain = ObjectValues<typeof EVMBlockchain>;
...
 export const EVMBlockchainTestnet = {
-    GOERLI: "goerli",
+    ARBITRUM_SEPOLIA: "arbitrum-sepolia",
+    BASE_GOERLI: "base-goerli",
+    BASE_SEPOLIA: "base-sepolia",
+    BSC_TESTNET: "bsc-testnet",
+    ETHEREUM_GOERLI: "ethereum-goerli",
     ETHEREUM_SEPOLIA: "ethereum-sepolia",
-    MUMBAI: "mumbai",
-    ZKATANA: "zkatana",
```
afeight, line 21: did zkatana get dropped?

maxisch: sadly it's here, as a mainnet:
```
export const EVMBlockchain = {
    ETHEREUM: "ethereum",
    POLYGON: "polygon",
    BSC: "bsc",
    OPTIMISM: "optimism",
    ARBITRUM: "arbitrum",
    BASE: "base",
    ZORA: "zora",
    ARBITRUMNOVA: "arbitrumnova",
    ZKATANA: "zkatana",
} as const;
```
I raised the concern here: https://crossmint.slack.com/archives/C06LNUL04TH/p1709145570085719
But that's how it is right now in crossbit-main. This work is just to sync main and sdk. Any changes we would like to do would be have to be in a follow up.

afeight (approved):

PR #401, opened 2024-02-29
SCW-391:Improve SDK Demo site

lluiserena:
## Description

Improvements on the SDK AA Demo to do testing with Playwright

-Create / Get Wallet using Fireblocks
-Input text for Sign Message
-Sign Typed data
-Verify message

## Test plan

Run  `yarn dev` on `packages/client/wallets/aa/front/`

### DIFF
--- a/packages/client/wallets/aa/front/pages/index.tsx
+++ b/packages/client/wallets/aa/front/pages/index.tsx
@@ -1,9 +1,12 @@
 import {
-    getOrCreateWallet,
+    getOrCreateWalletEthers,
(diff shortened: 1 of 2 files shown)


MatiNavCrossmint (approved):
I like was being built here!

Just a warning, the index file is growing a lot and it already has almost 300 lines. So, I'd say that if we want to keep adding functionality to it, we need to split the front between pages/components ( that's up to you how to handle it ). Ideally, each component should be focused on specific responsibilities, so it makes it easier to understand for someone who touches this for first time

PR #406, opened 2024-03-01
fix: lock and sync deps

mPaella:
## Description

no reason to not sync these deps with the latest versions, due to syncing yarn.lock doesnt update much - if didnt sync, would be a couple hundred lines

## Test plan

ci passes, `yarn install` doesnt update lock file

### DIFF
--- a/packages/client/wallets/aa/front/package.json
+++ b/packages/client/wallets/aa/front/package.json
@@ -1,25 +1,25 @@
 {
-  "name": "front",
-  "version": "0.1.0",
-  "private": true,
-  "scripts": {
-    "dev": "next dev -p 3001",
-    "build": "next build",
-    "start": "next start",
-    "lint": "next lint"
-  },
-  "dependencies": {
-    "ethers": "5.7.2",
-    "next": "13.5.6",
-    "react": "^18",
-    "react-dom": "^18",
-    "@crossmint/client-sdk-aa": "file: ../../../",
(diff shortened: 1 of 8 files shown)


packages/client/wallets/aa/front/package.json
```diff
@@ -1,25 +1,25 @@
...
+        "build": "next build",
+        "start": "next start",
+        "lint": "next lint"
+    },
+    "dependencies": {
+        "ethers": "5.7.2",
+        "next": "13.5.6",
+        "react": "^18",
+        "react-dom": "^18",
+        "@crossmint/client-sdk-aa": "0.2.0",
```
mPaella, line 16: "file" imports should never be used

maxisch (approved):

packages/client/wallets/aa/front/package.json
afeight: doesn't this mean that CI is useless?
If we update the package but the frontend is still running against 0.2.0, then it won't go red
so you first have to release the broken version, have CI go red, then fix it?

packages/client/wallets/aa/front/package.json
```diff
@@ -1,25 +1,25 @@
...
+        "@crossmint/common-sdk-base": "0.0.5"
```
afeight, line 17: this one though for sure should not be file

afeight (commented):
just one concern

PR #413, opened 2024-03-05
Use our own rpc urls instead of zerodevs default (infura)

maxisch:
## Description

ZeroDev depends on infura. This adds another layer of dependencies which we don't control. We submit our own rpcs now.
ZeroDev's code:
image
where `CHAIN_ID_TO_NODE` looks like this (and more):
image

## Test plan

Manual Testing creating and retrieving wallets.


afeight (approved):
sweet

PR #415, opened 2024-03-06
SCW-403: Get policies by address endpoint call

lluiserena:
## Description

Added new call to a new crossbit endpoint that returns the list of policies for the address of the AA on a specific chain. 
Response example
```
[
"send_transaction"
]
```

This has been doing to be used on the AA SDK tests, to complete the paymaster test plan.

## Test plan

Instantiate the _CrossmintWalletService_ and call _getWalletPolicies_


packages/client/wallets/aa/src/api/CrossmintWalletService.ts
```diff
@@ -124,4 +124,12 @@ export class CrossmintWalletService extends BaseCrossmintService {
             `Error fetching NFTs for wallet: ${address}`
         );
     }
+
+    async getWalletPolicies(address: string, chain: string) {
```
afeight, line 128: it's hard to tell what should be a part of the SDK and what's just for us–does this function have any use for a client?
If not, we can make a subclass of `CrossmintWalletService` that we use in the test site where we add this, but it shouldn't go into the SDK

lluiserena: Only will be used for testing. Moved to the 'front' package in a subclass that extends the CrossmintWalletService.

PR #416, opened 2024-03-06
WC-AA: Add workaround to support odd values on sendTransaction

MatiNavCrossmint:
## Description

/**
 * We need to support odd value and data fields in a tx, as that data comes from WC
 * ZeroDev implements hexlify, but doesn't support to send an odd string in value or data fields
 * so we need to make the hexlify fix manually when sending a transaction.
 * That fix can be found here:
 * https://github.com/ethers-io/ethers.js/commit/a12030ad29aa13c02aa75d9e0860f4986a0043b4#diff-047e7ebfdbd5c41e762cda03593bd15ed8b3121dece67262729dfcbde7040818R221
 * And more context on this issue:
 * https://github.com/ethers-io/ethers.js/issues/614
 *
 * Applying to it hexValue is used as a workaround on Crossbit, but doesn't cover all the cases (i.e. 0x71afd498d0000).
 * This function is the same as hexValue but without hexStripZeros, which is what makes it to not work.
 */

## Test plan

Described in this PR: https://github.com/Paella-Labs/crossbit-main/pull/11826

- Swap through Uniswap


afeight (approved):
nice!

PR #427, opened 2024-03-11
api key: env to baseURL + folder struct changes

mPaella:
## Description

adds `environmentToCrossmintBaseURL`, and changes folder structure slightly moving apiKey folder outside of utils

we have multiple different implementations of this similar function (aa sdk, xm connect, client-sdk but doesnt use api keys so its not exactly the same), so hopefully eventually we can converge to only this one

## Test plan

ci


packages/common/base/src/apiKey/utils/environmentToCrossmintBaseURL.ts
```diff
@@ -0,0 +1,15 @@
...
+    environment: APIKeyEnvironmentPrefix,
+    customMap?: Partial<Record<APIKeyEnvironmentPrefix, string>>
+): string {
+    switch (environment) {
+        case "development":
+            return customMap?.development || "http://localhost:3000";
+        case "staging":
+            return customMap?.staging || "https://staging.crossmint.com";
+        case "production":
+            return customMap?.production || "https://crossmint.com";
```
mPaella, line 13:
```suggestion
            return customMap?.production || "https://www.crossmint.com";
```

PR #437, opened 2024-03-18
WC SDK: Fix library imported from base

MatiNavCrossmint:
## Description

We were importing `blockchainToCopyName`, when the actual function was `blockchainToDisplayName`.

Also, started adding `yalc` commands to make our life easier when connecting packages locally. This only Impacts locally when running the app 

## Test plan

Make a successful connection using WC SDK.Screenshot 2024-03-18 at 12 40 06 PM


afeight (approved):
nice!

PR #440, opened 2024-03-19
Debug flag

guillermodoghel:
## Description
Adds a debug flag for the event console

## Test plan
https://www.loom.com/share/969ff9e7516448eca28207f650860455?sid=5398d991-f10f-493b-bab3-00b1782f733d


alfonso-paella:
Not sure where the implementation of this mode is, but I'd pls ensure this only works if the environment is NOT production.

I think some devs will fuck up and have client side logic that checks for purchase:succeeded and deliver goods and stuff like that... so best to not make it easy for their users to exploit that

PR #454, opened 2024-04-08
SCW-461: Replace Mumbai with Amoy

lluiserena:
## Description

Replaced `'polygon-mumbai'` with `'polygon-amoy'` on AA SDK.

## Test plan

Use the AA SDK with `polygon-amoy`


packages/client/wallets/aa/front/package.json
```diff
@@ -9,8 +9,8 @@
         "lint": "next lint"
     },
     "dependencies": {
-        "@crossmint/client-sdk-aa": "0.2.1",
-        "@crossmint/common-sdk-base": "0.0.6",
+        "@crossmint/client-sdk-aa": "file: ../../../",
+        "@crossmint/common-sdk-base": "file: ../../../../../../common/base",
```
afeight, line 13: we'll have to switch this back–do you need any SDK releases?

packages/client/wallets/aa/src/utils/helpers.ts
```diff
@@ -18,8 +18,8 @@ export function transformBackwardsCompatibleChains<B extends BlockchainIncluding
     chain: B | BackwardsCompatibleChains
 ): B {
     switch (chain) {
-        case "mumbai":
-            return "polygon-mumbai" as B;
+        case "amoy":
+            return "polygon-amoy" as B;
```
afeight, line 22: I don't think we need this, right?
We had this because previously we used `mumbai` but converted it later to `"polygon-mumbai"`

Basically I think we can remove this case entirely
```suggestion
```

afeight (commented):
this all looks good! Just some minor comments

I see changes to base SDK as well as AA SDK, so we may have to split this up into separate PRs so that we can merge & release base before updating AA SDK

packages/client/wallets/aa/front/package.json
lluiserena: The front demo was done to test the changes locally before push them. The problem rn is that the frontend demo is not compatible with v.5.0.0.

I'll switch this back to have it working so we can merge it, but we need to keep it in mind.

lluiserena:
> this all looks good! Just some minor comments
> 
> I see changes to base SDK as well as AA SDK, so we may have to split this up into separate PRs so that we can merge & release base before updating AA SDK

I have updated the PR and created a new one for the client/base package changes https://github.com/Crossmint/crossmint-sdk/pull/455

MatiNavCrossmint (approved):
LGTM!

PR #467, opened 2024-04-16
VC: credential from locator

vmc-crossmint:
## Description
- Allowing to retrieve a nft from a locator 
- Added some utils like 'isEncryptedCrential'.
- Minimal refactor

## Test plan
Unit test

### DIFF
--- a/packages/client/verifiable-credentials/src/presentation/getMetadata.ts
+++ b/packages/client/verifiable-credentials/src/presentation/getMetadata.ts
@@ -1,7 +1,7 @@
-import { CrossmintAPI } from "@/services/crossmintAPI";
 import { ethers, utils } from "ethers";
 
 import { nftZeroEight } from "../ABI/upgradeable721-v0.8";
+import { CrossmintAPI } from "../services/crossmintAPI";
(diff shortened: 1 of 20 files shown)


packages/client/verifiable-credentials/src/presentation/getMetadata.ts
```diff
@@ -24,11 +23,33 @@ export class ContactMetadataService {
         }
         console.debug(`Found contract ${uri} for contract ${contractAddress}`);
 
+        return this.getFromIpfs(uri);
+    }
+
+    async getContractWithVCMetadata(
```
alberto-crossmint, line 29: I find a bit of inconsistent the usage of Collection and Contract sometimes referring to the same entity. Do we have a guideling on when to use which? Why Contract here and not Collection when the return type is Collection?

alberto-crossmint (approved):

packages/client/verifiable-credentials/src/presentation/getMetadata.ts
vmc-crossmint: Agreed, will be addressed in next refactor

PR #470, opened 2024-04-23
SCW346: Set custodian for tokens

lluiserena:
## Description

Implemented SetCustodianForTokens with the new ZeroDev SDK version. 

## Test plan

Call 'setCustodianForTokens' when using the AA SDK.

### DIFF
--- a/packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
+++ b/packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
@@ -3,10 +3,7 @@ import { GenerateSignatureDataInput, SignerMap, SignerType } from "@/types";
 import {
     SCW_SERVICE,
     TransactionError,
-    convertData,
-    decorateSendTransactionData,
     errorToJSON,
-    getNonce,
 } from "@/utils";
 import type { Deferrable } from "@ethersproject/properties";
 import { type TransactionRequest } from "@ethersproject/providers";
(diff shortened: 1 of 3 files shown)


packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -238,28 +236,33 @@ export class EVMAAWallet<B extends EVMBlockchainIncludingTestnet = EVMBlockchain
...
+                entryPoint: ENTRYPOINT_ADDRESS_V07,
                 plugins: {
                     sudo: this.ecdsaValidator,
                     regular: sessionKeyValidator,
                 },
             });
-            const serializedSessionKeyAccount = await serializeSessionKeyAccount(sessionKeyAccount);
+            console.log({sessionKeyAccount})
+            const serializedSessionKeyAccount = await serializePermissionAccount(sessionKeyAccount);
+            console.log({serializedSessionKeyAccount})
```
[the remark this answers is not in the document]
lluiserena: Yeah sorry, forgot to delete them. 
Deleted!

afeight (approved):
looks good!

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -262,28 +261,31 @@ export class EVMAAWallet<B extends EVMBlockchainIncludingTestnet = EVMBlockchain
...
-                validatorData: {
-                    paymaster: oneAddress,
-                },
-            });
+              })
 
+            const sessionKeyValidator = await toPermissionValidator(this.publicClient, {
+                entryPoint: ENTRYPOINT_ADDRESS_V07,
+                signer: sessionKeySigner,
+                policies: [],
```
MatiNavCrossmint, line 274: I'm curious about leaving policies as an empty array. 

From my understanding, we want to get full control through the session key right? If so, I understand the way to go here as the [sudo policy ](https://docs.zerodev.app/sdk/permissions/policies/sudo). 

WDYT @lluiserena ?

afeight: I don't think we want sudo policy because we cannot have access to transfer fungible tokens from crossmint.com
but that does bring up the question–does this empty policy allow that still? Or do we need to add policy based on 721/1155 functions?

PR #472, opened 2024-04-24
AA: Fix Base RPC

MatiNavCrossmint:
## Description

Llama RPC caused a `net_version` error. Changing the url resolves it

## Test plan

Created AA locally


afeight (approved):
yes plz

PR #436, opened 2024-03-18
add Blockchain alias [Fixes SCW-454]

afeight:
## Description

Get us back to the agreed upon API surface

## Test plan

Linked locally and could do 
```
import { Blockchain } from '@crossmint/client-sdk-aa';

Blockchain.POLYGON
```
as the docs say

PR #474, opened 2024-04-25
AA: Add verify message helper

MatiNavCrossmint:
## Description

Add verify message helper

## Test plan

- Sign message with AA && verify same message => true
- Sign Message with AA && verify diff message => false

## Implementation
code

### DIFF
--- a/packages/client/wallets/aa/src/utils/blockchain.ts
+++ b/packages/client/wallets/aa/src/utils/blockchain.ts
@@ -1,7 +1,18 @@
+import { getBundlerRPC } from "@/blockchain";
 import type { Deferrable } from "@ethersproject/properties";
 import { type TransactionRequest } from "@ethersproject/providers";
 import { BigNumber, BigNumberish } from "ethers";
 import { BytesLike, hexlify } from "ethers/lib/utils";
+import { createPublicClient, http } from "viem";
+
+import { EVMBlockchainIncludingTestnet } from "@crossmint/common-sdk-base";
+
+export type VerifyMessageParams = {
+    address: `0x${string}`;
+    message: string;
+    signature: `0x${string}` | Uint8Array;
+    chain: EVMBlockchainIncludingTestnet;
+};
 
 /**
  * We need to support odd value and data fields in a tx, as that data comes from WC
@@ -50,4 +61,16 @@ export async function convertData(
     const resolvedData = await Promise.resolve(data);
     if (resolvedData === undefined) return undefined;
     return hexlify(resolvedData) as `0x${string}`;
+}
+
+export async function verifyMessage({ address, message, signature, chain }: VerifyMessageParams) {
+    const publicClient = createPublicClient({
+        transport: http(getBundlerRPC(chain)),
+    });
+
+    return publicClient.verifyMessage({
+        address,
+        message,
+        signature,
+    });
 }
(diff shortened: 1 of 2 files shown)


packages/client/wallets/aa/src/utils/blockchain.ts
```diff
@@ -51,3 +62,15 @@ export async function convertData(
     if (resolvedData === undefined) return undefined;
     return hexlify(resolvedData) as `0x${string}`;
 }
+
+export async function verifyMessage({ address, message, signature, chain }: VerifyMessageParams) {
+    const publicClient = createPublicClient({
+        transport: http(getBundlerRPC(chain)),
```
afeight, line 68: Does chain matter? I think signatures are the same on all chain, so if we can confirm, I'd like to hard-code this to one EVM chain (maybe Ethereum) so that the client doesn't have to think about the variable

packages/client/wallets/aa/src/utils/blockchain.ts
```diff
@@ -1,7 +1,18 @@
...
 import { BytesLike, hexlify } from "ethers/lib/utils";
+import { createPublicClient, http } from "viem";
+
+import { EVMBlockchainIncludingTestnet } from "@crossmint/common-sdk-base";
+
+export type VerifyMessageParams = {
+    address: `0x${string}`;
+    message: string;
+    signature: `0x${string}` | Uint8Array;
+    chain: EVMBlockchainIncludingTestnet;
```
afeight, line 14: if we do need to keep chain, this should be `Blockchain` – we've made an alias because for users of the SDK, `EVMBlockchainIncludingTestnet` is a rather unsightly type name haha

MatiNavCrossmint: I agree, although the issue is that we need to use a type there, and the options are:
- `EVMBlockchainIncludingTestnet`
- or create a new type ( I'd hate that 😅)

`Blockchain` or `EVMBlockchain` won't work because those don't include testnet chains.

Also, the user will see it differently:
 Screenshot 2024-04-26 at 11 38 51 AM

If `EVMBlockchainIncludingTestnet` sounds bad then I vote to not export `VerifyMessageParams`. I just exported it to give to the developer a quick type if needed. 

LMK

afeight: ah, somehow this got reverted but we need it to match the API we outlined in EDD
https://github.com/Crossmint/crossmint-sdk/pull/436/files

afeight: I added this because in our EDD, we specifically call this type "Blockchain"
https://github.com/Crossmint/crossmint-sdk/pull/436/files
so we can import it from within the SDK itself rather than common-sdk, but we may need to move this somewhere else to make it more accessible

MatiNavCrossmint: Even when used as in the example, from the implementation project it's seen as the original name type

code
Screenshot 2024-04-26 at 1 24 56 PM

afeight: hmmm ok I wonder if we need to do this in a different way then. For now we can merge this

afeight (approved):

packages/client/wallets/aa/src/utils/blockchain.ts
MatiNavCrossmint: Sure thing, If we find a way I'll change it

PR #479, opened 2024-04-29
AA: Add Zkyoto support and deprecate Zkatana

MatiNavCrossmint:
## Description

AA: Add Zkyoto support and deprecate Zkatana

## Test plan

Created AA Wallet and minted through it

### DIFF
--- a/packages/common/base/src/blockchain/types/evm.ts
+++ b/packages/common/base/src/blockchain/types/evm.ts
@@ -10,6 +10,7 @@ export const EVMBlockchain = {
     ZORA: "zora",
     ARBITRUMNOVA: "arbitrumnova",
     ZKATANA: "zkatana",
+    ZKYOTO: "zkyoto",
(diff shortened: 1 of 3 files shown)


packages/common/base/src/blockchain/types/evm.ts
```diff
@@ -10,6 +10,7 @@ export const EVMBlockchain = {
     ZORA: "zora",
     ARBITRUMNOVA: "arbitrumnova",
     ZKATANA: "zkatana",
+    ZKYOTO: "zkyoto",
```
maxisch, line 13: we should remove ZKATANA to be consistent with the other changes

MatiNavCrossmint: I will leave it to handle it's removal in another PR, thanks for the catch!!

afeight (approved):

PR #480, opened 2024-05-02
SCW-536: Accept userId as entry point versioning

maxisch:
## Description

Accept email or userId when querying for entry point versioning

## Test plan

Manual testing

### DIFF
--- a/packages/client/wallets/aa/src/api/CrossmintWalletService.ts
+++ b/packages/client/wallets/aa/src/api/CrossmintWalletService.ts
@@ -1,6 +1,6 @@
 import { GenerateSignatureDataInput, StoreAbstractWalletInput, UserIdentifier } from "@/types";
 
-import { EVMBlockchainIncludingTestnet } from "@crossmint/common-sdk-base";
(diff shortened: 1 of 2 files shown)


packages/client/wallets/aa/src/api/CrossmintWalletService.ts
```diff
@@ -26,9 +26,16 @@ export class CrossmintWalletService extends BaseCrossmintService {
...
+        userIdentifier: UserIdentifierParams,
+        chain: EVMBlockchainIncludingTestnet
+    ) {
+        const identifier = userIdentifier.email
+            ? `email=${encodeURIComponent(userIdentifier.email)}`
+            : `userId=${userIdentifier.userId}`;
+
         return this.fetchCrossmintAPI(
-            `v1-alpha1/wallets/entry-point-version?email=${encodeURIComponent(email)}&chain=${chain}`,
+            `v1-alpha1/wallets/entry-point-version?${identifier}&chain=${chain}`,
```
afeight, line 38: oh now I see–using query params in GET makes the JSON harder. If possible that would make the code easier in crossbit-main, but if it's too messy then it makes sense to keep as-is

afeight (approved):

PR #481, opened 2024-05-02
AA: Add Zkyoto support 

MatiNavCrossmint:
## Description

AA: Add Zkyoto support

Also, improved types and removed/explained the "as ..." used through the EVMAAWallet

## Test plan

Created AA Wallet locally and minted through it.

### DIFF
--- a/packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
+++ b/packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
@@ -257,19 +257,9 @@ export class EVMAAWallet<B extends EVMBlockchainIncludingTestnet = EVMBlockchain
     getSigner<Type extends SignerType>(type: Type): SignerMap[Type] {
         switch (type) {
             case "viem": {
-                //zerodev returns a type where "account" could be undefined, and "walletClient" requires it to be defined
-                // and this hack only changes that type. So, the change is:
-                // KernelSmartAccount<EntryPoint, HttpTransport, TChain> | undefined => KernelSmartAccount<EntryPoint, HttpTransport, TChain>
-                const kernelClientHackType = this.kernelClient as KernelAccountClient<
-                    EntryPoint,
-                    HttpTransport,
-                    TChain,
-                    KernelSmartAccount<EntryPoint, HttpTransport, TChain>
-                >;
-
                 return {
                     publicClient: this.publicClient,
-                    walletClient: kernelClientHackType,
+                    walletClient: this.kernelClient,
                 };
             }
             default:
--- a/packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
+++ b/packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
@@ -51,7 +51,14 @@ export class EVMAAWallet<B extends EVMBlockchainIncludingTestnet = EVMBlockchain
     // account represents the one got from eth_requestAccounts.
     // we need to expose it because createKernelAccountClient returns that it can be undefined, and some methods require it to be specifically included because of that.
     public account: KernelSmartAccount<EntryPoint, HttpTransport, TChain>;
-    private kernelClient: ReturnType<typeof createKernelAccountClient>;
+    private kernelClient: ReturnType<
+        typeof createKernelAccountClient<
+            EntryPoint,
+            HttpTransport,
+            TChain,
+            KernelSmartAccount<EntryPoint, HttpTransport, TChain>
+        >
+    >;
     private entryPoint: EntryPoint;
     chain: B;
 
@@ -114,10 +121,6 @@ export class EVMAAWallet<B extends EVMBlockchainIncludingTestnet = EVMBlockchain
 
     async signMessage(message: string | Uint8Array) {
         try {
-            if (this.kernelClient.account == null) {
-                throw new Error("Kernel client needs to be initialized before signing message.");
-            }
-
             let messageAsString: string;
             if (message instanceof Uint8Array) {
                 const decoder = new TextDecoder();
@@ -128,7 +131,6 @@ export class EVMAAWallet<B extends EVMBlockchainIncludingTestnet = EVMBlockchain
 
             return await this.kernelClient.signMessage({
                 message: messageAsString,
-                account: this.kernelClient.account,
             });
         } catch (error) {
             logError("[SIGN_MESSAGE] - ERROR", {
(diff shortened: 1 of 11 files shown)


packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -330,11 +375,11 @@ export class EVMAAWallet<B extends EVMBlockchainIncludingTestnet = EVMBlockchain
                 throw new Error("New version info not found");
             }
 
-            const enableSig = await this.kernelClient.account.kernelPluginManager.getPluginEnableSignature(
-                this.kernelClient.account.address
+            const enableSig = await this.kernelClient.account?.kernelPluginManager.getPluginEnableSignature(
+                this.kernelClient.account?.address
```
afeight, line 379: above on line 359 we have the `== null` check, we should be able to remove these `?` now, right?
```suggestion
            const enableSig = await this.kernelClient.account.kernelPluginManager.getPluginEnableSignature(
                this.kernelClient.account.address
```

we also use `this.kernelClient.account` a bunch of places in this function, we may want to do
```
const { account } = this.kernelClient
if (account == null) {
  throw Error

File: packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts [hunk 2]
```diff
@@ -235,9 +256,19 @@ export class EVMAAWallet<B extends EVMBlockchainIncludingTestnet = EVMBlockchain
...
             case "viem": {
+                //zerodev returns a type where "account" could be undefined, and "walletClient" requires it to be defined
+                // and this hack only changes that type. So, the change is:
+                // KernelSmartAccount<EntryPoint, HttpTransport, TChain> | undefined => KernelSmartAccount<EntryPoint, HttpTransport, TChain>
+                const kernelClientHackType = this.kernelClient as KernelAccountClient<
+                    EntryPoint,
+                    HttpTransport,
+                    TChain,
+                    KernelSmartAccount<EntryPoint, HttpTransport, TChain>
+                >;
```
afeight, line 267: maybe we should do some validation here then before changing it to a more "dangerous" type
What do you think about
```suggestion
                if (this.kernelClient.account == null) {
                    throw new Error(...)
                }
                const kernelClientHackType = this.kernelClient as KernelAccountClient<
                    EntryPoint,
                    HttpTransport,
                    TChain,
                    KernelSmartAccount<EntryPoint, HttpTransport, TChain>
                >;
```

MatiNavCrossmint:
@afeight the `kernelClient` type was improved and account is shown as defined now. So all comments related to it doesn't apply anymore

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -37,17 +38,26 @@ import {
...
     private ecdsaValidator: KernelValidator<entryPoint, "ECDSAValidator">;
-    private account: KernelSmartAccount<EntryPoint, HttpTransport, TChain>;
-    private kernelClient: KernelAccountClient<
-        entryPoint,
-        HttpTransport,
-        TChain,
-        KernelSmartAccount<entryPoint, HttpTransport, TChain>
+    // account represents the one got from eth_requestAccounts.
+    // we need to expose it because createKernelAccountClient returns that it can be undefined, and some methods require it to be specifically included because of that.
+    public account: KernelSmartAccount<EntryPoint, HttpTransport, TChain>;
```
afeight, line 53: this part I don't understand. I think we should keep this private if there's any way to do so as this is the kind of internal we want to hide from the developer

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -348,4 +376,35 @@ export class EVMAAWallet<B extends EVMBlockchainIncludingTestnet = EVMBlockchain
...
+                );
+            }
+            return {
+                // It's on zerodev doc that we need to ignore ts errros on maxFeePerGas and maxPriorityFeePerGas
+                // https://docs.zerodev.app/sdk/faqs/use-with-gelato#transaction-configuration
+                maxFeePerGas: "0x0" as any,
+                maxPriorityFeePerGas: "0x0" as any,
+            };
+        }
+    }
```
afeight, line 409: this file is already quite long with 400 lines–do these rely on some internal aspect or can we put these into a separate helper file?
`hasEIP1559Support(chain: Blockchain)`

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -139,6 +156,12 @@ export class EVMAAWallet<B extends EVMBlockchainIncludingTestnet = EVMBlockchain
         }
     }
 
+    //TODO @matias: review this method.
+    // First, I would like to use TransactionRequest from viem instead of ethers.
+    // Second, we need to check if chain supports eip-1559 ro not:
+    // - If it does, we need to send maxFeePerGas and maxPriorityFeePerGas
+    // - If it doesn't, we need to send gasPrice
+    // And with the use of viem TransactionRequest, we can specify the TransactionRequest type (eip1559 or legacy) and be more accurate
```
afeight, line 164: hm yeah, also I'm not sure why this accepts a Deferrable–it makes our life more difficult

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
afeight: ah ok, sorry for missing some context 
I see that this is for [this change](https://github.com/Crossmint/crossmint-sdk/pull/481/files#diff-74abb96fec665aa4e65410adc9d1d80a34750de5bd7e3a01222827ee21c0217fR20)

To make sure I understand, what's happening is that:
With a non-eip1559-supported chain (astar), `signTypedData` doesn't work unless you use different parameters (include `account`)
If that's the case, would we be able to leave `account` private by doing this?
```
async signTypedData(params: TypedDataDefinition) {
    try {
        return await this.kernelClient.signTypedData({ ...params, account: this.account });
    } catch (error) {
        logError("[SIGN_TYPED_DATA] - ERROR", {
            service: SCW_SERVICE,
            error: errorToJSON(error),
            signer: this.kernelClient.account,
        });
        throw new Error(`Error signing typed data. If this error persists, please contact support.`);
    }
}

afeight: and then `reservoirAdapter` doesn't have to change?

afeight (approved):
approve for now but would like to keep it as `private account` before merge if the outlined route would allow it
https://github.com/Crossmint/crossmint-sdk/pull/481/files#r1589770592

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
MatiNavCrossmint: yes, that's something that takes my attention too

PR #482, opened 2024-05-03
SCW-538: Transfer function proposal

maxisch:
## Description

Leverage typescript for a nice transfer API

## Test plan

Manual

### DIFF
--- a/packages/client/wallets/aa/src/blockchain/token/Tokens.ts
+++ b/packages/client/wallets/aa/src/blockchain/token/Tokens.ts
@@ -1,9 +1,9 @@
 import { BigNumber } from "ethers";
 
-import { EVMBlockchain, isEVMBlockchain } from "@crossmint/common-sdk-base";
+import { EVMBlockchainIncludingTestnet, isEVMBlockchain } from "@crossmint/common-sdk-base";
 
 export interface EVMToken {
-    chain: EVMBlockchain;
+    chain: EVMBlockchainIncludingTestnet;
     contractAddress: string;
 }
 
--- a/packages/client/wallets/aa/src/blockchain/token/Tokens.ts
+++ b/packages/client/wallets/aa/src/blockchain/token/Tokens.ts
@@ -1,18 +1,81 @@
-import { EVMBlockchain } from "@crossmint/common-sdk-base";
+import { BigNumber } from "ethers";
+
+import { EVMBlockchain, isEVMBlockchain } from "@crossmint/common-sdk-base";
 
 export interface EVMToken {
     chain: EVMBlockchain;
     contractAddress: string;
-    tokenId?: string;
 }
+
+export interface NFTEVMToken extends EVMToken {
+    tokenId: string;
+    type: NFT;
+}
+
+export interface SFTEVMToken extends EVMToken {
+    tokenId: string;
+    type: SFT;
+}
+
+export interface ERC2OEVMToken extends EVMToken {
+    type: FT;
+}
+
 export interface SolanaToken {
     mintHash: string;
     chain: "solana";
+    type: NFT;
 }
 export interface CardanoToken {
     chain: "cardano";
     assetId: string;
+    type: NFT;
+}
+
+export function isEVMToken(value: unknown): value is EVMToken {
+    if (typeof value !== "object" || value === null) {
+        return false;
+    }
+    const possibleEVMToken = value as Partial<EVMToken>;
+    return (
+        typeof possibleEVMToken.chain === "string" &&
+        typeof possibleEVMToken.contractAddress === "string" &&
+        isEVMBlockchain(possibleEVMToken.chain)
+    );
+}
+
+export function isNFTEVMToken(value: unknown): value is NFTEVMToken {
+    if (!isEVMToken(value)) {
+        return false;
+    }
+    const possibleNFTEVMToken = value as NFTEVMToken;
+    return possibleNFTEVMToken.type === "NFT";
+}
+
+export function isSFTEVMToken(value: unknown): value is SFTEVMToken {
+    if (!isEVMToken(value)) {
+        return false;
+    }
+    const possibleSFTEVMToken = value as SFTEVMToken;
+    return possibleSFTEVMToken.type === "SFT";
+}
+
+export function isERC20EVMToken(value: unknown): value is ERC2OEVMToken {
+    if (!isEVMToken(value)) {
+        return false;
+    }
+    const possibleERC20EVMToken = value as ERC2OEVMToken;
+    return possibleERC20EVMToken.type === "FT";
 }
 
 export type Token = EVMToken | SolanaToken | CardanoToken;
-export type TokenType = "NFT" | "SFT" | "FT";
+
+export type NFT = "NFT";
+export type SFT = "SFT";
+export type FT = "FT";
+export type TokenType = NFT | SFT | FT;
+
+export type ERC20TransferType = { token: ERC2OEVMToken; amount: BigNumber };
+export type SFTTransferType = { token: SFTEVMToken; quantity: number };
+export type NFTTransferType = { token: NFTEVMToken | SolanaToken | CardanoToken };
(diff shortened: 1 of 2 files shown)


MatiNavCrossmint (approved):
This looks good, thanks for handling this!

BTW, should we make that the demo consumes this transfer function? If you tested it on there, maybe you can push a new PR with the changes!

packages/client/wallets/aa/src/blockchain/token/Tokens.ts
```diff
@@ -1,18 +1,81 @@
...
+    type: NFT;
+}
+
+export interface SFTEVMToken extends EVMToken {
+    tokenId: string;
+    type: SFT;
+}
+
+export interface ERC2OEVMToken extends EVMToken {
+    type: FT;
```
afeight, line 21: I think we're adding a layer of indirection we don't need– if these are strings, they're treated as consts / their own type in TS
```suggestion
    type: "nft";
}

export interface SFTEVMToken extends EVMToken {
    tokenId: string;
    type: "sft";
}

export interface ERC2OEVMToken extends EVMToken {
    type: "ft";
```

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -192,65 +206,75 @@ export class EVMAAWallet<B extends EVMBlockchainIncludingTestnet = EVMBlockchain
...
+        }
+
         const contractAddress = evmToken.contractAddress as `0x${string}`;
         const publicClient = this.kernelClient.extend(publicActions);
         let transaction;
+        let tokenId;
         try {
-            if (amount !== undefined) {
-                // Transfer ERC20
+            if (isERC20EVMToken(evmToken)) {
```
afeight, line 220: and when we then do `if (evmToken.type === "ft") {`...

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -192,65 +206,75 @@ export class EVMAAWallet<B extends EVMBlockchainIncludingTestnet = EVMBlockchain
...
                 const { request } = await publicClient.simulateContract({
                     account: this.account,
                     address: contractAddress,
                     abi: erc20,
                     functionName: "transfer",
-                    args: [toAddress, amount],
+                    args: [toAddress, (config as ERC20TransferType).amount],
```
afeight, line 226: ...then typescript should be able to infer that `config` is `ERC20TransferType` and allow this to be 
```suggestion
                    args: [toAddress, config.amount],
```
(I think)

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -192,65 +206,75 @@ export class EVMAAWallet<B extends EVMBlockchainIncludingTestnet = EVMBlockchain
...
                     abi: erc1155,
                     functionName: "safeTransferFrom",
-                    args: [this.getAddress(), toAddress, evmToken.tokenId, quantity, "0x00"],
+                    args: [this.getAddress(), toAddress, tokenId, (config as SFTTransferType).quantity, "0x00"],
+                    ...this.getLegacyTransactionFeesParamsIfApply(),
                 });
                 transaction = await publicClient.writeContract(request);
-            } else {
-                // Transfer ERC721
+            } else if (isNFTEVMToken(evmToken)) {
```
afeight, line 241: similar for these two

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -192,65 +206,75 @@ export class EVMAAWallet<B extends EVMBlockchainIncludingTestnet = EVMBlockchain
...
                     address: contractAddress,
                     abi: erc721,
                     functionName: "safeTransferFrom",
-                    args: [this.getAddress(), toAddress, evmToken.tokenId],
+                    args: [this.getAddress(), toAddress, tokenId],
+                    ...this.getLegacyTransactionFeesParamsIfApply(),
                 });
                 transaction = await publicClient.writeContract(request);
+            } else {
+                throw new WalletSdkError(`Token not supported`);
```
afeight, line 253: could even do a switch case and have the compiler assert this is not reachable

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -380,8 +404,8 @@ export class EVMAAWallet<B extends EVMBlockchainIncludingTestnet = EVMBlockchain
 
         if (hasEIP1559Support(this.chain)) {
             return {
-                maxFeePerGas: maxFeePerGas ? BigInt(maxFeePerGas.toString()) : undefined,
-                maxPriorityFeePerGas: maxPriorityFeePerGas ? BigInt(maxPriorityFeePerGas.toString()) : undefined,
+                ...(maxFeePerGas && { maxFeePerGas: BigInt(maxFeePerGas.toString()) }),
+                ...(maxPriorityFeePerGas && { maxPriorityFeePerGas: BigInt(maxPriorityFeePerGas.toString()) }),
```
afeight, line 408: this caught me off guard because I'm thinking we should do `maxFeePerGas != null && ...` but that's not true, we do want to ignore if it's 0 also
so maybe drop a comment
```suggestion
                // only include if non-null and non-zero
                ...(maxFeePerGas && { maxFeePerGas: BigInt(maxFeePerGas.toString()) }),
                ...(maxPriorityFeePerGas && { maxPriorityFeePerGas: BigInt(maxPriorityFeePerGas.toString()) }),
```

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -192,65 +206,75 @@ export class EVMAAWallet<B extends EVMBlockchainIncludingTestnet = EVMBlockchain
...
                 service: SCW_SERVICE,
                 error: errorToJSON(error),
-                tokenId: evmToken.tokenId,
+                tokenId: tokenId,
                 contractAddress: evmToken.contractAddress,
                 chain: evmToken.chain,
             });
-            throw new TransferError(`Error transferring token ${evmToken.tokenId}`);
+            throw new TransferError(
+                `Error transferring token ${evmToken.contractAddress}${!tokenId ? "" : ` tokenId=${tokenId}}`}`
```
afeight, line 274: standard practice is "contractAddress:tokenId` like `0x123:55`
```suggestion
                `Error transferring token ${evmToken.contractAddress}${tokenId == null ? "" : `:${tokenId}}`}`
```

packages/client/wallets/aa/src/blockchain/token/Tokens.ts
```diff
@@ -1,18 +1,81 @@
...
+    const possibleSFTEVMToken = value as SFTEVMToken;
+    return possibleSFTEVMToken.type === "SFT";
+}
+
+export function isERC20EVMToken(value: unknown): value is ERC2OEVMToken {
+    if (!isEVMToken(value)) {
+        return false;
+    }
+    const possibleERC20EVMToken = value as ERC2OEVMToken;
+    return possibleERC20EVMToken.type === "FT";
```
afeight, line 68: I'm hoping that using `if (config.token.type === "nft")` in EVMAAWallet.transfer means we can remove the need for these

afeight (approved):
sweet, this looks great. Much better to have distinct types to inform which case rather than multiple optional params
Have some ideas that may help remove some code, but not 100% sure it'll work so if it doesn't then feel free to ignore

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
maxisch: Sadly no, I tried that and checked again and typescript is not smart enough

PR #484, opened 2024-05-17
AA: Improve logging

MatiNavCrossmint:
## Description

Previous to this PR, we had some logging on AA but wasn't enough. We need to have a more detailed one where we can see and reproduce why a behavior happened with just seeing the logs. This PR makes that possible.

Screenshot 2024-05-17 at 10 19 19 AM

## Test plan

Tested locally and through using DD


afeight (approved):
would be great if we could get a unit test around `LoggerWrapper` because it's quite likely we accidentally make a change that breaks many functions
overall looks great ✅

PR #487, opened 2024-05-22
payment-errors: new handled error payments:recipient.missing

pgarciaegido:
## Description

Read: https://linear.app/crossmint/issue/PAY-2972/better-error-for-emailinputoptions

## Test plan

Will test in crossbit main

### DIFF
--- a/packages/client/base/src/types/errors.ts
+++ b/packages/client/base/src/types/errors.ts
@@ -3,7 +3,7 @@ export const PaymentErrors = {
     PAYMENT_METHOD_INVALID: "payments:payment-method.invalid",
     EMAIL_INVALID: "payments:email.invalid",
     CLIENT_ID_INVALID: "payments:client-id.invalid",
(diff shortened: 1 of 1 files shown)


packages/client/base/src/types/errors.ts
```diff
@@ -3,6 +3,7 @@ export const PaymentErrors = {
     PAYMENT_METHOD_INVALID: "payments:payment-method.invalid",
     EMAIL_INVALID: "payments:email.invalid",
     CLIENT_ID_INVALID: "payments:client-id.invalid",
+    MISSING_RECIPIENT: "payments:recipient.missing",
```
mPaella, line 6: should use required, not missing, to match headless

guillermodoghel: +1

guillermodoghel (approved):
lgtm!

PR #490, opened 2024-05-28
Simplification: Remove unused FB & Lit Pieces

devlyn37:
## Description

No longer using these, so dropping from SDK to reduce noise

## Test plan

TS Compiling and existing tests


packages/client/wallets/aa/src/utils/error.test.ts
devlyn37, line 1: These tests might as well say. 

```TypeScript
export const x = 5;

describe("Testing x value", () => {
    expect(x).toBe(5);
}) 
```

maxisch (approved):
much needed!

PR #491, opened 2024-05-28
Simplification: Drop Chain Generics

devlyn37:
## Description

Removing the generics here for a few reasons
- They don't seem necessary anymore
- It creates a mess when using our SDK pictured below

Screenshot 2024-05-28 at 12 20 47 PM

## Test plan

TS Compiling & Existing Tests


packages/client/wallets/aa/src/CrossmintAASDK.ts
```diff
@@ -1,21 +1,14 @@
 import { CrossmintWalletService } from "@/api";
 import { EVMAAWallet, TChain, entryPoint, getBundlerRPC } from "@/blockchain";
-import type { BackwardsCompatibleChains, CrossmintAASDKInitParams, WalletConfig } from "@/types";
-import {
-    CURRENT_VERSION,
-    WalletSdkError,
-    ZERO_DEV_TYPE,
```
maxisch, line 7: we will need `ZERO_DEV_TYPE ` and `CURRENT_VERSION` for `storeAbstractWallet`, but we'll add them later

maxisch (approved):

PR #492, opened 2024-05-28
Simplification: Drop Unfinished Upgrade Method

devlyn37:
## Description

The `upgradeVersion` method in it's current form:
- Creates a signature to enable a new plugin
- Saves that to CM's service

There's no reason to expose an unfinished SDK method!

## Test plan

TS compiling & existing tests


maxisch (approved):
This makes sense, we can add it whenever we finish this functionality

PR #488, opened 2024-05-23
WIP: Passkeys & Minimalism Exercise

devlyn37:
## Description

This next SDK update that includes passkey wallets is a good opportunity simplify and reduce our API surface!

This PR Does Two Things:

### An Exercise in Minimalism

AA has a lot of moving pieces, unless we're ruthless about simplifying & removing things as we go, we'll drown in unnecessary complexity! This PR drops a ton from the SDK, probably too much. The idea here is that the team will review, we'll figure out what's really needed, and drop the rest. I've annotated different pieces in the diff.

Big Questions:
- Does the "session key" setup we have actually work? What's the intention behind it?
- Why are we rolling out own signer methods (`sendTransaction`, `signMessage`, etc) in `EVMAAWallet`? How about clients grab the signer and use that instead?
- Most of our E2E testing here seems to be related to `ethers` or `fireblocks`, both things we don't plan on supporting atm, are they needed anymore?

### Add Passkey Wallet M1 

This PR adds a draft of the API we'll expose for creating passkey wallets. I'll annotate this within the PR!

Things that are totally WIP and not ready for review yet:
- Passkey edge cases & errors we return from the SDK
- The implementation of the SDK
    - I still have this branch using ZD's passkey server
    - We still need to store passkey wallets within the CM backend.

## Test plan

Not sure yet, there will be one though.

PR #499, opened 2024-05-29
Simplification: Drop Custodian Session Key Pt. 2

devlyn37:
## Description

This PR:
- Stops session keys from being created with new AA wallets in conjunction with  https://github.com/Paella-Labs/crossbit-main/pull/13254.
    - For these session keys have any effect, the user must sign.
    - We've removed the method to do that from the SDK, since it's not ready yet.
    - When/if we re-enable, we'll include custodial session key (A FB vault on CM Backend) creation within the method, not upon wallet creation.
- In anticipation of passkeys, organizes EOA wallet creation into a `EOAWalletService` class.

## Test plan

TS Compiling + Existing Tests

Will test E2E before merging once the `crossbit-main` PR lands

### DIFF
--- a/packages/client/wallets/aa/src/CrossmintAASDK.ts
+++ b/packages/client/wallets/aa/src/CrossmintAASDK.ts
@@ -1,26 +1,20 @@
 import { CrossmintWalletService } from "@/api";
-import { EVMAAWallet, getBundlerRPC } from "@/blockchain";
+import { getBundlerRPC } from "@/blockchain";
 import type { CrossmintAASDKInitParams, WalletConfig } from "@/types";
-import { CURRENT_VERSION, WalletSdkError, ZERO_DEV_TYPE, createOwnerSigner } from "@/utils";
-import { signerToEcdsaValidator } from "@zerodev/ecdsa-validator";
-import { createKernelAccount } from "@zerodev/sdk";
+import { WalletSdkError } from "@/utils";
 import { ENTRYPOINT_ADDRESS_V06, ENTRYPOINT_ADDRESS_V07 } from "permissionless";
 import { EntryPointVersion } from "permissionless/types/entrypoint";
 import { createPublicClient, http } from "viem";
 
-import {
-    EVMBlockchainIncludingTestnet,
-    UserIdentifierParams,
-    blockchainToChainId,
-    isEVMBlockchain,
-    validateAPIKey,
-} from "@crossmint/common-sdk-base";
+import { EVMBlockchainIncludingTestnet, UserIdentifierParams, validateAPIKey } from "@crossmint/common-sdk-base";
 
+import EOAWalletService from "./blockchain/wallets/eoa";
 import { LoggerWrapper, logPerformance } from "./utils/log";
 import { parseUserIdentifier } from "./utils/user";
 
 export class CrossmintAASDK extends LoggerWrapper {
-    crossmintService: CrossmintWalletService;
+    private readonly crossmintService: CrossmintWalletService;
+    private readonly eaoWalletService: EOAWalletService;
 
     private constructor(config: CrossmintAASDKInitParams) {
         super("CrossmintAASDK");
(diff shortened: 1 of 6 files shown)


packages/client/wallets/aa/src/CrossmintAASDK.ts
```diff
@@ -45,62 +40,18 @@ export class CrossmintAASDK extends LoggerWrapper {
             "GET_OR_CREATE_WALLET",
             async () => {
                 try {
-                    if (!isEVMBlockchain(chain)) {
```
devlyn37, line 48: Should have dropped this when I removed the chain generics [here](https://github.com/Crossmint/crossmint-sdk/pull/491), but this is no longer needed since TS enforces it.

packages/client/wallets/aa/src/blockchain/wallets/eoa.ts
```diff
@@ -0,0 +1,60 @@
...
+            index: BigInt(0),
+            entryPoint,
+        });
+
+        const wallet = new EVMAAWallet(account, this.crossmintService, publicClient, entryPoint, chain);
+        await this.crossmintService.storeAbstractWallet({
+            userIdentifier,
+            type: ZERO_DEV_TYPE,
+            smartContractWalletAddress: account.address,
+            eoaAddress: eoa.address,
```
maxisch, line 51: This is outdated, we should send something like this: https://github.com/Crossmint/crossmint-sdk/pull/488/files#diff-54c932ab0149942d2ac2dcc04b35da86081332beee12815702d8c02bdeba601fR54

With this new format: https://github.com/Crossmint/crossmint-sdk/pull/488/files#diff-c0a3c29265bf1b60a9a23113ccf53283a5864674ea6bc39e03a40d02515b575aR11

```
await this.crossmintService.storeAbstractWallet({
            userIdentifier,
            type: ZERO_DEV_TYPE,
            smartContractWalletAddress: account.address,
            signerData: {
                eoaAddress: eoa.address,
                type: "eoa",
            },
            version: CURRENT_VERSION,
            baseLayer: "evm",
            chainId: blockchainToChainId(chain),
            entryPointVersion,
        });
```
We can merge this if not and I'll submit a PR with those changes to main. Was pushing it in the other branch

maxisch: I mean, it will technically work as the endpoint is backwards compatible in crossbit. But let's adopt the new format here while we are at it

devlyn37: ofc, forgot about that, will update

packages/client/wallets/aa/src/utils/error.ts
```diff
@@ -11,31 +11,6 @@ export class NotAuthorizedError extends Error {
     }
 }
 
-export class RateLimitError extends Error {
```
devlyn37, line 14: Dropping since they're not used atm

packages/client/wallets/aa/src/utils/error.ts
devlyn37, line 1: dropping some unused errors

packages/client/wallets/aa/src/types/API.ts
```diff
@@ -6,14 +6,19 @@ export type StoreAbstractWalletInput = {
     userIdentifier: UserIdentifier;
     type: string;
     smartContractWalletAddress: string;
-    eoaAddress: string;
-    sessionKeySignerAddress: string;
+    signerData: EOASignerData;
```
maxisch, line 9: Are we adding the `PasskeysSignerData` in another PR?

maxisch (approved):
🚀

packages/client/wallets/aa/src/types/API.ts
devlyn37: Yeah coming soon

PR #504, opened 2024-05-30
Simplification: Drop `SignTypedMessage`

devlyn37:
## Description

This PR drops `SignTypedMessage`. Rational being:
- Our other high level methods add value in some way, either by having a clearer interface or converting types, etc. `EVMAAWallet.SignTypedMessage` does not, it's just a wrapper over `kernelClient`.
- `SignTypedMessage` should fall under advanced usage that an be done via `getSigner`

## Test plan

TS Compiling & Existing Tests

### DIFF
--- a/packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
+++ b/packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
@@ -23,7 +23,7 @@ import {
 } from "@zerodev/sdk";
 import { BigNumberish } from "ethers";
 import { EntryPoint } from "permissionless/types/entrypoint";
-import type { Hash, HttpTransport, PublicClient, TypedDataDefinition } from "viem";
+import type { Hash, HttpTransport, PublicClient } from "viem";
 import { Chain, http, publicActions } from "viem";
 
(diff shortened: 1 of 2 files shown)


packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
devlyn37, line 1: I'm considering dropping `signMessage` to since we don't add much value over `kernelClient.signMessage`, TAL at the code:

```TypeScript
    public async signMessage(message: string | Uint8Array) {
        return this.logPerformance("SIGN_MESSAGE", async () => {
            try {
                let messageAsString: string;
                if (message instanceof Uint8Array) {
                    const decoder = new TextDecoder();
                    messageAsString = decoder.decode(message);
                } else {
                    messageAsString = message;
                }

                return await this.kernelClient.signMessage({
                    message: messageAsString,
                });
            } catch (error) {
                throw new Error(`Error signing message. If this error persists, please contact support.`);
            }
        });
    }
```

alfonso-paella:
Is this a breaking change? If so, what's the plan for a responsible deprecation?

maxisch (approved):

devlyn37:
> Is this a breaking change? If so, what's the plan for a responsible deprecation?

I'll message Racino!

PR #505, opened 2024-06-02
Simplification: Signer Type

devlyn37:
## Description

To simplify `AAEVMWallet`, given that we only support one signer type atm, this PR:
- Adds a default param, so that SDK consumers can call `getSigner()` and get the `viem` signer by default.
- Removes some types that are overkill for now, but can be added back later w/o breaking the API.

## Test plan

Passkeys QA
TS Compiling & Existing Tests

### DIFF
--- a/packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
+++ b/packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
@@ -1,5 +1,4 @@
 import { logError, logInfo } from "@/services/logging";
-import { SignerMap, SignerType } from "@/types";
 import {
     SCW_SERVICE,
     TransactionError,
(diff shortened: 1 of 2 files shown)


packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -213,7 +212,7 @@ export class EVMAAWallet extends LoggerWrapper {
         });
     }
 
-    public getSigner<Type extends SignerType>(type: Type): SignerMap[Type] {
```
devlyn37, line 216: no point in having this elaborate type setup w/ one signer type. Simplifying where we can here

maxisch (approved):
makes sense!

PR #510, opened 2024-06-08
Simplification: Smart Account Client Handling & Types

devlyn37:
## Description

Context here is that we're in the process of minimizing our API surface, including removing the redundant methods:
- `EVMAAWallet.sendTransaction`
- `EVMAAWallet.signMessage`
- `EVMAAWallet.signTypedData`
In favor of letting developers using a `walletClient` we expose directly instead.

In order to completely replace these methods, we need to add some error handling & logging to the `walletClient` we expose. This PR:
- Creates a minimal wrapper over `SmartAccountClient` to show how we'll do this
- Updates `getSigner` to return a `SmartAccountClient` rather than a `KernelAccountClient` since our API should aim not to expose provider specific types.

Within the PRs that remove the methods I mention above, we'll replace any functionality within this wrapper.

## Test plan

Local Testing on SCW Demo on branch `main-w3a`:
- Confirmed that the pieces added in `toCrossmintSmartAccountClient` used within `EVMAAWallet` are working as expected by checking additional logs.
- Confirmed that minting still works.

Passkey Release QA.
TS Compiling & Existing Tests.

### DIFF
--- a/packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
+++ b/packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
@@ -7,15 +7,15 @@ import {
     convertData,
     decorateSendTransactionData,
     errorToJSON,
+    gelatoBundlerProperties,
     getNonce,
-    hasEIP1559Support,
+    usesGelatoBundler,
 } from "@/utils";
 import { resolveDeferrable } from "@/utils/deferrable";
 import { LoggerWrapper } from "@/utils/log";
 import type { Deferrable } from "@ethersproject/properties";
 import { type TransactionRequest } from "@ethersproject/providers";
 import { KernelAccountClient, KernelSmartAccount, createKernelAccountClient } from "@zerodev/sdk";
-import { BigNumberish } from "ethers";
 import { EntryPoint } from "permissionless/types/entrypoint";
 import type { Hash, HttpTransport, PublicClient, TypedDataDefinition } from "viem";
 import { Chain, http, publicActions } from "viem";
@@ -29,11 +29,6 @@ import { CrossmintWalletService } from "../../api/CrossmintWalletService";
 import { getBundlerRPC, getViemNetwork } from "../BlockchainNetworks";
 import { ERC20TransferType, SFTTransferType, TransferType } from "../token";
 import { paymasterMiddleware } from "./paymaster";
-
-type GasFeeTransactionParams = {
-    maxFeePerGas?: BigNumberish;
-    maxPriorityFeePerGas?: BigNumberish;
-};
 
 export class EVMAAWallet extends LoggerWrapper {
     public readonly chain: EVMBlockchainIncludingTestnet;
@@ -58,7 +53,7 @@ export class EVMAAWallet extends LoggerWrapper {
             chain: getViemNetwork(chain),
             entryPoint,
             bundlerTransport: http(getBundlerRPC(chain)),
-            ...(hasEIP1559Support(chain) && paymasterMiddleware({ entryPoint, chain })),
+            ...(!usesGelatoBundler(chain) && paymasterMiddleware({ entryPoint, chain })),
         });
         this.chain = chain;
         this.publicClient = publicClient;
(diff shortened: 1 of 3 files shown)


packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -87,11 +100,7 @@ export class EVMAAWallet extends LoggerWrapper {
 
     public async signTypedData(params: TypedDataDefinition) {
         return this.logPerformance("SIGN_TYPED_DATA", async () => {
-            try {
-                return await this.kernelClient.signTypedData(params);
-            } catch (error) {
-                throw new Error(`Error signing typed data. If this error persists, please contact support.`);
```
devlyn37, line 93: Dropping this try catch since smartAccountClient has duplicate error handling

devlyn37: Since all the other function that are on the chopping block have other logic that could still throw errors, I'm not removing anything from them yet.

packages/client/wallets/aa/src/blockchain/wallets/smartAccount.ts
```diff
@@ -0,0 +1,49 @@
+import { logInfo } from "@/services/logging";
+import { gelatoBundlerProperties, usesGelatoBundler } from "@/utils";
+import { SmartAccountClient } from "permissionless";
+import { EntryPoint } from "permissionless/types/entrypoint";
+
+import { EVMBlockchainIncludingTestnet } from "@crossmint/common-sdk-base";
+
+export function toCrossmintSmartAccountClient<Client extends SmartAccountClient<EntryPoint>>({
```
devlyn37, line 8: Anything missing from the methods we'll delete (the performance logging for example) will be added here in subsequent PRs.

packages/client/wallets/aa/src/blockchain/wallets/smartAccount.ts
```diff
@@ -0,0 +1,49 @@
...
+            } catch (error) {
+                throw new Error(`Error signing message. If this error persists, please contact support.`);
+            }
+        },
+        async signTypedData(data) {
+            try {
+                return smartAccountClient.signTypedData(data);
+            } catch (error) {
+                throw new Error(`Error signing typed data. If this error persists, please contact support.`);
+            }
```
devlyn37, line 29: I'm not sure we should be obfuscating the error like this, but copying behavior from existing functions for now. We can re-eval later.

packages/client/wallets/aa/src/blockchain/wallets/smartAccount.ts
```diff
@@ -0,0 +1,49 @@
...
+import { EVMBlockchainIncludingTestnet } from "@crossmint/common-sdk-base";
+
+export function toCrossmintSmartAccountClient<Client extends SmartAccountClient<EntryPoint>>({
+    crossmintChain,
+    smartAccountClient,
+}: {
+    crossmintChain: EVMBlockchainIncludingTestnet;
+    smartAccountClient: Client;
+}): Client {
+    return {
```
devlyn37, line 15: I'm not a huge fan of the style here, w/ the inline functions defined this way. But looking through `permissionless.js` and the ZD SDK this is the canon way to do this.

packages/client/wallets/aa/src/blockchain/wallets/smartAccount.ts
devlyn37, line 8: Eventually we can add functionality here.

packages/client/wallets/aa/src/blockchain/wallets/smartAccount.ts
maxisch: Out of curiosity, what other approach were you thinking of?

maxisch (approved):
Love this PR!

packages/client/wallets/aa/src/blockchain/wallets/smartAccount.ts
devlyn37: Something more similar to the `LoggerWrapper` we use, otherwise if `viem` and `permissionsless` we're more OOO style using decorators or something like that.

PR #511, opened 2024-06-10
Simplification: Client & Address Access

devlyn37:
## Description

This PR:
- Removes the `getSigner` function, replacing it with the `get client()` method. (I've annotated rational)
- Replaces the `getAddress` function in favor of `get client` for more ergonomic access

## Test plan

Passkey Release QA
TS Compiling & Existing Tests


packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -212,24 +212,14 @@ export class EVMAAWallet extends LoggerWrapper {
         });
     }
 
-    public getSigner(type: "viem" = "viem"): {
```
devlyn37, line 215: In AA land, "signer" usually applies to the underlying signing mechanism for an account. We even use it this way within `WalletConfig`. Check out [these docs](https://docs.pimlico.io/permissionless/how-to/accounts/use-kernel-account#create-the-signer) for `permissionless.js` the most popular library for AA.

`getSigner` conflicts with this, returning clients instead, likely confusing devs.

devlyn37: Additionally, given the key role that ZD SDK plays within this repo, our SDK is built on top of `viem`. Rather than surfacing other ways to interact with our SDK through different libs (`ethers.js`, `web3.js`, etc) here, let's double down on `viem`, and ship separate adapters if **really** needed.

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
```diff
@@ -212,24 +212,14 @@ export class EVMAAWallet extends LoggerWrapper {
...
-        publicClient: PublicClient;
-        walletClient: SmartAccountClient<EntryPoint, HttpTransport, Chain, SmartAccount<EntryPoint>>;
+    public get client(): {
```
devlyn37, line 215: Meant to interop w/ `viem` seamlessly. For example, with this method, interacting with a contract would look like:

```TS
const wallet = await sdk.getOrCreateWallet({userId: "joebob", jwt: "123"}, "base");
const contract = getContract({
  address: '0xFBA3912Ca04dd458c843e2EE08967fC04f3579c2',
  abi: nftCollection,
  client: wallet.client
})
const hash = await contract.mint()
```

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
devlyn37: Consider that `permissionless.js` the leading AA lib, and what the ZD SDK is built on, is built on `viem`

afeight: This makes a lot of sense–"signer" is an overloaded term we're using because we also have the "signer" that you use for `getOrCreateWallet`
and it was coming from the ethers analogy, less relevant now

afeight (approved):
makes a lot of sense to me

packages/client/wallets/aa/src/blockchain/wallets/EVMAAWallet.ts
alfonso-paella, line 215: Add JSDoc

devlyn37: done!

PR #512, opened 2024-06-10
Simplification: SDK & Wallet Naming

devlyn37:
## Description

This PR renames a couple key classes. I've annotated each change.

Rational for both changes:
- Bring naming closer to how we present the product in marketing & on our site.
- Name things more closely related to the functionality, rather than the technology used to achieve it.
    - For example, `Having AA` in the name doesn't make sense for non-evm chains

## Test plan

Passkey Release QA
TS Compiling & Existing Tests


packages/client/wallets/aa/src/SmartWalletSDK.ts
```diff
@@ -27,15 +27,11 @@ export class CrossmintAASDK extends LoggerWrapper {
...
+    static init(params: SmartWalletSDKInitParams): SmartWalletSDK {
+        return new SmartWalletSDK(params);
     }
 
-    async getOrCreateWallet(
-        user: UserIdentifierParams,
-        chain: EVMBlockchainIncludingTestnet,
-        walletConfig: WalletConfig
-    ) {
+    async getOrCreate(user: UserIdentifierParams, chain: EVMBlockchainIncludingTestnet, walletConfig: WalletConfig) {
```
afeight, line 34: this one I may push back on, even if the SDK has wallet in it, still feels more obvious what this does when it says that it creates wallets

afeight (commented):
big breaking changes lol
but they do make sense to me, barring `getOrCreateWallet`

afeight (approved):

packages/client/wallets/aa/src/SmartWalletSDK.ts
devlyn37: I added an explicit return type `Promise<EVMSmartWallet>`, does that sway you at all?

devlyn37: I think there's too much wallet in `SmartWalletSDK.getOrCreateWallet(...): Promise<EVMSmartWallet>` otherwise

packages/client/wallets/aa/src/SmartWalletSDK.ts
```diff
@@ -27,15 +27,15 @@ export class CrossmintAASDK extends LoggerWrapper {
...
     }
 
-    static init(params: CrossmintAASDKInitParams): CrossmintAASDK {
-        return new CrossmintAASDK(params);
+    static init(params: SmartWalletSDKInitParams): SmartWalletSDK {
+        return new SmartWalletSDK(params);
     }
 
-    async getOrCreateWallet(
+    async getOrCreate(
```
alfonso-paella, line 34: Im not sure this change makes sense to me

SmartWalletSDK.get() seems like it would return an SDK, not a wallet?

packages/client/wallets/aa/src/SmartWalletSDK.ts
devlyn37: reverted this change!

packages/client/wallets/aa/src/SmartWalletSDK.ts
devlyn37: reverted!

PR #513, opened 2024-06-11
Simplification: Further Pruning

devlyn37:
## Description

This PR removes unused or redundant pieces from the SDK.

## Test plan

TS Compiling & Existing Tests


packages/client/wallets/aa/src/blockchain/BlockchainNetwoks.test.ts
devlyn37, line 1: these tests all might as well read:

```TS
// In file x.ts
const foo = 5;

// in file x.test.ts
expect(foo).to.be(5);
```

devlyn37: You'll notice we weren't maintaining these tests either, see how most chains are missing

packages/client/wallets/aa/src/blockchain/BlockchainNetworks.ts
```diff
@@ -42,39 +41,6 @@ import { EVMBlockchainIncludingTestnet } from "@crossmint/common-sdk-base";
 
 import { logInputOutput } from "../utils/log";
 
-export function getFireblocksAssetId(chain: EVMBlockchainIncludingTestnet) {
```
devlyn37, line 45: Not used anymore

packages/client/wallets/aa/src/blockchain/plugins.ts
devlyn37, line 1: Non existent functionality atm

maxisch (approved):

PR #515, opened 2024-06-13
Auth: Window

MatiNavCrossmint:
## Description

Changes to allow cross-origin communication

## Test plan

https://www.loom.com/share/5a3edee2a7434270af50f3da417d98ca


packages/client/window/package.json
```diff
@@ -24,6 +24,7 @@
         "dev": "yarn clean && tsup src/index.ts --format esm,cjs --outDir ./dist --dts --sourcemap --watch"
     },
     "dependencies": {
-        "zod": "3.22.4"
+        "nanoid": "5.0.4",
```
afeight, line 27: what's `nanoid` do?

.changeset/serious-gifts-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@crossmint/client-sdk-window": major
+---
+
+Releasing Windows SDK to be used in cross origin communication
```
afeight, line 5: Windows sounds too much like the operating system 😅
```suggestion
Releasing Window SDK to be used in cross origin communication
```

.changeset/serious-gifts-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@crossmint/client-sdk-window": major
```
mPaella, line 2: Don't need a major version bump for this, it's more like a patch

packages/client/window/src/windows/Popup.ts
```diff
@@ -52,10 +53,9 @@ function createPopupString(width: number, height: number): string {
                 ? window.top.outerWidth / 2 + window.top.screenX - width / 2
                 : window.outerWidth / 2 + window.screenX - width / 2;
         } catch (e) {
-            console.error(e);
+            console.log("Error could be due to cross-origin policy", e);
```
guiwoda, line 56: is there a way to check for cross-origin before, instead of calculating this alternative in a catch block?

guiwoda: What I mean is, I'd rather have explicit code branches because of a known possible error, than expecting an error to be thrown and reacting to it.

afeight (approved):
+1 to woda's comment, but if it works then this looks good

afeight:
tested CSS changes on a project w/o tailwind: 
image

PR #516, opened 2024-06-13
Auth: Core

MatiNavCrossmint:
## Description

Implement the login to Crossmint using cross-origin communication

## Test plan

https://www.loom.com/share/5a3edee2a7434270af50f3da417d98ca


packages/client/auth-core/src/contexts/authlogin.tsx
```diff
@@ -0,0 +1,109 @@
...
+    const [modalOpen, setModalOpen] = useState(false);
+    const [user, setUser] = useState<User | null>(null);
+    const crossmintService = useMemo(() => new CrossmintService(apiKey, jwtToken, environment), [apiKey, jwtToken, environment]);
+
+    useEffect(() => {
+        const crossmintSession = document.cookie.split('; ').find(row => row.startsWith('crossmint-session'));
+        const crossmintSessionCookie = crossmintSession ? crossmintSession.split('=')[1] : null;
+        if (crossmintSessionCookie) {
+            setJwtToken(crossmintSessionCookie);
+            // here we should check how to get the user data
```
guiwoda, line 49: Should we add a ticket here?

packages/client/auth-core/src/contexts/authlogin.tsx
```diff
@@ -0,0 +1,109 @@
...
+        if (jwtToken == null) {
+            throw new Error("jwtToken is not set");
+        }
+
+        return crossmintService.fetchCrossmintAPI(fetchCrossmintParams);
+    }, [jwtToken, environment]);
+
+    useEffect(() => {
+        if (jwtToken) {
+            // todo make endpoint in crossmint that returns the user associated with the jwtToken
```
guiwoda, line 83: Ticket?

packages/client/auth-core/src/services/CrossmintService.ts
```diff
@@ -0,0 +1,59 @@
...
+        this.crossmintAPIHeaders = {
+            accept: "application/json",
+            "content-type": "application/json",
+            authorization: `Bearer ${jwtToken}`,
+            "x-api-key": apiKey,
+        };
+        this.crossmintBaseUrl = this.getUrlFromEnv(environment ?? "production");
+    }
+
+    async fetchCrossmintAPI({ endpoint, options = { method: "GET" }, onServerErrorMessage }: FetchCrossmintParams) {
```
guiwoda, line 30: Can we use Swagger / OpenAPI to generate a better client for our APIs?

packages/client/auth-core/src/utils/constants.ts
```diff
@@ -0,0 +1,11 @@
+// testing purposes
+export const CROSSMINT_STG_URL = "http://localhost:3000";
+// export const CROSSMINT_STG_URL = "https://staging.crossmint.com";
```
guiwoda, line 3: Should we leave this instead?
Maybe we could find a way to inject the testing values without committing them

packages/client/auth-core/src/utils/jwt.ts
```diff
@@ -0,0 +1,25 @@
+import { CrossmintService } from "@/services/CrossmintService";
+
+//TODO: this should resolve the user
```
guiwoda, line 3: tkt

guiwoda (commented):
Left a few comments! Will ping you offline

packages/client/auth-core/src/contexts/authlogin.tsx
```diff
@@ -0,0 +1,109 @@
...
+import AuthModal from "@/components/AuthModal";
+import { CrossmintService, FetchCrossmintParams } from "@/services/CrossmintService";
+import { CrossmintEnvironment } from "@/utils";
+import { ReactNode, createContext, useCallback, useContext, useEffect, useMemo, useRef, useState } from "react";
+
+type AuthContextType = {
+    user: User | null;
+    login: () => void;
+    logout: () => void;
+    fetchCrossmintAPI: (fetchCrossmintParams: FetchCrossmintParams) => Promise<any> | void;
```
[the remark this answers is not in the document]
MatiNavCrossmint: As the response depends on the endpoint, we cannot know beforehand what we expect to receive. We could use a generic type, where the client needs to specify it, but I would let that as an improvement as it would mean using a seemed strategy to the one used in the window package.

BTW, I assume this because on the CrossmintService used on AA we return a promise<any> too, so I think we can live with this just for the moment.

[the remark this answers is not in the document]
MatiNavCrossmint: The `fetchCrossmintAPI` was thought for internal purposes. As of now, this will be used by external devs, this function doesn't make sense anymore

packages/client/auth-core/src/services/CrossmintService.ts
MatiNavCrossmint: The `fetchCrossmintAPI` was thought for internal purposes. As of now, this will be used by external devs, this function doesn't make sense anymore.

packages/client/auth-core/src/utils/constants.ts
MatiNavCrossmint: We should remove it and add it when testing/developing, for sure!

packages/client/auth-core/src/services/CrossmintService.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export class CrossmintService {
+    protected crossmintAPIHeaders: Record<string, string>;
+    public readonly crossmintBaseUrl: string;
+    private static urlMap: Record<CrossmintEnvironment, string> = {
+        staging: CROSSMINT_STG_URL,
+        production: CROSSMINT_PROD_URL,
+    };
+
+    constructor(apiKey: string, jwtToken: string | null, environment?: CrossmintEnvironment) {
```
mPaella, line 20: you're not validating the api key, which can be done with `validateAPIKey(apiKey)`, the return of this tells you what environment the apikey is for - so you should use that and drop the environment param completely

guiwoda (approved):
LGTM

packages/client/auth-core/src/services/CrossmintService.ts
mPaella: actually, its a big smell that this code is nearly exactly the same as the CrossmintService in the aa sdk

which is why i made the `ApiClient` and `CrossmintApiClient` classes in the common sdk, pls use those classes! rather than having 3 different services that do ~the same thing, and all of us working separately on them, we should be centralizing it to these new classes

see this pr https://github.com/Crossmint/crossmint-sdk/pull/609/files

packages/client/auth-core/src/components/ActionModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+    show,
+    onClose = () => {},
+    children,
+}: ActionModalProps & { children: ReactNode }) {
+    // WAL-2574 - because of a recent change from another sdk in @headlessui/react, this file started complaining about the types
+    return (
+        <Transition.Root show={show} as={Fragment}>
+            <Dialog
+                as="div"
+                className="fixed inset-0 z-20 flex items-center justify-center overflow-y-auto"
```
mPaella, line 19: pretty sure these tailwind classes will not work in multiple cases:
1.) if the dev isnt using tailwind
2.) if they are: normally tailwind compiles to the smallest possible css file, so even if the dev is using tailwind these specific classes might not be in the outputted css file

MatiNavCrossmint: Perfect, thanks for highlighting it!

afeight:
tested CSS changes on a project w/o tailwind ✅
image

packages/client/auth-core/src/components/ActionModal.tsx
afeight: updated to vanilla css

packages/client/auth-core/src/services/CrossmintService.ts
afeight: this is a great point, will address in a separate PR though to unblock
[WAL-2620 : Update `CrossmintService` to use `CrossmintApiClient`](https://linear.app/crossmint/issue/WAL-2620/update-crossmintservice-to-use-crossmintapiclient)

packages/client/auth-core/package.json
```diff
@@ -0,0 +1,59 @@
+{
+    "name": "@crossmint/client-sdk-auth-core",
+    "version": "0.0.1",
```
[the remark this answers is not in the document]
afeight: Good Q, following up in slack

PR #520, opened 2024-06-18
embed: fix css

mPaella:
## Description

this issue presents in multiple live production clients, for example https://www.alexandriabooks.com element is not filling full width

image

coinbase, not filling full width
image

## Test plan

**before, css styles with `!important` not being applied at all:**
image

**now, styles apply correctly**
image


packages/client/ui/react-ui/src/components/embed/CrossmintPaymentElement.test.tsx
```diff
@@ -17,13 +17,13 @@ describe("CrossmintPaymentElement", () => {
         expect(iframe).toHaveAttribute("src");
         expect(iframe).toHaveAttribute("id", "crossmint-embedded-checkout.iframe");
         expect(iframe).toHaveStyle({
-            padding: "0px !important",
```
mPaella, line 20: multiple sources online site that passing `!important` through the style prop does not work, which is why this is not working...

pgarciaegido (approved):

PR #522, opened 2024-06-21
VC: credentials enpoint

vmc-crossmint:
## Description
Updated sdk to work with the new credential flow.
- Use the contract metadata to retrieve the credentials enpoint:
 - if crossmint proxy through crossmint apis
 - else get from ipfs
- Added credential sanity checks
- improved typing 
## Test plan
unit tests wip


packages/client/verifiable-credentials/src/presentation/getCredential.ts
```diff
@@ -38,3 +66,28 @@ export async function getCredentialFromId(
...
+    collection: CredentialsCollection,
+    tokenId: string,
+    environment: string
+): Promise<VerifiableCredentialType | null> {
+    const metadata = collection.metadata;
+    if (!isVerifiableCredentialContractMetadata(metadata)) {
+        throw new Error(`The collection provided is not a verifiable credential collection`);
+    }
+    const retrievalEndpoint = metadata.credentialMetadata.credentialsEndpoint;
+    // shall we route ipfs requests through the crossmint API?
```
vmc-crossmint, line 80: Once we decouple the crossmint part, it makes sense to have two different flows and when using crossmint probably we deliver a better ux proxying the request. For now maybe is better to leave it like this

alberto-crossmint (approved):

PR #523, opened 2024-06-25
Smart Wallets: Passkey Signer Base

devlyn37:
## Description

This PR lays the foundation for the passkey signer, adding a functional passkey signer option to the SDK! This PR does not include:
- User based JWT authentication and input parameter changes
- Robust error handling as defined within the SDK spec
- Nice JSDoc comments

Those will be handled in separate PRs.

## Test plan

Local Testing with SCW Demo:
- Created a new passkey wallet and went through the mint flow.
- In a new session logged on, and went through mint flow, confirming only a single passkey prompt was required.

Passkey Release QA.

## Release

We won't release a new version of the SDK until:
- The other pieces I mention above have landed
- We've done the "Passkey Release QA"


packages/client/wallets/aa/src/types/internal.ts
devlyn37, line 1: No need to expose this type since it's internal use only

packages/client/wallets/aa/src/blockchain/wallets/passkey.ts
```diff
@@ -0,0 +1,111 @@
...
+
+        if (signer.type !== "passkeys") {
+            throw new Error("Admin Mismatch"); // TODO custom error as defined within SDK spec
+        }
+
+        return serializePasskeyValidatorData(signer);
+    }
+
+    private getSignerData(validator: PasskeyValidator, passkeyName: string): PasskeySignerData {
+        const fields = deserializePasskeyValidatorData(validator.getSerializedData());
```
devlyn37, line 103: In an SDK update ZD has improved how this works iirc, so we should be able to cleanup. The upgrade will be non-trivial for other reasons, so leaving what feels like excess (de)serialization here for now.

PR #526, opened 2024-06-26
Smart Wallets: Restrict To Whitelabel

devlyn37:
## Description

This PR:
- Updates the user param within `SmartWalletSDK.getOrCreateWallet` according to the updated SDK spec.
- This limits the usage of the SDK to white label only for now.
- Drops some related code.

## Test plan

I keep a branch of the SCW demo site updated with changes to the SDK, Locally I:
- Updated the branch
- Ran through wallet creation and minting flows.

Further testing will be performed during Passkey release QA.

## Release

This will rollout along with the full passkeys release.


packages/client/wallets/aa/src/utils/user.ts
devlyn37, line 1: Not used anymore, and should probably be in the shared within the base sdk!

packages/client/wallets/aa/src/types/Config.ts
```diff
@@ -6,16 +6,15 @@ export type SmartWalletSDKInitParams = {
...
-    email?: string;
-    userId?: string;
-    phoneNumber?: string;
+type WhitelabelUserParams = {
+    /**
+     * A unique identifier for the user. This must match the value of the identifier within the JWT
+     * that is specified in the project settings (typically `sub`).
+     */
+    id: string;
+    jwt: string;
```
devlyn37, line 15: Only used as a placeholder for now!

maxisch (approved):
nice, makes sense to keep this lean while we only support white label

packages/client/wallets/aa/src/types/Config.ts
```diff
@@ -6,16 +6,15 @@ export type SmartWalletSDKInitParams = {
...
 };
-
-export type UserIdentifier =
-    | { type: "whiteLabel"; userId: string }
-    | { type: "email"; email: string }
-    | { type: "phoneNumber"; phoneNumber: string };
+export type UserParams = WhitelabelUserParams;
```
afeight, line 17: thinking through future-proofness of this
should we keep a label (e.g. `{ type: "whiteLabel";`) on each scenario so that we can still easily `switch` on it when we need to handle?

afeight (approved):
one comment, otherwise looking good

packages/client/wallets/aa/src/types/Config.ts
devlyn37: Going to land this PR and circle back on this before release, I think there's a way we can both:
- Have a good default passkey name from the user
- Drop that `id` property, only passing in the JWT in `UserParams`

So devs could call `SmartWalletSDK.getOrCreateWallet({jwt}, "base")`, and we could derive a bunch of details from the JWT:
- The id, like you mentioned
- Whether or not it should be grey label based on the issuer of the token.

In which case we won't need the label anymore

PR #528, opened 2024-06-28
Simplification: SDK Surface Audit 1

devlyn37:
## Description

When using the SDK I've noticed we (unintentionally?) surface many unnecessary parts. This makes the surface of the SDK huge and easy to break.

This PR:
- Removes all our constants (`ZD_AMOY_PROJECT_ID`, etc) from the package export
- Drops an unused function.

## Test plan

Build passing & existing tests + passkey release QA

### DIFF
--- a/packages/client/wallets/aa/src/utils/blockchain.ts
+++ b/packages/client/wallets/aa/src/utils/blockchain.ts
@@ -1,26 +1,4 @@
-import { getBundlerRPC } from "@/blockchain";
-import { createPublicClient, http } from "viem";
-
 import { EVMBlockchainIncludingTestnet } from "@crossmint/common-sdk-base";
-
-export type VerifyMessageParams = {
-    address: `0x${string}`;
-    message: string;
-    signature: `0x${string}` | Uint8Array;
-    chain: EVMBlockchainIncludingTestnet;
-};
-
-export async function verifyMessage({ address, message, signature, chain }: VerifyMessageParams) {
-    const publicClient = createPublicClient({
-        transport: http(getBundlerRPC(chain)),
-    });
-
-    return publicClient.verifyMessage({
-        address,
-        message,
-        signature,
-    });
-}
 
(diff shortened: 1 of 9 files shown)


packages/client/wallets/aa/src/utils/blockchain.ts
```diff
@@ -1,27 +1,5 @@
...
-    const publicClient = createPublicClient({
-        transport: http(getBundlerRPC(chain)),
-    });
-
-    return publicClient.verifyMessage({
-        address,
-        message,
-        signature,
-    });
-}
```
devlyn37, line 23: Instead call wallet.client.public.verifyMessage

packages/client/wallets/aa/src/utils/blockchain.ts
```diff
@@ -1,27 +1,5 @@
...
 import { EVMBlockchainIncludingTestnet } from "@crossmint/common-sdk-base";
 
-export type VerifyMessageParams = {
-    address: `0x${string}`;
-    message: string;
-    signature: `0x${string}` | Uint8Array;
-    chain: EVMBlockchainIncludingTestnet;
-};
-
-export async function verifyMessage({ address, message, signature, chain }: VerifyMessageParams) {
```
afeight, line 13: I think we want this, or at least ethers verify message didn't work with AAs, so we added this helper

afeight (approved):
most looks good, think we want to keep the verify helper though

packages/client/wallets/aa/src/utils/blockchain.ts
devlyn37: We expose a `public client` from `EVMSmartWallet` now, so `wallet.public.client.verifyMessage` would be equivalent.

devlyn37: We gotta add a doc example though, I think that's super obvious

devlyn37: going to merge this, can circle back before the release if you'd like

PR #532, opened 2024-07-02
VC: retrieve delegation sig

vmc-crossmint:
## Description
Retrieve the delegation signature from crossmint when not provided.
GET /api/unstable/credentials/delegationSignature

## Test plan
Tested manually

### DIFF
--- a/packages/client/verifiable-credentials/src/services/lit.ts
+++ b/packages/client/verifiable-credentials/src/services/lit.ts
@@ -1,7 +1,10 @@
+import { AuthSig } from "@lit-protocol/types";
+
 import { APIKeyUsageOrigin } from "@crossmint/common-sdk-base";
 
 import { crossmintAPI } from "../crossmintAPI";
 import { LitNetwork, Lit as LitRaw } from "../verifiableCredentialsSKD";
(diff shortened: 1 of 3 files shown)


packages/client/verifiable-credentials/src/services/lit.ts
```diff
@@ -25,6 +28,14 @@ export class Lit extends LitRaw {
             );
         }
 
-        super(network, capacityDelegationAuthSig, debug);
+        super(network, capacityDelegationAuthSig as AuthSig, debug);
```
alberto-crossmint, line 31: make it optional also in the parent?

packages/client/verifiable-credentials/src/services/lit.ts
```diff
@@ -25,6 +28,14 @@ export class Lit extends LitRaw {
...
+    }
+
+    async decrypt(base64Ciphertext: string): Promise<string> {
+        if (this.capacityDelegationAuthSig == null) {
+            console.debug("No capacity delegation auth sig provided, retrieving from Crossmint");
```
alberto-crossmint, line 36: warn the user that they will pay for the operation

alberto-crossmint (approved):

PR #536, opened 2024-07-03
Fix: Installs While Using Node 20

devlyn37:
## Description

It's impossible to install dependancies using node 20 right now. The error lists the dependency responsible, so I tracked down it's parent a few layers up, uninstalled and reinstalled to upgrade `^x.x.x` dependancies.
 

## Test plan

```bash
nvm use 20
yarn install
```

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -90,6 +90,14 @@
     "@babel/highlight" "^7.22.13"
     chalk "^2.4.2"
 
+"@babel/code-frame@^7.24.7":
+  version "7.24.7"
+  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.24.7.tgz#882fd9e09e8ee324e496bd040401c6f046ef4465"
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -12336,18 +12754,20 @@ eslint-plugin-jest@^25.3.0:
   dependencies:
     "@typescript-eslint/experimental-utils" "^5.0.0"
 
-eslint-plugin-jsdoc@^40.0.0:
-  version "40.3.0"
-  resolved "https://registry.yarnpkg.com/eslint-plugin-jsdoc/-/eslint-plugin-jsdoc-40.3.0.tgz#75a91ab71c41bb797db05a32d9528ce3ab613e90"
-  integrity sha512-EhCqpzRkxoT2DUB4AnrU0ggBYvTh3bWrLZzQTupq6vSVE6XzNwJVKsOHa41GCoevnsWMBNmoDVjXWGqckjuG1g==
+eslint-plugin-jsdoc@^46.9.0:
```
devlyn37, line 12757: This was the culprit

devlyn37:
> confirm that `yarn install` and `yarn build` now work for at least the AA package!

done

PR #537, opened 2024-07-03
Smart Wallets: Throw Nice Error When SDK Used Serverside (Closes WAL-2459, WAL-2460)

devlyn37:
## Description

From beta testing feedback, this PR fixes two issues related to the SDK running server side.

### Calling `window` in the module

We have the following code in the SDK.

```TS
function getBrowserLogger() {
    if (isLocalhost()) {
        return new ConsoleProvider();
    }

    return new DatadogProvider();
}

const { logInfo, logWarn, logError } = getBrowserLogger();
```

Which will run whenever the module is loaded. When used with Next.js (or maybe the latest versions of react in general, I'm not super familiar aha), this means even if we have code that only runs client side:

Image 7-2-24 at 8 17 PM

We get errors that look like this:

Image 7-2-24 at 8 16 PM

The issue is that `isLocalhost` accesses `window`. This PR adds a check before.

### A Nice Error

This PR also adds a nice error in the case that `SmartWalletSDK.init` is run server side, which is nicer than exposing the downstream failures that running on the server cause.

## Test plan

In a `create-next-app` starter, with the SDK installed, made sure that:
- With the code pictured above, there we're no errors
- When running `SmartWalletSDK.init` serverside, we throw the new error as expected.

### DIFF
--- a/packages/client/wallets/aa/src/utils/error.ts
+++ b/packages/client/wallets/aa/src/utils/error.ts
@@ -5,8 +5,6 @@ export class NotAuthorizedError extends Error {
 
     constructor(message: string) {
         super(message);
-
-        // ES5 workaround
         Object.setPrototypeOf(this, NotAuthorizedError.prototype);
     }
 }
@@ -16,8 +14,6 @@ export class TransferError extends Error {
 
     constructor(message: string) {
         super(message);
-
-        // ES5 workaround
         Object.setPrototypeOf(this, TransferError.prototype);
     }
 }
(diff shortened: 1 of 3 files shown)


packages/client/wallets/aa/src/utils/error.ts
```diff
@@ -27,8 +23,6 @@ export class TransactionError extends Error {
 
     constructor(message: string) {
         super(message);
-
-        // ES5 workaround
```
devlyn37, line 31: Seems redundant to have this everytime

PR #539, opened 2024-07-03
AA: increase timeout on paymaster

MatiNavCrossmint:
## Description

increase timeout to paymaster

## Test plan

Createad AA and minted locally


afeight (approved):
nice

PR #541, opened 2024-07-04
AA: Hide console logs locally

MatiNavCrossmint:
## Description

Hide console logs locally

## Test plan

Created AA wallet and minted locally (on branch that works with the demo)


devlyn37 (commented):
Ah just looked through the code and looks like it for staging & prod we don't log on the console at all. So logging sensitive pieces wouldn't be an issue anyways.

I think we should make this a toggle-able somehow so that:
- developers can still send us logs when they run into issues.
- For local dev, it's more obvious that this is an option.

devlyn37 (approved):

PR #542, opened 2024-07-05
WAL-2484: Set JWT in cookie

maxisch:
## Description

Set the JWT in a cookie for the passkeys server to consume.

Crossbit-main PR: https://github.com/Paella-Labs/crossbit-main/pull/13853

## Test plan

Creating a user in the scw_demo page.

### DIFF
--- a/packages/client/wallets/aa/src/blockchain/wallets/passkey.ts
+++ b/packages/client/wallets/aa/src/blockchain/wallets/passkey.ts
@@ -33,6 +33,8 @@ export class PasskeyWalletService {
     constructor(private readonly crossmintService: CrossmintWalletService) {}
 
     public async getOrCreate({ user, chain, publicClient, walletConfig, entrypoint }: PasskeyWalletParams) {
+        this.setJwtCookie(user.jwt);
+
         const validator = await this.getOrCreateSigner({
             user,
             entrypoint,
@@ -78,10 +80,10 @@ export class PasskeyWalletService {
         }
 
         return createPasskeyValidator(publicClient, {
-            passkeyServerUrl: this.crossmintService.getPasskeyServerUrl(user),
+            passkeyServerUrl: this.crossmintService.getPasskeyServerUrl(),
             entryPoint: entrypoint.address,
             passkeyName: signer.passkeyName,
-            credentials: "omit",
+            credentials: "include",
         });
     }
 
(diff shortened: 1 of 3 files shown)


packages/client/wallets/aa/src/blockchain/wallets/passkey.ts
```diff
@@ -103,4 +105,10 @@ export class PasskeyWalletService {
             type: "passkeys",
         };
     }
+
+    private setJwtCookie(jwt: string) {
+        const now = new Date();
+        const expireTime = new Date(now.getTime() + 1 * 60 * 60 * 1000); // Now + 1 hour
+        document.cookie = `user_jwt=${jwt}; expires=${expireTime.toUTCString()}; path=/`;
+    }
```
devlyn37, line 113: If we're not running on localhost, we should use the `secure` property here.

packages/client/wallets/aa/src/blockchain/wallets/passkey.ts
```diff
@@ -103,4 +105,10 @@ export class PasskeyWalletService {
             type: "passkeys",
         };
     }
+
+    private setJwtCookie(jwt: string) {
+        const now = new Date();
+        const expireTime = new Date(now.getTime() + 1 * 60 * 60 * 1000); // Now + 1 hour
+        document.cookie = `user_jwt=${jwt}; expires=${expireTime.toUTCString()}; path=/`;
```
devlyn37, line 112: If you don't specify it, this cookie will only be sent to servers w/ the domain of the client that set it. When `harrypotter.com` uses SDK, `crossmint.com` won't receive the cookies.

packages/client/wallets/aa/src/blockchain/wallets/passkey.ts
```diff
@@ -103,4 +105,10 @@ export class PasskeyWalletService {
             type: "passkeys",
         };
     }
+
+    private setJwtCookie(jwt: string) {
+        const now = new Date();
+        const expireTime = new Date(now.getTime() + 1 * 60 * 60 * 1000); // Now + 1 hour
```
devlyn37, line 111: maybe worth defining a const `JWT_COOKIE_EXPIRY_TIME` or something like that.

devlyn37 (approved):
Left a couple important comments to check out before landing!

packages/client/wallets/aa/src/blockchain/wallets/passkey.ts
maxisch: I read a bit on this, what options do we have? We can't set cookies cross domains. Should we set the cookie server side (for crossmint.com) on the passkeys server on let's say the first interaction?

devlyn37: That sounds like a good idea to me, maybe we could have some "handshake" endpoint on the passkeys controller to set the cookie

maxisch:
Closing as we are not moving forward with this approach

PR #543, opened 2024-07-06
Simplification: SDK Surface Audit 3

devlyn37:
## Description

This PR completes the SDK surface audit, makes all package exports explicit, and removes all but the root `./index.ts` files. This makes sure we don't unintentionally export anything, and that we have a clear view of the entire SDK surface.

This PR also moves around a functions & types.

## Test plan

TS Compiling & Existing Tests


packages/client/wallets/aa/src/blockchain/BlockchainNetworks.ts
```diff
@@ -72,9 +70,9 @@ export const getUrlProviderByBlockchain = logInputOutput((chain: EVMBlockchainIn
         throw new Error(`Url provider not found for chain ${chain}`);
     }
     return url;
-}, "getUrlProviderByBlockchain");
+};
 
-export const getBlockExplorerByBlockchain = logInputOutput((chain: EVMBlockchainIncludingTestnet) => {
```
maxisch, line 77: why did we remove these? `logInputOutput`

maxisch (approved):

packages/client/wallets/aa/src/blockchain/BlockchainNetworks.ts
devlyn37: Ah, I forgot hit submit on a comment I wrote when I put this PR up:

> To try and clear some cyclical dep build issues I was running into, I dropped logging from these functions.
> 
> Going to leave this change in since logging everything here is probably overkill

PR #544, opened 2024-07-08
AA: Toggle logging on datadog

MatiNavCrossmint:
## Description

Add toggle to let external developer to log on Datadog while developing locally to let us help them

## Test plan

Tested on stable version creating an AA Wallet and minting.

### DIFF
--- a/packages/client/wallets/aa/src/SmartWalletSDK.ts
+++ b/packages/client/wallets/aa/src/SmartWalletSDK.ts
@@ -9,6 +9,7 @@ import { createPublicClient, http } from "viem";
 import { EVMBlockchainIncludingTestnet, UserIdentifierParams, validateAPIKey } from "@crossmint/common-sdk-base";
 
 import EOAWalletService from "./blockchain/wallets/eoa";
+import { logger } from "./services/logging";
 import { LoggerWrapper, logPerformance } from "./utils/log";
 import { parseUserIdentifier } from "./utils/user";
 
@@ -17,6 +18,7 @@ export class SmartWalletSDK extends LoggerWrapper {
     private readonly eaoWalletService: EOAWalletService;
 
     private constructor(config: SmartWalletSDKInitParams) {
+        logger.setDDLoggerOnLocalhost(Boolean(config.logging));
         super("SmartWalletSDK");
         this.crossmintService = new CrossmintWalletService(config.clientApiKey);
         this.eaoWalletService = new EOAWalletService(this.crossmintService);
--- a/packages/client/wallets/aa/src/blockchain/wallets/EVMSmartWallet.ts
+++ b/packages/client/wallets/aa/src/blockchain/wallets/EVMSmartWallet.ts
@@ -1,4 +1,4 @@
-import { logError } from "@/services/logging";
+import { logger } from "@/services/logging";
 import { SCW_SERVICE, TransferError, errorToJSON } from "@/utils";
 import { LoggerWrapper } from "@/utils/log";
 import { KernelAccountClient, KernelSmartAccount, createKernelAccountClient } from "@zerodev/sdk";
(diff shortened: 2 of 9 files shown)


packages/client/wallets/aa/src/SmartWalletSDK.ts
```diff
@@ -17,6 +18,7 @@ export class SmartWalletSDK extends LoggerWrapper {
     private readonly eaoWalletService: EOAWalletService;
 
     private constructor(config: SmartWalletSDKInitParams) {
+        logger.setDDLoggerOnLocalhost(Boolean(config.logging));
```
afeight, line 21: `Boolean(config.logging)` is kinda weird, more idiomatic would be:
```suggestion
        logger.setDDLoggerOnLocalhost(config.logging === true);
```

afeight: Rather than decide "yes I want logging" or "no I don't want logging", much more extensible in my mind to be able to provide a logger.
Don't want logging? => provide `undefined`, omit
Want logging to console? => provide `ConsoleLogger` (maybe we need to provide)
Want logging to your production logs but not the console? => Provide a custom logger that hooks up to your service

packages/client/wallets/aa/src/services/logging/index.ts
```diff
@@ -1,16 +1,49 @@
...
+    public static getInstance() {
+        if (!LoggerProvider.instance) {
+            LoggerProvider.instance = new LoggerProvider();
+        }
+        return LoggerProvider.instance;
+    }
+}
 
-export { logInfo, logWarn, logError };
+export const logger = LoggerProvider.getInstance();
```
afeight, line 49: I've got a bunch of thoughts here:
1. I don't think `export const logger` adds value, we already have the singleton set up, so I think better to just allow people to use `LoggerProvider.getLogger()`–this way the rest of the code looks more natural w/ dependency injection
2. `logOnDatadogOnLocalhost`–doesn't seem like the pattern we want relative to being able to provide a logger
3. Rather than this `LoggerProvider` having all the `logInfo`, `logWarning` set on itself, it'd be cleaner if it lived up to its name, and just provided a `Logger`, with the ability to set what that logger is. So together I'd be thinking more like:

```
interface Logger {
    logInfo: (message: string, context?: object) => void;
    logWarn: (message: string, context?: object) => void;
    logError: (message: string, context?: object) => void;
}

export class LoggerProvider {
    private static instance: LoggerProvider;
    private logger: Logger

    private constructor() {
        const defaultLogger = this.shouldLogToDatadog() ? new DatadogProvider() : new EmptyLogger()
        this.logger = defaultLogger;
    }

    public static setLogger(logger: Logger) {
        LoggerProvider.instance.logger = logger;
    }

    public static getLogger() {
        return LoggerProvider.instance.logger!
    }

    private shouldLogToDatadog() {
        return !isLocalhost();
    }

    public static getInstance() {
        if (!LoggerProvider.instance) {
            LoggerProvider.instance = new LoggerProvider();
        }
        return LoggerProvider.instance;
    }
}

class EmptyLogger implements Logger {
    logInfo()  {}
    logWarn()  {}
    logError()  {}
}

File: packages/client/wallets/aa/src/blockchain/wallets/EVMSmartWallet.ts [hunk 3]
```diff
@@ -111,7 +111,7 @@ export class EVMSmartWallet extends LoggerWrapper {
                 const { request } = await client.simulateContract(tx);
                 return client.writeContract(request);
             } catch (error) {
-                logError("[TRANSFER] - ERROR_TRANSFERRING_TOKEN", {
+                logger.logError("[TRANSFER] - ERROR_TRANSFERRING_TOKEN", {
```
afeight, line 114: and then with that change, I'd expect `LoggerWrapper` to have a property:
```
logger: Logger = LoggerProvider.getLogger()
```
so that this code stays, but without the global var 
`import { logger } from "@/services/logging";`

packages/client/wallets/aa/src/utils/auth.ts
```diff
@@ -11,7 +11,7 @@ export const parseToken = (token: any) => {
             typeof window !== "undefined" ? window.atob(base64) : Buffer.from(base64, "base64").toString();
         return JSON.parse(jsonPayload || "");
     } catch (err) {
-        logError("[PARSE_TOKEN] - ERROR", {
+        logger.logError("[PARSE_TOKEN] - ERROR", {
```
afeight, line 14: but in other instances, this would be fine to do `LoggerProvider.getLogger().logError(...)`

PR #545, opened 2024-07-08
WAL-2455: Remove id from userparams

maxisch:
## Description

Removing id from User Params and inferring the passkeys name from the JWT token. 

## Test plan

Create a user with a passkey.

This is how it looks like
image

### DIFF
--- a/packages/client/wallets/aa/src/types/Config.ts
+++ b/packages/client/wallets/aa/src/types/Config.ts
@@ -7,11 +7,6 @@ export type SmartWalletSDKInitParams = {
 };
 
 export type UserParams = {
-    /**
-     * A unique identifier for the user. This must match the value of the identifier within the JWT
-     * that is specified in the project settings (typically `sub`).
-     */
-    id: string;
     jwt: string;
 };
 
@@ -30,11 +25,6 @@ export type ViemAccount = {
 
 export type PasskeySigner = {
     type: "PASSKEY";
-
-    /**
-     * Displayed to the user during passkey registration or signing prompts.
-     */
-    passkeyName: string;
 };
 
 export type EOASigner = EIP1193Provider | Web3AuthSigner | ViemAccount;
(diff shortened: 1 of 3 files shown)


packages/client/wallets/aa/src/types/Config.ts
```diff
@@ -30,11 +25,6 @@ export type ViemAccount = {
 
 export type PasskeySigner = {
     type: "PASSKEY";
-
-    /**
-     * Displayed to the user during passkey registration or signing prompts.
-     */
-    passkeyName: string;
```
devlyn37, line 37: I think we should still be able to specify this, with the userId derived from the jwt acting as the default. That said, I always like having less stuff, wdyt?

devlyn37 (commented):
Left one question

packages/client/wallets/aa/src/types/Config.ts
maxisch: If we do this, then I believe we have to switch the whole approach to obtain the userId client side instead of server side, right? Something like a pre step os checking if no passkeyName was submitted, then parse the jwt, check the jwtConfig of the api key (not sure we expose that now) to see what we use as a claim, and then proceed with that. I guess we can always do that later if we want to and start this way? No strong opinion tbh, just that one because the work is already done haha

maxisch: Mmmm maybe there is another option. We could expose an optional passkeyName here where you commented, and internally given that ZD sends a required passkeyName (the `no-op` string I chose) If null we can send an empty string (or the no-op string) as an indicator to the backend to infer the passkey name to the jwt. And we don't have to port that logic to the client. 
How does that sound?

devlyn37: perfect

maxisch:
Closing in favor of a fresh start here: https://github.com/Crossmint/crossmint-sdk/pull/550

PR #547, opened 2024-07-09
AA JWT Cookie: Storing server side

maxisch:
## Description

Attempt to store cookie server side. 
Even though I see the `Set-Cookie: ...` header, nothing is stored in the browser.

## Test plan

Manual testing


maxisch:
Closing as this third party cookie approach will never work

PR #550, opened 2024-07-11
WAL-2455: Remove id from userparams

maxisch:
## Description

Removing id from User Params and making the passkeyName optional. If not provided, infer from the JWT token (projectUserId).

## Test plan

Create a user with a passkey.

### DIFF
--- a/packages/client/wallets/aa/src/types/Config.ts
+++ b/packages/client/wallets/aa/src/types/Config.ts
@@ -7,11 +7,6 @@ export type SmartWalletSDKInitParams = {
 };
 
 export type UserParams = {
-    /**
-     * A unique identifier for the user. This must match the value of the identifier within the JWT
-     * that is specified in the project settings (typically `sub`).
-     */
-    id: string;
     jwt: string;
 };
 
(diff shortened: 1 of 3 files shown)


packages/client/wallets/aa/src/types/Config.ts
```diff
@@ -33,8 +28,9 @@ export type PasskeySigner = {
 
     /**
      * Displayed to the user during passkey registration or signing prompts.
+     * If not provided, a default name will be used.
```
devlyn37, line 31: would be useful to mention that the default comes from the jwt

packages/client/wallets/aa/src/types/Config.ts
```diff
@@ -33,8 +28,9 @@ export type PasskeySigner = {
...
      */
-    passkeyName: string;
+    passkeyName?: string;
```
devlyn37, line 33: It would be good to add a default `walletConfig` to `getOrCreateWallet`, with passkeys set to the default so clients can make calls that look like

```TS
sdk.getOrCreateWallet({ jwt }, "base")
```

devlyn37 (approved):
Left a couple suggestions, but LGMT

maxisch:
closing in favor of: https://github.com/Crossmint/crossmint-sdk/pull/587

PR #551, opened 2024-07-12
migration: pnpm and node 20

mPaella:
## Description
closes https://linear.app/crossmint/issue/INF-584/sdk-repo-migrate-from-yarn-workspace-to-pnpm-workspace

i've explained this a couple times, but the biggest difference between pnpm and other package managers is the handling of implicit dependencies

[pls read this for a refresher
](https://crossmint.slack.com/archives/C06CANW7UTD/p1712273088723279?thread_ts=1712236800.909339&cid=C06CANW7UTD)

this is why in many of our packages i've had to add additional packages, because we are using them, but they were installed _implicitly_ - now they are explicit

## Test plan

pnpm build + pnpm test working


.github/actions/pnpm-install/action.yml
```diff
@@ -0,0 +1,66 @@
+name: "Setup Node.js, PNPM, and install dependencies"
```
mPaella, line 1: same action as we have in crossmint-main

packages/client/ui/react-ui/src/components/hosted/styles.ts
```diff
@@ -48,4 +49,11 @@ export const useStyles = createUseStyles({
...
     },
-});
+};
+
+export const useStyles: (
+    data?:
+        | CustomStylingProps & {
+              theme?: any;
+          }
+) => Classes<"crossmintButton" | "crossmintImg" | "crossmintParagraph" | "@global"> = createUseStyles(styles);
```
mPaella, line 59: this is a difficult one, but the problem is that previously the return type was inferrered - and then when this was compiled it throws error 
```
The inferred type of 'useStyles' cannot be named without a reference to '.pnpm/jss@10.10.0/node_modules/jss'. This is likely not portable. A type annotation is necessary
```

so, what im adding here is a defined type - its not pretty, but it works....

packages/client/wallets/aa/front/components/CreateWallet.tsx
```diff
@@ -9,7 +9,7 @@ interface CreateWalletComponentProps {
     onWalletCreated: (wallet: EVMAAWallet) => void;
 }
 
-const CreateWalletComponent: React.FC<CreateWalletComponentProps> = ({ onWalletCreated }) => {
+function CreateWalletComponent({ onWalletCreated }: CreateWalletComponentProps) {
```
mPaella, line 12: `React.FC` is not the proper way to define a react component anymore, so when updating some versions these components must be updated as well

devlyn37 (approved):
TY!

PR #557, opened 2024-07-15
chore: separate + organize apps from libs

mPaella:
## Description

separate apps from libs, and adds separate build commands (now if u just wanna build the libs, u dont have to build 4 react/nextjs applications also)

## Test plan

ci


apps/wallets/aa/nextjs/package.json
```diff
@@ -15,7 +15,8 @@
         "ethers": "5.7.2",
         "next": "13.5.6",
         "react": "^18",
-        "react-dom": "^18"
+        "react-dom": "^18",
+        "viem": "2.13.1"
```
mPaella, line 19: pnpm again being awesome, discovering incorrectly used dependencies

we were importing and using viem, but its not set as a dep - now it is

devlyn37 (approved):

PR #562, opened 2024-07-15
changesets: tweak commands

mPaella:
## Description

tweak command so its easier to use changesets

## Test plan

commands work


packages/client/base/package.json
```diff
@@ -31,8 +31,5 @@
         "@ethersproject/transactions": "5.7.0",
         "@solana/web3.js": "1.78.5",
         "@types/uuid": "9.0.4"
-    },
-    "publishConfig": {
```
mPaella, line 35: public is assumed

publishConfig should be used to specify when the package is restricted access, and if u dont want to publish at all it you would use private: true

PR #564, opened 2024-07-16
Add error for projects without non-custodial wallets enabled

bryan-eastwood3:
## Description

This PR contains:
- A resolution for [WAL-2452](https://linear.app/crossmint/issue/WAL-2452/throw-good-error-when-project-is-configured-with-custodial-wallets)
- A proposal for how we should organize errors, namely that:
  - All errors should be associated with a string code
  - This code can be thrown from our server and processed and re-thrown by our SDK (both of these are similar to how Payments works)
  - We should rely on the string error code instead of error subtypes, mostly for the sake of simplicity and DRY

Needs https://github.com/Paella-Labs/crossbit-main/pull/13973 to be landed to throw our new error and resolve the linked issue

## Test plan

TBA

### DIFF
--- a/packages/client/wallets/smart-wallet/src/error/index.ts
+++ b/packages/client/wallets/smart-wallet/src/error/index.ts
@@ -114,3 +114,9 @@ export class AdminAlreadyUsedError extends ConfigError {
         super("This signer was already used to create another wallet. Please use a different signer.");
     }
 }
+
+export class NonCustodialWalletsNotEnabledError extends ConfigError {
+    constructor() {
+        super("Non-custodial wallets are not enabled for this project");
+    }
+}
--- a/packages/client/wallets/smart-wallet/src/api/APIErrorService.ts
+++ b/packages/client/wallets/smart-wallet/src/api/APIErrorService.ts
@@ -5,6 +5,7 @@ import {
     JWTExpiredError,
     JWTIdentifierError,
     JWTInvalidError,
+    NonCustodialWalletsNotEnabledError,
     OutOfCreditsError,
     SmartWalletSDKError,
     UserWalletAlreadyCreatedError,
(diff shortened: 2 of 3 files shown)


packages/client/wallets/aa/src/api/APIErrorService.ts
```diff
@@ -0,0 +1,41 @@
+import { CrossmintServiceError } from "@/types/Error";
+import { AAErrors } from "@/types/errors";
+
+export class APIErrorService {
+    constructor(
+        private errorMessages: Record<AAErrors, string> = {
+            "wallets:aa:noncustodial-wallets-not-enabled": "<Some error message>",
```
bryan-eastwood3, line 7: Feel free to leave thoughts ;)

devlyn37 (commented):
This approach strays pretty far from the SDK spec. 

The main benefits I see for using the error subtype is being able to add properties. Some examples of where this might be useful are:
- Tx revert errors: (error name, etc)
- JWT validation errors: (expiry, issuer, etc)

packages/client/wallets/smart-wallet/src/error/index.ts
```diff
@@ -114,3 +114,9 @@ export class AdminAlreadyUsedError extends ConfigError {
         super("This signer was already used to create another wallet. Please use a different signer.");
     }
 }
+
+export class NonCustodialWalletsNotEnabledError extends ConfigError {
```
devlyn37, line 118: LGTM, let's have Alfonso audit all these before completing milestone 2.

devlyn37, line 118: We should probably have an error code here!

devlyn37 (approved):

PR #570, opened 2024-07-19
Smart Wallets: Isolate Web3Auth Specific Code Into Adapter Package

devlyn37:
## Description

Currently, we give Web3Auth pretty first class treatment, including their own signer type in the core smart wallet SDK. This does not reflect how we see the smart wallet product expanding. Developers should be able to bring their own signers, and pass them into the SDK generically, we **do not want** to write custom integrations for every signer.

In the interest of keeping the core SDK minimal, this PR extracts all web3auth related code into a separate adapter package. When clients pass in a `Web3AuthSigner`, we end up using their SDK to produce a `EIP1193Provider` and using that to sign for ops, messages, etc. Because of that, the adapter was pretty simple to write, since all it needs to do is instantiate the signer, which can easily be passed into the core smart wallet SDK.

Using the package looks like this:

```TS
import {
    Blockchain,
    SmartWalletSDK,
} from "@crossmint/client-sdk-aa-smart-wallet";
import { getWeb3AuthSigner } from "@crossmint/client-sdk-smart-wallet-web3auth-adapter";

const signer: EIP1193Provider = await getWeb3AuthSigner({
    type: "WEB3_AUTH",
    clientId: "X"
    web3AuthNetwork: "Y"
    verifierId: "Z"
    jwt,
    chain,
});

const eip1193WalletConfig = {
    signer
};

const wallet = await smartWalletSDK.getOrCreateWallet(userIdentifier, chain, eip1193WalletConfig);
```

## Test plan

Within the `passkey-release` branch of the `swc-demo`, I created a new web3auth wallet and completed the minting flow successfully.


packages/client/wallets/web3auth-adapter/src/auth.ts
```diff
@@ -11,10 +6,6 @@ export const parseToken = (token: any) => {
             typeof window !== "undefined" ? window.atob(base64) : Buffer.from(base64, "base64").toString();
         return JSON.parse(jsonPayload || "");
     } catch (err) {
-        logError("[PARSE_TOKEN] - ERROR", {
```
alfonso-paella, line 14: why was this dropped

packages/client/wallets/web3auth-adapter/src/networks.ts
```diff
@@ -0,0 +1,133 @@
...
+        ["apex", null],
+    ]).get(chain)!;
+
+    if (url == null) {
+        throw new Error(`Url provider not found for chain ${chain}`);
+    }
+    return url;
+}
+
+export function getBlockExplorerByBlockchain(chain: EVMBlockchainIncludingTestnet) {
```
alfonso-paella, line 36: Why are these part of web3auth? seems like a generic utility

alfonso-paella: actually this applies to most of the file

alfonso-paella (approved):
approved with comments

packages/client/wallets/web3auth-adapter/src/networks.ts
devlyn37: All the code in this file was exclusively used for creating the web3auth signer!

packages/client/wallets/web3auth-adapter/src/auth.ts
devlyn37: Two reasons:
- Doesn't seem that useful.
- I didn't want to copy over our entire logging setup.

packages/client/wallets/web3auth-adapter/src/networks.ts
alfonso-paella: It may be worth moving later to a core package so its usable by other parts of the SDK

PR #572, opened 2024-07-20
feat: api client

mPaella:
## Description

this is a common api client, which _should_ be used for basically all client + server side crossmint api interactions

first one being with embedded checkout v3

### DIFF
--- a/packages/common/base/src/apiClient/ApiClient.ts
+++ b/packages/common/base/src/apiClient/ApiClient.ts
@@ -7,7 +7,7 @@ export abstract class ApiClient {
     abstract get baseUrl(): string;
 
     private async makeRequest(path: string, init: RequestInit) {
-        return fetch(`${ApiClient.normalizePath(this.baseUrl)}/${path}`, {
+        return fetch(`${ApiClient.normalizePath(this.baseUrl)}/${ApiClient.normalizePath(path)}`, {
             ...init,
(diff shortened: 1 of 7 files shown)


packages/common/base/src/apiClient/ApiClient.ts
```diff
@@ -0,0 +1,41 @@
...
+        return this.makeRequest(path, { ...params, method: "PUT" });
+    }
+
+    async delete(path: string, params: Omit<RequestInit, "method">) {
+        return this.makeRequest(path, { ...params, method: "DELETE" });
+    }
+
+    async patch(path: string, params: Omit<RequestInit, "method">) {
+        return this.makeRequest(path, { ...params, method: "PATCH" });
+    }
```
mPaella, line 34: eventually we probably want these functions to be more like the ones in crossmint-main, for ex with retries + json parsng, but this is just the strart

packages/common/base/src/apiClient/CrossmintApiClient.ts
```diff
@@ -0,0 +1,52 @@
...
+        switch (environment) {
+            case "production":
+                url = "https://www.crossmint.com";
+                break;
+            case "staging":
+                url = "https://staging.crossmint.com";
+                break;
+            case "development":
+                url = "http://localhost:3000";
+                break;
```
pgarciaegido, line 45: I would add a comment here making it explicit this is a utility for us crossmint devs

pgarciaegido (commented):
OK

PR #578, opened 2024-07-20
build: dont publish source maps

mPaella:
## Description

https://linear.app/crossmint/issue/INF-625/sdk-repo-dont-publish-maps-to-final-dist

## Test plan
dist doesnt have source maps when built for prod


packages/client/base/package.json
```diff
@@ -20,6 +20,7 @@
     ],
     "scripts": {
         "build": "tsup src/index.ts --clean --format esm,cjs --outDir ./dist --minify --dts --sourcemap",
+        "build:prod": "tsup src/index.ts --clean --format esm,cjs --outDir ./dist --minify --dts",
```
mPaella, line 23: its really ugly to have to repeat the commands like this, but its bc we arent currently able to share configs between projects, which we will eventually do, then the duplication can be removed

devlyn37:
closing stale PR for now

PR #579, opened 2024-07-20
feat: react sdk provider

mPaella:
## Description

top level react provider exposing apiClient

## Test plan

tested with embed v3


pgarciaegido (commented):
OK

PR #588, opened 2024-07-22
embed v3: common service

mPaella:
## Description

common service

## Test plan

local with demo app
iframe has correct params
image

### DIFF
--- a/packages/client/base/src/services/embed/v3/crossmintEmbeddedCheckoutV3Service.ts
+++ b/packages/client/base/src/services/embed/v3/crossmintEmbeddedCheckoutV3Service.ts
@@ -1,0 +1,52 @@
+import { CrossmintEmbeddedCheckoutV3Props } from "@/types/embed/v3/CrossmintEmbeddedCheckoutV3Props";
+import { embeddedCheckoutV3IncomingEvents, embeddedCheckoutV3OutgoingEvents } from "@/types/embed/v3/events";
+
+import { IFrameWindow } from "@crossmint/client-sdk-window";
+import { CrossmintClientSideApiClient } from "@crossmint/common-sdk-base";
+
+export type CrossmintEmbeddedCheckoutV3ServiceProps = { apiClient: CrossmintClientSideApiClient };
+
+export function crossmintEmbeddedCheckoutV3Service({ apiClient }: CrossmintEmbeddedCheckoutV3ServiceProps) {
+    function getIFrameUrl(props: CrossmintEmbeddedCheckoutV3Props) {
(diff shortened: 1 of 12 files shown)


packages/client/base/src/services/embed/v3/crossmintEmbeddedCheckoutV3Service.ts
```diff
@@ -0,0 +1,52 @@
...
+                );
+            } else if (typeof value === "string") {
+                if (value === "undefined") {
+                    continue;
+                }
+                queryParams.append(key, value);
+            } else if (["boolean", "number"].includes(typeof value)) {
+                queryParams.append(key, value.toString());
+            }
+        }
```
mPaella, line 34: nearly exactly the same as https://github.com/Crossmint/crossmint-sdk/blob/main/packages/client/base/src/services/embed/crossmintIFrameService.ts

i think it will need to be changed a bit later tho, and dont want to change existing services, so fine with duplicating for now

packages/client/ui/react-ui/src/components/embed/v3/EmbeddedCheckoutV3IFrame.tsx
```diff
@@ -4,46 +4,41 @@ import { useEffect, useRef, useState } from "react";
...
 
     useEffect(() => {
         if (!iframeClient) {
             return;
         }
-
         iframeClient.on("ui:height.changed", (data) => setHeight(data.height));
+
+        return () => {
+            iframeClient.off("ui:height.changed");
```
pgarciaegido, line 34: Does this mean "unsubscribe" from the event?

packages/client/base/src/services/embed/v3/crossmintEmbeddedCheckoutV3Service.ts
```diff
@@ -0,0 +1,52 @@
...
+
+        return `${urlWithPath}?${queryParams.toString()}`;
+    }
+
+    function createIframeClient(iframe: HTMLIFrameElement) {
+        return IFrameWindow.initExistingIFrame(iframe, {
```
pgarciaegido, line 40: How does this service know that an iframe already exists?

mPaella: bc its passed it, so it must be defined

packages/client/ui/react-ui/src/components/embed/v3/EmbeddedCheckoutV3IFrame.tsx
mPaella: yeah

pgarciaegido (approved):

PR #591, opened 2024-07-23
chore: Renamed log prefix

jmderby:
## Description

Renamed our smart wallet contract prefix from `SDK_AA` to `Smart-Wallet-SDK`


packages/client/wallets/aa/src/utils/constants.ts
```diff
@@ -21,7 +21,7 @@ export const ZD_ARBITRUM_SEPOLIA_PROJECT_ID = "8b8b6d1a-184c-4198-8f29-c07e63aad
 export const ZD_BASE_PROJECT_ID = "5535aa3b-4f9c-45af-9c38-0072369564a3";
 export const ZD_ARBITRUM_NOVA_PROJECT_ID = "b6a6db7c-65de-4f74-9d8f-f70d9c083531";
 export const RELAY_API_KEY = "c8999b7e-8bb7-416d-a854-dac8632f9ee6_crossmint";
-export const SCW_SERVICE = "SDK_AA";
+export const SCW_SERVICE = "Smart-Wallet-SDK";
```
alfonso-paella, line 24: This may be a bit too long, if it appears too many times. 

What about SC_SDK or SCW_SDK?

------

Also i wanted to ensure you've triple checked that this only affects the logs and nothing else

alfonso-paella (changes requested):
.

packages/client/wallets/aa/src/utils/constants.ts
jmderby: @alfonso-paella
 I did a `ctrl + shift + f` in my editor and only found instances relating to logging but i'm not confident. @devlyn37 are you able to speak to this change? 
 
 Also, I renamed it to `SCW_SDK`

devlyn37: Yep only logs

devlyn37 (approved):

PR #592, opened 2024-07-23
feat: Added zod env vars

jmderby:
## Description

Introduces environment variable validation using Zod in `env.ts`. Validates keys like `REACT_APP_CROSSMINT_API_KEY_PROD` ensuring they meet criteria (e.g., non-empty strings). Errors guide setup without halting execution.

## Test plan

1. Validation: Misconfigure .env to trigger error messages.
2. Type Safety: Access `env` properties in TypeScript, expecting IntelliSense to reflect Zod schema types.

### DIFF
--- a/apps/wallets/smart-wallet/react/src/app/env.ts
+++ b/apps/wallets/smart-wallet/react/src/app/env.ts
@@ -3,7 +3,10 @@ import { z } from "zod";
 
 export const env = createEnv({
     server: {
-        /* add server side .env here */
+        CI: z.boolean().optional(),
+        GOOGLE_TEST_EMAIL: z.string().optional(),
+        GOOGLE_TEST_PASSWORD: z.string().optional(),
+        GOOGLE_TEST_ACCOUNT_PRIVATE_KEY: z.string().optional(),
     },
 
     /**
@@ -12,6 +15,7 @@ export const env = createEnv({
      */
     clientPrefix: "REACT_APP_",
     client: {
+        REACT_APP_BASE_URL: z.string().optional(),
         REACT_APP_CROSSMINT_API_KEY_PROD: z.string().optional(),
         REACT_APP_CROSSMINT_API_KEY_STG: z.string().min(1),
         REACT_APP_FIREBASE_API_KEY: z.string().min(1),
--- a/apps/wallets/smart-wallet/react/src/app/env.ts
+++ b/apps/wallets/smart-wallet/react/src/app/env.ts
@@ -1,0 +1,37 @@
+import { z } from "zod";
+
+const envSchema = z.object({
+    REACT_APP_CROSSMINT_API_KEY_PROD: z.string(),
+    REACT_APP_CROSSMINT_API_KEY_STG: z.string().min(1),
+    REACT_APP_FIREBASE_API_KEY: z.string().min(1),
+    REACT_APP_FIREBASE_AUTH_DOMAIN: z.string().min(1),
+    REACT_APP_FIREBASE_PROJECT_ID: z.string().min(1),
+    REACT_APP_FIREBASE_STORAGE_BUCKET: z.string().min(1),
+    REACT_APP_FIREBASE_MESSAGING_SENDER_ID: z.string().min(1),
+    REACT_APP_FIREBASE_APP_ID: z.string().min(1),
+    REACT_APP_WEB3_AUTH_CLIENT_ID_PROD: z.string().min(1),
+    REACT_APP_WEB3_AUTH_CLIENT_ID_STG: z.string().min(1),
+    REACT_APP_WEB3_AUTH_NETWORK_PROD: z.string().min(1),
+    REACT_APP_WEB3_AUTH_NETWORK_STG: z.string().min(1),
+    REACT_APP_WEB3_AUTH_VERIFIER_ID_PROD: z.string().min(1),
+    REACT_APP_WEB3_AUTH_VERIFIER_ID_STG: z.string().min(1),
+});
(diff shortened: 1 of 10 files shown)


apps/wallets/smart-wallet/react/src/app/env.ts
```diff
@@ -0,0 +1,37 @@
+import { z } from "zod";
+
+const envSchema = z.object({
+    REACT_APP_CROSSMINT_API_KEY_PROD: z.string(),
```
alfonso-paella, line 4: I think I would make prod optional, as it will be harder to get such key and i think many folks will just want to test in staging

apps/wallets/smart-wallet/react/src/app/env.ts
```diff
@@ -0,0 +1,37 @@
...
+    REACT_APP_CROSSMINT_API_KEY_STG: z.string().min(1),
```
alfonso-paella, line 5: Or maybe we can specify somehow in zod that we need at least oen of these two, but dont need both?

apps/wallets/smart-wallet/react/src/app/env.ts
```diff
@@ -0,0 +1,37 @@
...
+    REACT_APP_FIREBASE_API_KEY: z.string().min(1),
+    REACT_APP_FIREBASE_AUTH_DOMAIN: z.string().min(1),
+    REACT_APP_FIREBASE_PROJECT_ID: z.string().min(1),
+    REACT_APP_FIREBASE_STORAGE_BUCKET: z.string().min(1),
+    REACT_APP_FIREBASE_MESSAGING_SENDER_ID: z.string().min(1),
+    REACT_APP_FIREBASE_APP_ID: z.string().min(1),
+    REACT_APP_WEB3_AUTH_CLIENT_ID_PROD: z.string().min(1),
+    REACT_APP_WEB3_AUTH_CLIENT_ID_STG: z.string().min(1),
```
alfonso-paella, line 13: same for these and the 2 pairs below

apps/wallets/smart-wallet/react/src/app/env.ts
```diff
@@ -0,0 +1,37 @@
...
+    const errorMessage = `
+    Before running the application, you need to set up your environment variables:
+    1. Copy the '.env.example' file in the root directory and rename the copy to '.env'.
+    2. Open the '.env' file and fill in the necessary values as described in the readme.md.`;
+    console.error(errorMessage);
+    console.error("Zod Errors: ", errors);
+}
+
+// log environment variable warning in console but don't block.
+export const env = envResult.success
```
mPaella, line 32: should use https://env.t3.gg/docs/nextjs rather than hand rolling

apps/wallets/smart-wallet/react/src/app/env.ts
jmderby: One way to achieve this would be something like:

```ts
const clientSchema = z
    .object({
        REACT_APP_CROSSMINT_API_KEY_PROD: z.string().optional(),
        REACT_APP_CROSSMINT_API_KEY_STG: z.string().optional(),
        REACT_APP_FIREBASE_API_KEY: z.string().min(1),
        REACT_APP_FIREBASE_AUTH_DOMAIN: z.string().min(1),
        REACT_APP_FIREBASE_PROJECT_ID: z.string().min(1),
        REACT_APP_FIREBASE_STORAGE_BUCKET: z.string().min(1),
        REACT_APP_FIREBASE_MESSAGING_SENDER_ID: z.string().min(1),
        REACT_APP_FIREBASE_APP_ID: z.string().min(1),
        REACT_APP_WEB3_AUTH_CLIENT_ID_PROD: z.string().optional(),
        REACT_APP_WEB3_AUTH_CLIENT_ID_STG: z.string().optional(),
        REACT_APP_WEB3_AUTH_NETWORK_PROD: z.string().optional(),
        REACT_APP_WEB3_AUTH_NETWORK_STG: z.string().optional(),
        REACT_APP_WEB3_AUTH_VERIFIER_ID_PROD: z.string().optional(),
        REACT_APP_WEB3_AUTH_VERIFIER_ID_STG: z.string().optional(),
    })
    .refine(
        (data) => {
            // Custom validation logic
            const keys = Object.keys(data);
            for (let i = 0; i < keys.length; i += 2) {
                const prodKey = keys[i];
                const stgKey = keys[i + 1];
                // Check if both or neither are defined
                if (
                    ((data as any)[prodKey] && (data as any)[stgKey]) ||
                    (!(data as any)[prodKey] && !(data as any)[stgKey])
                ) {
                    return false;
                }
            }
            return true;
        },
        {
            message: "For each prod/stg pair, exactly one must be defined.",
        }
    );

...
```

but i notice we run into type inference issues.

apps/wallets/smart-wallet/react/src/app/env.ts
jmderby: good call out, I switched it to agnostic core :+1:

apps/wallets/smart-wallet/react/src/app/env.ts
devlyn37: @jmderby I would just do whatever is easiest here for now.

The demo improvement project next cycle will include dropping that chain toggle on the home page, meaning we can have a single set of env vars.

apps/wallets/smart-wallet/react/tsconfig.json
devlyn37, line 1: What's this change?

devlyn37 (approved):

PR #593, opened 2024-07-23
Smart Wallets: Web3Auth Adapter Iteration

devlyn37:
## Description

This PR:
- Removes a redundant parameter from the `getWeb3AuthSigner` function.
- Exports some essential types for using the function from the package.

## Test plan

Locally, within the demo site create a new web3auth wallet.

### DIFF
--- a/packages/client/wallets/web3auth-adapter/src/index.ts
+++ b/packages/client/wallets/web3auth-adapter/src/index.ts
@@ -4,7 +4,7 @@ import { TORUS_NETWORK_TYPE, Web3Auth } from "@web3auth/single-factor-auth";
 import type { EIP1193Provider } from "viem";
 
 import {
-    EVMBlockchainIncludingTestnet,
+    EVMBlockchainIncludingTestnet as Blockchain,
     blockchainToChainId,
     blockchainToDisplayName,
 } from "@crossmint/common-sdk-base";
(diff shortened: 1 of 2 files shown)


packages/client/wallets/web3auth-adapter/src/index.ts
```diff
@@ -17,13 +17,14 @@ import {
...
+export type { TORUS_NETWORK_TYPE } from "@web3auth/single-factor-auth";
+export { EVMBlockchainIncludingTestnet as Blockchain } from "@crossmint/common-sdk-base";
 export type Web3AuthSignerParams = {
-    type: "WEB3_AUTH";
     clientId: string;
     verifierId: string;
     web3AuthNetwork: TORUS_NETWORK_TYPE;
     jwt: string;
-    chain: EVMBlockchainIncludingTestnet;
+    chain: Blockchain;
```
afeight, line 27: nice

packages/client/wallets/web3auth-adapter/src/index.ts
```diff
@@ -17,13 +17,14 @@ import {
     getUrlProviderByBlockchain,
 } from "./networks";
 
+export type { TORUS_NETWORK_TYPE } from "@web3auth/single-factor-auth";
+export { EVMBlockchainIncludingTestnet as Blockchain } from "@crossmint/common-sdk-base";
 export type Web3AuthSignerParams = {
-    type: "WEB3_AUTH";
```
afeight, line 21: wait this isn't redundant, it makes our life easy and IMO better experience for the developer

afeight: maybe I missed something, previously we've used this as a param for `getOrCreateWallet`, like:
```
const signer = {
  type: "WEB3_AUTH",
  ...
}
getOrCreateWallet(user, signer)
```
do we now use `getWeb3AuthSigner` instead?

afeight (approved):
oh, now reading https://github.com/Crossmint/crossmint-sdk/pull/570
yep makes sense

PR #597, opened 2024-07-23
Smart Wallets: Error Observability & Handling (Resolves WAL-2559)

devlyn37:
## Description

The SDK exposes two main pieces:
- `SmartWalletSDK.getOrCreateWallet`
- the `EVMSmartWallet`, along with its `viem` clients

This PR:
- Adds `ErrorProcessor` class and uses them to both places, which makes sure that:
    - A nice error is bubbled up to the developer
    - The original error is recorded.
- Improves the decoration we apply to `smartAccountClients`.

## Test plan

Locally:
- Verified the client decorator properly decorates: `sendTransaction`, `sendUserOperation`, `writeContract`, `signMessage`, `signTypedData`, by confirming through logging and error handling work as expected. ✅ 
- Tested all cases for `ErrorProcessor`. ✅ 
    - Viem errors are bubbled up
    - SmartWalletSDK errors are bubbled up
    - In all other cases the fallback is thrown


packages/client/wallets/smart-wallet/src/blockchain/wallets/clientDecorator.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    private async execute<M extends TxnMethod | SignMethod>(
+        target: SmartAccountClient<EntryPoint>,
+        prop: M,
+        // eslint-disable-next-line @typescript-eslint/ban-types
+        originalMethod: Function,
+        args: any[],
+        crossmintChain: EVMBlockchainIncludingTestnet
+    ) {
```
devlyn37, line 77: I tried taking the generics a little further here by having something like:

```TS
private async execute<M extends TxnMethod | SignMethod>(
    target: SmartAccountClient<EntryPoint>,
    prop: M,
    originalMethod: SmartAccountClient<EntryPoint>[M],
    args: Parameters<SmartAccountClient<EntryPoint>[M]>,
    crossmintChain: EVMBlockchainIncludingTestnet
): Promise<ReturnType<SmartAccountClient<EntryPoint>[M]>>
```

But TS isn't smart enough to figure out narrow types from a tuple yet, so theres no real benefit, and it isn't v/ compatible with the typing `proxy` and `Reflect` use, so requires a lot of casting and is v/ messy for no benefit.

packages/client/wallets/smart-wallet/src/error/index.ts
```diff
@@ -17,24 +14,12 @@ export class SmartWalletSDKError extends Error {
...
-    }
-}
-
 export class TransferError extends SmartWalletSDKError {
     constructor(message: string) {
         super(message, SmartWalletErrors.TRANSFER);
     }
 }
 
-export class TransactionError extends SmartWalletSDKError {
```
devlyn37, line 32: If there's a error during a transaction or signing, we want to throw either the corresponding viem error, or an error related to the signer type (For ex. we'll catch and format webauthn errors w/ better messages for passkeys)

packages/client/wallets/smart-wallet/src/error/index.ts
```diff
@@ -17,24 +14,12 @@ export class SmartWalletSDKError extends Error {
     }
 }
 
-export class NotAuthorizedError extends SmartWalletSDKError {
```
devlyn37, line 20: This wasn't being used, and should be pulled in from the base SDK if possible.

packages/client/wallets/smart-wallet/src/blockchain/wallets/clientDecorator.ts
```diff
@@ -0,0 +1,127 @@
...
+                    );
+            },
+        }) as Client;
+    }
+
+    private async execute<M extends TxnMethod | SignMethod>(
+        target: SmartAccountClient<EntryPoint>,
+        prop: M,
+        // eslint-disable-next-line @typescript-eslint/ban-types
+        originalMethod: Function,
```
devlyn37, line 74: linter doesn't like this, but this is more descriptive than `any`.

Using `SmartAccountClient<EntryPoint>[M]` doesn't play nicely the proxy code.

packages/client/wallets/smart-wallet/src/blockchain/wallets/clientDecorator.ts
```diff
@@ -0,0 +1,127 @@
...
+    constructor(private readonly errorProcessor: ErrorProcessor) {}
+
+    public decorate<Client extends SmartAccountClient<EntryPoint>>({
+        crossmintChain,
+        smartAccountClient,
+    }: {
+        crossmintChain: EVMBlockchainIncludingTestnet;
+        smartAccountClient: Client;
+    }): Client {
+        return new Proxy(smartAccountClient, {
```
bryan-eastwood3, line 50: Pretty cool

packages/client/wallets/smart-wallet/src/SmartWalletSDK.ts
```diff
@@ -48,7 +60,14 @@ export class SmartWalletSDK extends LoggerWrapper {
         return logPerformance(
             "GET_OR_CREATE_WALLET",
             async () => {
-                return await this.smartWalletService.getOrCreate(user, chain, walletConfig);
+                try {
```
bryan-eastwood3, line 63: It might be cool to use a `Proxy` for this pattern, we do this in a bunch of places in the SDK

packages/client/wallets/smart-wallet/src/blockchain/wallets/clientDecorator.ts
bryan-eastwood3: There may be a better type to use but probably not worth messing with TS for a private method

packages/client/wallets/smart-wallet/src/blockchain/wallets/clientDecorator.ts
```diff
@@ -0,0 +1,127 @@
...
+    /*
+     * Chain that ZD uses Gelato as for bundler require special parameters:
+     * https://docs.zerodev.app/sdk/faqs/use-with-gelato#transaction-configuration
+     */
+    private addGelatoBundlerProperties(
+        crossmintChain: EVMBlockchainIncludingTestnet,
+        txnParams: { maxFeePerGas?: bigint; maxPriorityFeePerGas?: bigint }
+    ) {
+        if (usesGelatoBundler(crossmintChain)) {
+            return { ...txnParams, maxFeePerGas: "0x0" as any, maxPriorityFeePerGas: "0x0" as any };
```
bryan-eastwood3, line 122: If this is critical to make the code actually work I'm worried about putting it in a class that seems to mostly be for logging/proxy and isn't easy to find from reading the implementation classes. I'm not super familiar with what all this code does, so lmk if what I'm saying is applicable

devlyn37: Context:
- We expose `viem` wallet and public clients through the `EVMSmartWallet`.
- Under the hood, those clients are from `ZeroDev`.
    - When using certain chains, the ZeroDev client will fail unless these values are here.
    - We want devs to be able to call any of the client methods, and not have to worry about this.

devlyn37: With that context in mind, do you have any reservations. If so, any ideas for alternatives?

packages/client/wallets/smart-wallet/src/SmartWalletSDK.ts
devlyn37: There's 2 instances atm I think, I wanna wait a little longer until there's 3 or 4 before addressing, will be able to make a better abstraction with more cases.

packages/client/wallets/smart-wallet/src/blockchain/wallets/clientDecorator.ts
bryan-eastwood3: In that case it makes sense. If there was a clean way to do it by subclassing the viem wallet that might be preferable, but not a huge deal

bryan-eastwood3 (approved):

PR #397, opened 2024-02-27
Base part for embedded v2

guillermodoghel:
## Description
Embedded v2 base branch. Adds `hideCardForm` and `debug` params. And makes all embedded session to go to
"/sdk/2023-06-09/embeddedCheckout"

On main we control if we forward to v1 or v2 accordingly to a WEBLAB config. Currently forwarding all FIAT sessions to the old version

## Test plan

Screenshot 2024-03-22 at 12 33 36 PM

PR #603, opened 2024-07-25
embed: fix fiat live update params

mPaella:
## Description

i have no idea how this was not found earlier, but this was introduced [here](https://github.com/Crossmint/crossmint-sdk/pull/397/files) - the live params update over window messaging was not tested/implemented in fiat at all

## Test plan

BEFORE:
updates not applying

https://github.com/user-attachments/assets/4a7b5f4f-fa88-4bd5-b115-80ae83dd4a5d

AFTER:

https://github.com/user-attachments/assets/632d4fce-04a2-45fb-be87-957206fd1689


guillermodoghel (approved):
lgtm

PR #606, opened 2024-07-25
AA: Add `AdminAlreadyUsedError`

bryan-eastwood3:
## Description

Handle this error now that it's thrown by the server

## Test plan

Verified that the error is thrown by crossbit-main, have verified that this error checking logic works for any error response that includes the `code` field


packages/client/wallets/smart-wallet/src/error/index.ts
```diff
@@ -38,7 +40,7 @@ export class CrossmintServiceError extends SmartWalletSDKError {
 
 export class NotAuthorizedError extends SmartWalletSDKError {
     constructor(message: string) {
-        super(message, SmartWalletErrors.NOT_AUTHORIZED);
+        super(message, undefined, SmartWalletErrors.NOT_AUTHORIZED);
```
bryan-eastwood3, line 43: This was a bug

packages/client/wallets/smart-wallet/src/error/index.ts
```diff
@@ -79,3 +81,16 @@ export class JWTIdentifierError extends NotAuthorizedError {
...
+export class ConfigError extends SmartWalletSDKError {
+    constructor(message: string) {
+        super(message, undefined, SmartWalletErrors.ERROR_WALLET_CONFIG);
+    }
+}
+
+export class AdminAlreadyUsedError extends ConfigError {
+    public readonly code = SmartWalletErrors.ERROR_ADMIN_SIGNER_ALREADY_USED;
+    constructor() {
+        super("This external signer was already used to create another wallet. Please use a different signer.");
```
devlyn37, line 94: You should drop the word `external` here I think

devlyn37 (approved):

PR #607, opened 2024-07-25
Throw good error when out of credits

bryan-eastwood3:
## Description

Handle the 402 exception we throw when out of credits. Looking for feedback on the error and message since it isn't from the spec

## Test plan

Mocked crossbit-main to throw this error and ensured the SDK handled it properly


devlyn37 (approved):
LGTM

cc @alfongj to review error message.

PR #622, opened 2024-07-31
VC: Use vc only enpoint

vmc-crossmint:
## Description
Use the `credentials nft only` endpoint to speed up credentials retrieval.

## Test plan

Unit tests

### DIFF
--- a/packages/client/verifiable-credentials/src/presentation/getNfts.ts
+++ b/packages/client/verifiable-credentials/src/presentation/getNfts.ts
@@ -3,10 +3,9 @@ import { Nft, isVcChain } from "@/verifiableCredentialsSDK";
 import { crossmintAPI } from "../crossmintAPI";
 import { CrossmintWalletNft } from "../types/nfts";
 
-async function fetchPaginatedData(url: string): Promise<any[]> {
+async function* fetchPaginatedData(url: string): AsyncGenerator<any[]> {
     let page = 1;
     let hasMore = true;
-    let allData: any[] = [];
     const perPage = 50;
     const headers = crossmintAPI.getHeaders();
 
@@ -23,7 +22,7 @@ async function fetchPaginatedData(url: string): Promise<any[]> {
                 );
             }
             const data = await response.json();
-            allData = [...allData, ...data];
+            yield data;
             if (data.length < perPage) {
                 hasMore = false;
             } else {
@@ -35,8 +34,6 @@ async function fetchPaginatedData(url: string): Promise<any[]> {
             throw new Error(`Failed to fetch data: ${error.message}`);
         }
     }
-
-    return allData;
 }
 
 export async function getWalletNfts(chain: string, wallet: string): Promise<CrossmintWalletNft[]> {
@@ -44,14 +41,24 @@ export async function getWalletNfts(chain: string, wallet: string): Promise<Cros
     const headers = crossmintAPI.getHeaders();
     const url = `${baseUrl}/api/v1-alpha1/wallets/${chain}:${wallet}/nfts`;
 
-    return (await fetchPaginatedData(url)) as CrossmintWalletNft[];
+    const allData: CrossmintWalletNft[] = [];
+    for await (const data of fetchPaginatedData(url)) {
+        allData.push(...data);
+    }
+
+    return allData;
 }
 
 export async function getWalletVCNfts(chain: string, wallet: string): Promise<CrossmintWalletNft[]> {
     const baseUrl = crossmintAPI.getBaseUrl();
     const url = `${baseUrl}/api/v1-alpha1/wallets/${chain}:${wallet}/credential_nfts`;
 
-    return (await fetchPaginatedData(url)) as CrossmintWalletNft[];
+    const allData: CrossmintWalletNft[] = [];
+    for await (const data of fetchPaginatedData(url)) {
+        allData.push(...data);
+    }
+
+    return allData;
 }
 
 export function filterVCCompErc721(nfts: CrossmintWalletNft[]): Nft[] {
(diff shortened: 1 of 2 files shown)


packages/client/verifiable-credentials/src/presentation/getNfts.ts
```diff
@@ -41,6 +41,29 @@ export async function getWalletNfts(chain: string, wallet: string) {
     return allData;
 }
 
+export async function getWalletVCNfts(chain: string, wallet: string): Promise<CrossmintWalletNft[]> {
+    const baseUrl = crossmintAPI.getBaseUrl();
+    const headers = crossmintAPI.getHeaders();
+
+    const url = `${baseUrl}/api/v1-alpha1/wallets/${chain}:${wallet}/nfts/credentials`;
```
alberto-crossmint, line 48: Depends on the discussion on https://github.com/Paella-Labs/crossbit-main/pull/14192

packages/client/verifiable-credentials/src/presentation/getNfts.ts
```diff
@@ -3,44 +3,57 @@ import { Nft, isVcChain } from "@/verifiableCredentialsSDK";
...
         }
     }
 
     return allData;
 }
 
+export async function getWalletNfts(chain: string, wallet: string): Promise<CrossmintWalletNft[]> {
+    const baseUrl = crossmintAPI.getBaseUrl();
+    const headers = crossmintAPI.getHeaders();
+    const url = `${baseUrl}/api/v1-alpha1/wallets/${chain}:${wallet}/nfts`;
```
alberto-crossmint, line 45: Can we have something similar to routes.ts in the main repo to centralize all these routes? So we don't need to maintain them spread out in the code

packages/client/verifiable-credentials/src/presentation/getNfts.ts
```diff
@@ -3,44 +3,57 @@ import { Nft, isVcChain } from "@/verifiableCredentialsSDK";
...
+
+    return (await fetchPaginatedData(url)) as CrossmintWalletNft[];
+}
+
+export async function getWalletVCNfts(chain: string, wallet: string): Promise<CrossmintWalletNft[]> {
+    const baseUrl = crossmintAPI.getBaseUrl();
+    const url = `${baseUrl}/api/v1-alpha1/wallets/${chain}:${wallet}/credential_nfts`;
```
alberto-crossmint, line 52: Same

packages/client/verifiable-credentials/src/presentation/getNfts.ts
```diff
@@ -56,7 +69,7 @@ export function filterVCCompErc721(nfts: CrossmintWalletNft[]): Nft[] {
 }
 
 export async function getWalletVcCompatibleNfts(chain: string, wallet: string): Promise<Nft[]> {
-    const nfts = await getWalletNfts(chain, wallet);
+    const nfts = await getWalletVCNfts(chain, wallet);
     if (nfts == null) {
         throw new Error("Failed to get nfts");
     }
```
alberto-crossmint, line 75: Do we need the filterVCCompErc721 anymore? Isn't this ensured by the credentials-specific endpoint now?

packages/client/verifiable-credentials/src/presentation/getNfts.ts
```diff
@@ -3,44 +3,57 @@ import { Nft, isVcChain } from "@/verifiableCredentialsSDK";
 import { crossmintAPI } from "../crossmintAPI";
 import { CrossmintWalletNft } from "../types/nfts";
 
-export async function getWalletNfts(chain: string, wallet: string) {
+async function fetchPaginatedData(url: string): Promise<any[]> {
```
alberto-crossmint, line 6: I think this method can become useless if the response size is huge. One option would be paginating it and transfer this pagination param to the backend but this seems like the perfect case for using an AsyncGenerator https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncGenerator

Wdyt?

alberto-crossmint: I know if the response is huge our backend would crash even before but thinking on the long term if we solve the pagination problem with the NFTs fetch :D

packages/client/verifiable-credentials/src/presentation/getNfts.ts
vmc-crossmint: Yes but that method also projects them to the correct type, the type returned by the wallet api is different

packages/client/verifiable-credentials/src/presentation/getNfts.ts
```diff
@@ -3,40 +3,59 @@ import { Nft, isVcChain } from "@/verifiableCredentialsSDK";
 import { crossmintAPI } from "../crossmintAPI";
 import { CrossmintWalletNft } from "../types/nfts";
 
-export async function getWalletNfts(chain: string, wallet: string) {
+async function* fetchPaginatedData(url: string): AsyncGenerator<any[]> {
```
alberto-crossmint, line 6: Oh sorry I should have been more clear! But everything you did here is directly usable. Only one follow up:

- instead of yielding the whole page let's yield one by one

```typescript
              for (const item of data) {
                    yield item;
                }

                if (data.length < perPage) {
                    hasMore = false;
                } else {
                    console.debug(`Processed ${data.length} items from page ${page}`);
                    page++;
                }
```

The user facing functions can return the generator. This way:
- We provide the client a generator that gets the best latency possible in the worst case scenario
- They can retrieve only n items and it won't get stuck if the user has like 1 million credentials
- we abstract away the pagination in a single function and the others (getWalletNfts and getWalletsVCNfts) are completely agnostic

Everything else lgtm !

alberto-crossmint (approved):

PR #624, opened 2024-07-31
Window: Add Cross origin param

MatiNavCrossmint:
## Description

Window: Add Cross-origin param

## Test plan

Auth flow tested locally

### DIFF
--- a/packages/client/window/src/utils/popupStrategy.ts
+++ b/packages/client/window/src/utils/popupStrategy.ts
@@ -6,20 +6,8 @@ interface CreatePopupService {
 export class CreatePopupStrategy {
     protected createPopupService: CreatePopupService;
 
-    constructor() {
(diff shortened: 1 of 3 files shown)


packages/client/window/src/utils/popupStrategy.ts
```diff
@@ -6,20 +6,8 @@ interface CreatePopupService {
...
-
-    private isCrossmintOrigin(): boolean {
-        try {
-            const url = new URL(window.location.origin);
-            return url.hostname.endsWith("crossmint.com");
-        } catch (e) {
-            console.error("Invalid URL", e);
-            return false;
-        }
+    constructor(crossOrigin: boolean) {
```
maxisch, line 9: maybe add `crossOrigin: boolean = false` as it previously defaulted to false in the catch block

maxisch (approved):

PR #627, opened 2024-08-01
Smart Wallets: Update Chain Export

devlyn37:
## Description

This PR:
- removes the `SmartWalletChain` export from `@crossmint/client-sdk-smart-wallet`, and replaces it with the chains from the core sdk package, renamed from `EVMChainsIncludingTestnet` to `Chain`.
- renames the chain export from `@crossmint/client-sdk-smart-wallet-web3auth-adapter`, which is already doing what I mention above, from `Blockchain` to `Chain`.

These are breaking changes, but I only want to do a patch version bump, the SDK shipped less than 24 hrs ago and is in EAP mode so I'm comfortable breaking semantic versioning rules.

## Test plan

Updated the Demo to use the new export and TS compiled.

### DIFF
--- a/apps/wallets/smart-wallet/react/src/app/utils/createAAWallet/createMetamaskWallet.ts
+++ b/apps/wallets/smart-wallet/react/src/app/utils/createAAWallet/createMetamaskWallet.ts
@@ -1,7 +1,6 @@
 import { EIP1193Provider } from "viem";
 
-import { SmartWalletSDK } from "@crossmint/client-sdk-smart-wallet";
-import { SmartWalletChain } from "@crossmint/client-sdk-smart-wallet";
+import { Chain, SmartWalletSDK } from "@crossmint/client-sdk-smart-wallet";
 
 type WindowWithEthereum = Window & {
     ethereum: EIP1193Provider;
@@ -24,13 +23,9 @@ export const createMetamaskAAWallet = async (isProd: boolean) => {
                   clientApiKey: process.env.REACT_APP_CROSSMINT_API_KEY_STG || "",
               });
 
-        return await xm.getOrCreateWallet(
-            userIdentifier,
-            isProd ? SmartWalletChain.POLYGON : SmartWalletChain.POLYGON_AMOY,
-            {
-                signer: window.ethereum as any,
-            }
-        );
+        return await xm.getOrCreateWallet(userIdentifier, isProd ? Chain.POLYGON : Chain.POLYGON_AMOY, {
+            signer: window.ethereum as any,
+        });
     } catch (error) {
         console.error("Error creating MetaMask wallet:", error);
         throw error;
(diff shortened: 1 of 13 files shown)


packages/client/wallets/smart-wallet/src/index.ts
```diff
@@ -1,4 +1,4 @@
-export { blockchainToChainId } from "@crossmint/common-sdk-base";
+export { blockchainToChainId, EVMBlockchainIncludingTestnet as Chain } from "@crossmint/common-sdk-base";
```
devlyn37, line 1: I think `Chain` is better than `Blockchain` here. fwiw it's how we've named things in `crossbit-main` to

apps/wallets/smart-wallet/react/src/app/utils/createAAWallet/createMetamaskWallet.ts
```diff
@@ -24,13 +23,9 @@ export const createMetamaskAAWallet = async (isProd: boolean) => {
...
               });
 
-        return await xm.getOrCreateWallet(
-            userIdentifier,
-            isProd ? SmartWalletChain.POLYGON : SmartWalletChain.POLYGON_AMOY,
-            {
-                signer: window.ethereum as any,
-            }
-        );
+        return await xm.getOrCreateWallet(userIdentifier, isProd ? Chain.POLYGON : Chain.POLYGON_AMOY, {
```
afeight, line 26: nit: `xm` => `smartWalletSDK` or like `smartWallets`

apps/wallets/smart-wallet/react/src/app/utils/createAAWallet/createMetamaskWallet.ts
```diff
@@ -24,13 +23,9 @@ export const createMetamaskAAWallet = async (isProd: boolean) => {
...
+            signer: window.ethereum as any,
+        });
```
afeight, line 28: nit: might be a bit too much going on in one line
```suggestion
        const blockchain = isProd ? Chain.POLYGON : Chain.POLYGON_AMOY
        return await xm.getOrCreateWallet(userIdentifier, blockchain, { signer: window.ethereum as any });
```

afeight (approved):
tentative approval to unblock, but do feel that `Blockchain` is better for our users because: 
1. they're not typically as blockchain-savvy as us
2. code editors will find `blockchain` if you type `chain` or `blockchain`, but not find `chain` if you try to type `blockchain`

apps/wallets/smart-wallet/react/src/app/utils/createAAWallet/createMetamaskWallet.ts
devlyn37: Yeah demo is a huge mess atm, jon and I will be cleaning up in separate PRs

apps/wallets/smart-wallet/react/src/app/utils/createAAWallet/createMetamaskWallet.ts
devlyn37: agree, but out of scope here. will cleanup demo later w/ jon

PR #632, opened 2024-08-01
Smart Wallets: Passkey Prompt Error (Resolves WAL-2464)

devlyn37:
## Description

This PR:
- lays the foundation for catching passkey specific errors
- adds an error for when a passkey prompt is cancelled or times out.

## Test plan

Locally, using the demo site, tested the error both when registering a passkey for a new wallet and signing a user operation.

### DIFF
--- a/crossmint-sdk.code-workspace
+++ b/crossmint-sdk.code-workspace
@@ -29,7 +29,7 @@
             "path": "packages/client/ui/react-ui"
         },
         {
-            "name": "⚛️ @crossmint/client-sdk-smart-wallet-react-starter",
+            "name": "⚛️🔒 @crossmint/client-sdk-smart-wallet-react-starter",
             "path": "apps/wallets/smart-wallet/react"
         },
         {
--- a/packages/client/wallets/smart-wallet/src/blockchain/wallets/passkey.ts
+++ b/packages/client/wallets/smart-wallet/src/blockchain/wallets/passkey.ts
@@ -3,11 +3,11 @@ import { type PasskeySignerData, displayPasskey } from "@/types/API";
 import type { PasskeySigner, UserParams, WalletParams } from "@/types/Config";
 import type { AccountAndSigner, PasskeyValidatorSerializedData, WalletCreationParams } from "@/types/internal";
 import { PasskeyValidatorContractVersion, WebAuthnMode, toPasskeyValidator } from "@zerodev/passkey-validator";
-import { type KernelValidator, createKernelAccount } from "@zerodev/sdk";
+import { KernelSmartAccount, type KernelValidator, createKernelAccount } from "@zerodev/sdk";
 import { WebAuthnKey, toWebAuthnKey } from "@zerodev/webauthn-key";
 import type { EntryPoint } from "permissionless/types/entrypoint";
 
-import { PasskeyMismatchError } from "../../error";
+import { PasskeyMismatchError, PasskeyPromptError } from "../../error";
 
 export interface PasskeyWalletParams extends WalletCreationParams {
     walletParams: WalletParams & { signer: PasskeySigner };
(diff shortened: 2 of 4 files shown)


packages/client/wallets/smart-wallet/src/blockchain/wallets/passkey.ts
```diff
@@ -102,6 +106,32 @@ export class PasskeyAccountService {
...
         };
     }
+
+    private mapError(error: any, passkeyName: string) {
+        if (error.code === "ERROR_PASSTHROUGH_SEE_CAUSE_PROPERTY" && error.name === "NotAllowedError") {
+            return new PasskeyPromptError(passkeyName);
+        }
+
+        return error;
+    }
```
devlyn37, line 116: Initially I tried also adding `error instanceof WebAuthnError` here first, which I imported in from the `@simpleWebAuthn` (what zerodev is using under the hood), but didn't work.

packages/client/wallets/smart-wallet/src/error/index.ts
```diff
@@ -119,6 +120,18 @@ export class UserWalletAlreadyCreatedError extends SmartWalletSDKError {
     }
 }
 
+export class PasskeyPromptError extends SmartWalletSDKError {
+    public passkeyName: string;
+
+    constructor(passkeyName: string) {
+        super(
+            `Prompt was either cancelled or timed out for passkey ${passkeyName}`,
+            SmartWalletErrors.ERROR_PASSKEY_PROMPT
```
devlyn37, line 129: There should be an `undefined` here that's inbetween.

I fixed in this PR: https://github.com/Crossmint/crossmint-sdk/pull/633

packages/client/wallets/smart-wallet/src/blockchain/wallets/passkey.ts
```diff
@@ -102,6 +106,32 @@ export class PasskeyAccountService {
...
+                return async (...args: Parameters<Client["signUserOperation"]>) => {
+                    try {
+                        return await Reflect.get(target, prop, receiver).call(target, ...args);
+                    } catch (error) {
+                        throw this.mapError(error, passkeyName);
+                    }
+                };
+            },
+        });
+    }
```
afeight, line 134: making sure I understand correctly. Here we're wrapping the viem client that we'd normally pass back such that when they call `signUserOperation`, we catch errors and map them to a friendlier error via `mapError`?
Nice, just the types scare me lol

packages/client/wallets/smart-wallet/src/blockchain/wallets/passkey.ts
```diff
@@ -102,6 +106,32 @@ export class PasskeyAccountService {
             Authorization: `Bearer ${user.jwt}`,
         };
     }
+
+    private mapError(error: any, passkeyName: string) {
+        if (error.code === "ERROR_PASSTHROUGH_SEE_CAUSE_PROPERTY" && error.name === "NotAllowedError") {
```
afeight, line 111: this isn't a var in the zerodev SDK anywhere we'd be able to point to is it?
would be nicer if there were a `import { PASSKEY_ERROR_CODE } from "zerodev"` that we could use instead of what's now a bit of a magic string

packages/client/wallets/smart-wallet/src/blockchain/wallets/passkey.ts
devlyn37: That generic is named poorly, we're actually wrapping the viem/permissionless `account`, any of the txn method we expose from the our viem client on `EVMSmartWallet`, `sendTransaction`, `sendUserOperation`, or `writeContract`, call this `signUserOperation` method under the hood, so by mapping errors here we get this error handling everywhere.

devlyn37: Writing this out though I need to make sure that message signing is covered too

packages/client/wallets/smart-wallet/src/blockchain/wallets/passkey.ts
devlyn37: It's a `simpleWebAuthn/browser` error, but they don't export the error codes.

speaking of which, neither do we, I'll add that in another PR.

packages/client/wallets/smart-wallet/src/blockchain/wallets/passkey.ts
devlyn37: Covered all signing method, and improved types where it matters!

crossmint-sdk.code-workspace
```diff
@@ -29,7 +29,7 @@
             "path": "packages/client/ui/react-ui"
         },
         {
-            "name": "⚛️ @crossmint/client-sdk-smart-wallet-react-starter",
+            "name": "⚛️🔒 @crossmint/client-sdk-smart-wallet-react-starter",
```
devlyn37, line 32: Makes it a little easier to pick out that this is smart wallet related.

PR #633, opened 2024-08-01
Smart Wallets: Passkey Registration Error (Resolves WAL-2557)

devlyn37:
## Description

We receive a pretty vague error through the SDK when our passkey server fails to verify a registration, either because it took too long, or the signature was invalid.

This PR adds error handling for this case!

## Test plan

Locally reproduced the error, confirming the new improved error is thrown now.

### DIFF
--- a/packages/client/wallets/smart-wallet/src/blockchain/wallets/passkey.ts
+++ b/packages/client/wallets/smart-wallet/src/blockchain/wallets/passkey.ts
@@ -3,11 +3,11 @@ import { type PasskeySignerData, displayPasskey } from "@/types/API";
 import type { PasskeySigner, UserParams, WalletParams } from "@/types/Config";
 import type { AccountAndSigner, PasskeyValidatorSerializedData, WalletCreationParams } from "@/types/internal";
 import { PasskeyValidatorContractVersion, WebAuthnMode, toPasskeyValidator } from "@zerodev/passkey-validator";
-import { type KernelValidator, createKernelAccount } from "@zerodev/sdk";
+import { KernelSmartAccount, type KernelValidator, createKernelAccount } from "@zerodev/sdk";
 import { WebAuthnKey, toWebAuthnKey } from "@zerodev/webauthn-key";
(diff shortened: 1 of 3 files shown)


packages/client/wallets/smart-wallet/src/blockchain/wallets/passkey.ts
```diff
@@ -108,6 +108,10 @@ export class PasskeyAccountService {
     }
 
     private mapError(error: any, passkeyName: string) {
+        if (error.message === "Registration not verified") {
```
devlyn37, line 111: This is pretty brittle XD, but there's literally no other properties the error has, so doing the best we can here.

afeight: same Q here, there's no const from zerodev/viem that we can point to?
Otherwise if this is the best we can do, then is what it is

packages/client/wallets/smart-wallet/src/error/index.ts
```diff
@@ -126,12 +127,26 @@ export class PasskeyPromptError extends SmartWalletSDKError {
     constructor(passkeyName: string) {
         super(
             `Prompt was either cancelled or timed out for passkey ${passkeyName}`,
+            undefined,
```
afeight, line 130: seeing that we missed this in the previous PR, should this constructor take an object param instead of 3 separates?

afeight (approved):
minor, otherwise G2G

packages/client/wallets/smart-wallet/src/blockchain/wallets/passkey.ts
devlyn37: That's the case unfortunately

Code in the ZD SDK
```TS
if (!registerVerifyResult.verified) {
    throw new Error("Registration not verified")
}
```

packages/client/wallets/smart-wallet/src/error/index.ts
devlyn37: Good idea, I'll do it on top of these PRs

PR #634, opened 2024-08-01
Smart Wallets: Passkey Incompatible Authenticator Error (Resolves WAL-2587)

devlyn37:
## Description

Mystery solved, full investigation here: https://linear.app/crossmint/issue/WAL-2587/mystery-webauthn-error

This PR handles an error that only occurs when a newly registered passkey's public key has a format incompatible with the P256 signature verification that happens onchain. We've only been able to reproduce this with older yubikeys.

## Test plan

Reproduced the error, and made sure the improved error is now thrown.


bryan-eastwood3 (approved):
Nice!

PR #639, opened 2024-08-05
feat: Crossmint class + CrossmintApiClient changes

mPaella:
## Description

changes according to discussion with alfon

## Test plan

not a ton to test

example surface, of a parent class extending the api client
```
export class CrossmintSmartWalletApiClient extends CrossmintApiClient {
    constructor(crossmint: Crossmint) {
        super(crossmint, {
            internalConfig: {
                sdkMetadata: {
                    name: "crossmint-smart-wallet",
                    version: "1.0.0",
                },
                apiKeyExpectations: {
                    usageOrigin: "client", // u dont need to specify this, but its how u would gate for client/server keys
                },
            },
        });
    }


    getPassKeys() {
          return this.get("/2024-06-09/passkeys")
     }
}

const xm = new Crossmint({
    apiKey: "123",
});

const smartWalletApiClient = new CrossmintSmartWalletApiClient(xm);
const passKeys = await smartWalletApiClient.getPassKeys()
```


packages/common/base/src/apiClient/CrossmintApiClient.ts
```diff
@@ -1,36 +1,45 @@
...
-};
-
-export type CrossmintApiClientCtorWithoutExpectationsParams = {
-    apiKey: string;
-    overrideBaseUrl?: string;
-    extraAuthHeaders?: HeadersInit;
+export type CrossmintApiClientInternalConfig = {
+    sdkMetadata: {
+        name: string;
+        version: string;
```
alfonso-paella, line 11: Is there any way this can be retrieved auotmatically from the package somehow, to ensure it's always in sync?

packages/common/base/src/apiClient/CrossmintApiClient.ts
```diff
@@ -1,36 +1,45 @@
...
+    };
+    apiKeyExpectations?: ValidateAPIKeyPrefixExpectations;
```
alfonso-paella, line 13: hmmm.. not sure id do this but ok to keep for now

packages/common/base/src/Crossmint/Crossmint.ts
```diff
@@ -0,0 +1,13 @@
+export type CrossmintConfig = {
+    apiKey: string;
+    jwt?: string;
+    overrideBaseUrl?: string;
+};
+
+export class Crossmint {
```
alfonso-paella, line 7: Not sure if you intended to do this on another PR, but I would add a public get projectId() that extracts it from the key, as a convenience funciton for debugging / logging if we ask a dev to tell us their project ID. 

Not super critical but seems like 3 LOC to add

packages/common/base/src/Crossmint/Crossmint.ts
```diff
@@ -0,0 +1,13 @@
...
+    constructor(public config: CrossmintConfig) {}
+
+    setJwt(jwt: string | undefined) {
```
alfonso-paella, line 10: should we use a class setter

alfonso-paella, line 10: Can you add a typedoc? I think we're going to start requiring these for SDK reference autogeneration. PRobably cursor can write it for you

packages/common/base/src/apiClient/CrossmintApiClient.ts
```diff
@@ -1,36 +1,45 @@
...
-        const apiKeyValidationResult = validateAPIKey(apiKey, expectations);
+    constructor(
+        public crossmint: Crossmint,
+        {
+            internalConfig,
+        }: {
+            internalConfig: CrossmintApiClientInternalConfig;
+        }
+    ) {
+        const apiKeyValidationResult = validateAPIKey(crossmint.config.apiKey, internalConfig.apiKeyExpectations);
```
alfonso-paella, line 28: I feel like we should prob do some form of validation when they call new Crossmint({ apiKey }) if its malformed

alfonso-paella (approved):

packages/common/base/src/Crossmint/Crossmint.ts
mPaella: the docs arent generated from the types?

packages/common/base/src/apiClient/CrossmintApiClient.ts
mPaella: inside of the individual sdks, we basically can, but not here

PR #642, opened 2024-08-06
JWKS: Working privy integration

guiwoda:
## Description
PoC that shows we can leverage Privy auth with smart wallets.

## Test plan
This uses a client API key configured with the following:
```
"jwtConfig": {
    "jwksUri": "https://auth.privy.io/api/v1/apps/clz8iksv40d95odgp6jcai4vp/jwks.json",
    "iss": "privy.io",
    "userIdClaim": "sub"
  }
```

You can change the app id in the code (or better yet, change the code to make it an env var or something) and also in the client api key used, so that you use your own Privy app.

### DIFF
--- a/apps/wallets/smart-wallet/react/src/app/App.tsx
+++ b/apps/wallets/smart-wallet/react/src/app/App.tsx
@@ -1,3 +1,4 @@
+import { PrivyProvider } from "@privy-io/react-auth";
 import { Toaster } from "react-hot-toast";
 
 import { AppProvider } from "./AppContext";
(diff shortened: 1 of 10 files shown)


apps/wallets/smart-wallet/react/src/app/App.tsx
```diff
@@ -1,16 +1,33 @@
...
+            config={{
+                // Customize Privy's appearance in your app
+                appearance: {
+                    theme: "light",
+                    accentColor: "#676FFF",
+                    logo: "https://cdn.prod.website-files.com/653a93effa45d5e5a3b8e1e8/653b06fbe475503198236e11_LOGO.svg",
+                },
+                // Create embedded wallets for users who don't have a wallet
+                embeddedWallets: {
+                    createOnLogin: "users-without-wallets",
```
guiwoda, line 20: This is probably wrong, but I wasn't looking at what Privy did for me when logging in. I wanted to make sure our backend validates Privy's JWTs correctly.

PR #646, opened 2024-08-08
feat: Crossmint type + useCrossmint

mPaella:
## Description

as discussed:
- changing `Crossmint` from a class to a type
- creating `useCrossmint` and `CrossmintProvider`
- i added a little utility function `createCrossmint`, which just validates the api key as well

## Test plan

not much to test, just a simple type and hook


packages/common/base/src/types/Crossmint.ts
```diff
@@ -0,0 +1,20 @@
+import { validateAPIKey } from "@/apiKey";
+
+export type Crossmint = {
+    apiKey: string;
+    jwt?: string;
+    overrideBaseUrl?: string;
+};
+
+export function createCrossmint(config: Crossmint): Crossmint {
```
mPaella, line 9: this is mainly internal, just to make sure the api key is valid from the start

packages/common/base/src/types/Crossmint.ts
```diff
@@ -0,0 +1,20 @@
...
+    const { apiKey, jwt, overrideBaseUrl } = config;
+    const apiKeyValidationResult = validateAPIKey(apiKey);
```
MatiNavCrossmint, line 11: I'm curious about the security concerns of validating the API Key client side. I know that we are not exposing any secret, but I wonder if we are making it easier to malicious attackers.

I asked chatGPT, which said:

```
Exposing the validateAPIKey function, which includes decoding and validating API keys, in a package that can be imported by external developers can be considered a bad practice due to several security concerns. Here are the main reasons why:

Security Concerns
Reverse Engineering:
By providing the decoding and validation logic, you make it easier for someone to reverse engineer how your API keys are structured and validated. This can lead to unauthorized access if someone figures out how to generate valid API keys.

Signature Verification:
If someone understands how the signature verification works, they might attempt to create fake API keys that pass the validation. This can compromise the security of your API.

Public Key Exposure:
Exposing the public keys used for signature verification can be a security risk if not managed properly. Even though public keys are meant to be public, knowing the exact keys used for validation can aid in reverse engineering efforts.
```

and the first suggestion was to make it server-side

mPaella: thats just not really true, they would need our private key to make a fake api key

thats like saying are blockchain explorers insecure bc they show the signatures of a transaction...

mPaella: if theres truly a malicious attacker, they would easily be able to figure it out anyways

packages/client/ui/react-ui/src/hooks/useCrossmint.tsx
```diff
@@ -0,0 +1,26 @@
...
+    setCrossmint: Dispatch<SetStateAction<Crossmint>>;
+}
+
+const CrossmintContext = createContext<CrossmintContext | null>(null);
+
+export function CrossmintProvider({
+    children,
+    ...createCrossmintParams
+}: { children: React.ReactNode } & Parameters<typeof createCrossmint>[0]) {
+    const [crossmint, setCrossmint] = useState<Crossmint>(createCrossmint(createCrossmintParams));
```
alfonso-paella, line 16: Discussing this on DMs -- this may cause issues if some consumer of this hook does something like...

```
const { crossmint } = useCrossmint();
const [walletSDK, setWalletSDK] = useState()<WalletsSDK>;

useEffect(() => {
  const _walletSDK = WalletSDK.from(crossmint); // this crossmint ref will be out of date later
  setWalletSDK(_walletSDK); 
}, [])

```

I think some people may choose to do this if TheirSDK.from() is a heavy initialization function. 

An additional issue is that typescript SDKs that mutate jwt, now would need to know react or take generic callback functions for updating the JWT, which seems like a lot of work

There's an alternative where we return a reference here, and observe it for updates, and emit a state change (on e.g. a 'version' property) so we maintain reactivity

devlyn37:
Closing as I think another PR with similar changes has landed

PR #648, opened 2024-08-08
docs: typedoc automated SDK documentation

dmulvi:
## Description

wip: this is the start of automating the documentation of our SDKs

## Test plan

testing locally


packages/client/wallets/smart-wallet/custom-plugin.mjs
devlyn37, line 1: I'd guess most of this isn't smart wallet specific and will eventually go to some shared spot like the `@crossmint/sdk-common` right?

devlyn37 (approved):
LGTM, left a few comments and questions but non-blocking. Approving assuming you'll fix whatever is breaking the build here!

packages/client/wallets/smart-wallet/tsconfig.typedoc.json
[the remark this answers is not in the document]
dmulvi: good call!

packages/client/wallets/smart-wallet/custom-plugin.mjs
dmulvi: Correct, most of it is not smart wallet specific. For now keeping as-is until I can make the time to look into pushing some of this code up a few levels in the SDK. That will require some more testing, at which point I'll fix here as well.

PR #653, opened 2024-08-10
Smart Wallets: More Targeted Logging (Resolves WAL-2453)

devlyn37:
## Description

A few SDK versions ago we we're having a really hard time figuring out what was happening during issues from clients, and we ended up taking the nuclear approach for logs, logging the input and output of basically every single function.

This PR removes `LoggerWrapper`, the class responsible for that behavior, and drops a couple other small pieces. There's a few reasons to do this:
- Our current logs are way too noisy
- When on localhost, we log JWTs in the browser, which can make clients think we do so in other envs.

The key pieces that need strong logging, like when sending transactions, signing messages, etc are all covered by the `SmartAccountClient` decoration we do now.

## Test plan

Went through wallet fetching/creation and minting flow on the demo site and confirmed logs work as expected


packages/client/wallets/smart-wallet/src/utils/log.ts
```diff
@@ -1,115 +1,15 @@
...
-    }
-}
 
 export async function logPerformance<T>(name: string, cb: () => Promise<T>, extraInfo?: object) {
     const start = new Date().getTime();
     const result = await cb();
     const durationInMs = new Date().getTime() - start;
     const args = { durationInMs, ...extraInfo };
-    logInfoIfNotInLocalhost(`[${SCW_SERVICE} - ${name} - TIME] - ${beautify(args)}`, { args });
+    logInfo(`[${SCW_SERVICE} - ${name} - TIME] - ${beautify(args)}`, { args });
```
devlyn37, line 9: Not needed

packages/client/wallets/smart-wallet/src/utils/log.ts
```diff
@@ -1,115 +1,15 @@
...
-        return logPerformance(name, cb, this.extraInfo);
-    }
-}
 
 export async function logPerformance<T>(name: string, cb: () => Promise<T>, extraInfo?: object) {
     const start = new Date().getTime();
     const result = await cb();
     const durationInMs = new Date().getTime() - start;
     const args = { durationInMs, ...extraInfo };
-    logInfoIfNotInLocalhost(`[${SCW_SERVICE} - ${name} - TIME] - ${beautify(args)}`, { args });
```
devlyn37, line 81: not needed

jmderby: nice

jmderby:
changes lgtm, logs still work in browser but are reduced :+1:

jmderby (approved):

PR #654, opened 2024-08-11
AA: Move error codes from AA package to base package

bryan-eastwood3:
## Description

This follows the convention of our other errors, namely those from payments. Resolves WAL-2588.

## Test plan

It builds


packages/client/base/src/types/errors.ts
```diff
@@ -29,7 +29,31 @@ export const PaymentErrors = {
 } as const;
 export type PaymentErrors = (typeof PaymentErrors)[keyof typeof PaymentErrors];
 
+export const SmartWalletErrors = {
```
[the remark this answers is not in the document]
devlyn37: Maybe you could export it as `SmartWalletErrorCode` from the smart wallet package. That way you don't break the naming convention here, but the exports from the package are more clear

packages/client/base/src/types/errors.ts
```diff
@@ -29,7 +29,31 @@ export const PaymentErrors = {
...
 export const CrossmintErrors = {
     ...PaymentErrors,
+    ...SmartWalletErrors,
+    NOT_AUTHORIZED: "not-authorized",
+    CROSSMINT_SERVICE: "crossmint-service",
+    JWT_EXPIRED: "not-authorized.jwt-expired",
+    JWT_INVALID: "not-authorized.jwt-invalid",
+    JWT_DECRYPTION: "not-authorized.jwt-decryption",
+    JWT_IDENTIFIER: "not-authorized.jwt-identifier",
+    OUT_OF_CREDITS: "out-of-credits",
```
devlyn37, line 56: Nice, these'll be useful in auth and basically every SDK moving forward

packages/client/wallets/smart-wallet/src/error/index.ts
devlyn37, line 1: Rather than using `CrossmintErrors` here, we should:
- Import `SmartWalletErrors` from the base SDK
- Alias that to and export it as `SmartWalletErrorCodes`
- Move all the generic errors (NotAuthorized, JWT....) to the `client-sdk-base`

That way:
- We export `SmartWalletErrorCode` and `SmartWalletError` which are more clear than `SmartWalletErrors` & `SmartWalletError`
- Other SDKs, Auth for example, will re-use those generic errors right away.

packages/client/wallets/smart-wallet/src/api/APIErrorService.ts
```diff
@@ -22,7 +25,7 @@ export type CrossmintAPIErrorCodes =
 
 export class APIErrorService {
     constructor(
-        private errors: Partial<Record<CrossmintAPIErrorCodes, (apiResponse: any) => SmartWalletError>> = {
+        private errors: Record<CrossmintAPIErrorCodes, (apiResponse: any) => SmartWalletError> = {
```
bryan-eastwood3, line 28: Unrelated but we didn't need this Partial

packages/client/wallets/smart-wallet/src/index.ts
```diff
@@ -14,14 +14,7 @@ export type {
...
 
 export {
-    TransferError,
-    CrossmintServiceError,
     SmartWalletError,
-    JWTDecryptionError,
-    JWTExpiredError,
-    JWTIdentifierError,
-    JWTInvalidError,
-    NotAuthorizedError,
```
devlyn37, line 24: Probably still want to export these from this package, still use for a dev and we don't want people to have to download both packages ideally.

devlyn37 (approved):
Left a couple comments to check out!

PR #659, opened 2024-08-13
Next Demo: React UI library integration (part 1)

jmderby:
## Description

This PR depends on `jonathan/client-sdk-react-ui-auth-provider` and contains the changes to use the new hook and auth provider.

This PR also contains various UI tweaks to correspond to the mock design better, thanks to the help of Pablo. 

## Test plan

Regular demo flow still works as expected.

PR #660, opened 2024-08-13
Client React UI SDK: Wallet provider wrapper for useAuth core (part 2)

jmderby:
## Description

This contains the changes for adding the smart wallet state within the useAuth hook from auth core by creating a wrapper provider and extending functionality of auth core.

Hooks added to React UI:
- `useAuth` extends regular auth core `useAuth` hook but adds wallet context.
- `CrossmintAuthProvider` extends regular `AuthProvider` from core but adds context provider and `createWallet` function.

## Test plan

Nextjs demo flow ([part 1)](https://github.com/Crossmint/crossmint-sdk/pull/659)  all works as expected using this PR's changes


packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -0,0 +1,56 @@
+import { createContext, useEffect, useMemo, useState } from "react";
+
+import { AuthProvider, useAuth as useAuthCore } from "@crossmint/client-sdk-auth-core";
+import type { AuthProviderParams } from "@crossmint/client-sdk-auth-core/src/contexts/authlogin";
+import { EVMSmartWallet, SmartWalletSDK } from "@crossmint/client-sdk-smart-wallet";
+
+export const WalletContext = createContext<{ wallet: EVMSmartWallet | null; isLoadingWallet: boolean }>({
+    wallet: null,
+    isLoadingWallet: false,
```
devlyn37, line 9: To conform to the PRD now, I think we should drop this. Within the demo we can check `jwt != null && wallet == null` as a replacement

packages/client/auth-core/package.json
```diff
@@ -34,11 +28,12 @@
...
     ],
     "scripts": {
-        "build": "tsup src/clientIndex.ts src/serverIndex.ts --clean --external react,react-dom --format esm,cjs --outDir ./dist --minify --dts --sourcemap",
-        "dev": "tsup src/clientIndex.ts src/serverIndex.ts --clean --external react,react-dom --format esm,cjs --outDir ./dist --dts --sourcemap --watch",
+        "build": "tsup src/index.ts --clean --format esm,cjs --outDir ./dist --minify --dts --sourcemap",
+        "dev": "tsup src/index.ts --clean --format esm,cjs --outDir ./dist --dts --sourcemap --watch",
         "test": "jest --passWithNoTests"
     },
     "dependencies": {
+        "@crossmint/client-sdk-smart-wallet": "workspace:*",
```
devlyn37, line 36: Don't think we need this here!

packages/client/ui/react-ui/src/hooks/useAuth.ts
```diff
@@ -0,0 +1,15 @@
+import { WalletContext } from "@/providers/CrossmintAuthProvider";
+import { useContext } from "react";
+
+import { useAuth as useAuthCore } from "@crossmint/client-sdk-auth-core";
+
+export function useAuth() {
+    const walletContext = useContext(WalletContext);
+
+    if (!walletContext) {
+        throw new Error("useAuth must be used within an WalletProvider");
```
devlyn37, line 10: Should this say "... within a CrossmintAuthProvider" instead, since that's what's exposed to external devs?

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -0,0 +1,57 @@
+import { createContext, useEffect, useMemo, useState } from "react";
+
+import { AuthProvider, type AuthProviderParams, useAuth as useAuthCore } from "@crossmint/client-sdk-auth-core";
+import { EVMSmartWallet, SmartWalletSDK } from "@crossmint/client-sdk-smart-wallet";
+
+type AuthWalletProviderParams = AuthProviderParams & {
+    embeddedWallets: {
+        type: "evm-smart-wallet";
+        defaultChain: "polygon-amoy" | "base-sepolia";
```
devlyn37, line 9: This should be of type `Chain` from the smart wallet package rather than just these two! I initially set it that way to speed up building a demo

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const smartWalletSDK = useMemo(() => SmartWalletSDK.init({ clientApiKey: props.apiKey }), [props.apiKey]);
+
+    const getOrCreateWallet = async (jwt: string) => {
+        try {
+            const wallet = await smartWalletSDK.getOrCreateWallet({ jwt }, props.embeddedWallets.defaultChain);
+            setWallet(wallet);
+        } catch (e: any) {
+            console.error("There was an error creating a wallet ", e);
+            throw e;
+        }
```
devlyn37, line 40: How do errors like this typically get exposed to external devs? How would an external dev catch and handle this?

devlyn37 (commented):
Left a couple comments and questions

There's also a few pieces you should annotate here!
- auth-core client/server exports => single export
- populating initial auth-core state w/ the session cookie

apps/payments/nextjs/package.json
```diff
@@ -36,6 +36,8 @@
     "devDependencies": {
         "@types/react": "18.2.21",
         "@types/react-dom": "18.2.7",
+        "@zerodev/passkey-validator": "5.4.0",
```
devlyn37, line 39: Could you pls make a ticket to investigate and fix later and include it in this file (ie. WAL-XXX)

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
devlyn37: On second thought I actually think this is fine for the initial release

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
jmderby: Good point, this runs only in the useEffect :thinking:
Maybe we just console.error for now and figure out error handling later. Wdy think?

devlyn37: sgtm

apps/payments/nextjs/package.json
jmderby: Ticket made, WAL-2732!

jmderby, line 39: I've made a ticket to capture this investigation. This is our workaround for now. 
WAL-2732: https://linear.app/crossmint/issue/WAL-2732/fix-appspaymentsnextjs-app-build-error

devlyn37 (approved):
Don't forget those annotations on the auth-core changes!

Approving w/ context here:
https://crossmint.slack.com/archives/C07EW047VA5/p1723753428848339?thread_ts=1723588171.769669&cid=C07EW047VA5

packages/client/auth-core/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export { useAuth, AuthProvider, type AuthProviderParams } from "./contexts/authlogin";
```
jmderby, line 1: We've removed the distinction between server/client here and just have a single file for exporting everything.

packages/client/auth-core/src/contexts/authlogin.tsx
```diff
@@ -17,14 +17,22 @@ const AuthContext = createContext<AuthContextType>({
...
 };
 
+const getJwtFromCookie = (): string | null => {
+    if (typeof document === "undefined") {
+        return null; // Check if we're on the client-side
+    }
+    const crossmintSession = document.cookie.split("; ").find((row) => row.startsWith("crossmint-session"));
+    return crossmintSession ? crossmintSession.split("=")[1] : null;
+};
+
```
jmderby, line 33: This function helps set the jwt on initial render right away rather than having to make a call, then set it afterwards.

packages/client/auth-core/src/index.ts
devlyn37: This fixed some build issues and simplifies things for now, can be reversed later if the need arises

jmderby:
Few annotations i'm bumping as a reminder: 
- https://github.com/Crossmint/crossmint-sdk/pull/660#discussion_r1719103468
- https://github.com/Crossmint/crossmint-sdk/pull/660#discussion_r1719102923
- https://github.com/Crossmint/crossmint-sdk/pull/660#discussion_r1718960995

PR #661, opened 2024-08-13
Smart Wallets: Client Caching Pt. 1 Organization

devlyn37:
## Description

I initially prototyped client caching in another branch, but the changes we're too big, so I'm breaking it down into a few steps. The first is lays the groundwork and organizes our classes.
- Extracting code for fetching smart wallet configuration into a new class: `AccountConfigFacade`. This is where we'll add caching logic soon.
- Updated `WalletCreationParams` and typeguards to be more robust and include `existingSignerConfig`

I've annotated other changes in the diff.

## Test plan

On the demo site, for both EOA and Passkey signer types, I:
- Created a new wallet, and minted it.
- Logged out.
- Logged back in and minted again.

### DIFF
--- a/packages/client/wallets/smart-wallet/src/SmartWalletSDK.ts
+++ b/packages/client/wallets/smart-wallet/src/SmartWalletSDK.ts
@@ -3,8 +3,12 @@ import { stringify } from "viem";
 import { validateAPIKey } from "@crossmint/common-sdk-base";
 
 import { CrossmintWalletService } from "./api/CrossmintWalletService";
-import { SmartWalletChain } from "./blockchain/chains";
+import type { SmartWalletChain } from "./blockchain/chains";
 import type { EVMSmartWallet } from "./blockchain/wallets";
+import { AccountConfigFacade } from "./blockchain/wallets/account/config";
+import { AccountCreator } from "./blockchain/wallets/account/creator";
+import { EOACreationStrategy } from "./blockchain/wallets/account/eoa";
+import { PasskeyCreationStrategy } from "./blockchain/wallets/account/passkey";
 import { ClientDecorator } from "./blockchain/wallets/clientDecorator";
 import { SmartWalletService } from "./blockchain/wallets/service";
 import { SmartWalletError } from "./error";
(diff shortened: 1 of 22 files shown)


packages/client/wallets/smart-wallet/src/SmartWalletSDK.ts
```diff
@@ -36,10 +40,18 @@ export class SmartWalletSDK {
...
+        const accountCreator = new AccountCreator(
+            new EOACreationStrategy(),
+            new PasskeyCreationStrategy(crossmintService.getPasskeyServerUrl(), clientApiKey)
         );
+        const smartWalletService = new SmartWalletService(
+            crossmintService,
+            new AccountConfigFacade(crossmintService),
+            accountCreator,
+            new ClientDecorator(errorProcessor)
+        );
```
devlyn37, line 52: Pulling all instantiation up to here while it's still manageable, it let's us inject more cleanly (ex. passing API key in directly to passkeyCreationStrategy, etc)

packages/client/wallets/smart-wallet/src/blockchain/wallets/account/config.ts
devlyn37, line 1: Everything in this class is the same as before, but now we return an instance of `SignerConfig` rather than the data directly, more context on that  annotated on `signer.ts`

packages/client/wallets/smart-wallet/src/blockchain/wallets/account/creator.ts
devlyn37, line 1: Changes:
- Moved and renamed the class
- Condensed error handling

packages/client/wallets/smart-wallet/src/blockchain/wallets/account/eoa.ts
devlyn37, line 1: Only changes here are adapting to `SignerConfig` class usage in `EOACreationParams`, and moving and renaming the class.

packages/client/wallets/smart-wallet/src/blockchain/wallets/account/passkey.ts
devlyn37, line 1: changes:
- move and rename class
- adapt to `signerConfig` usage
- Rather than injecting the entire `crossmintWalletService`, inject what we we're grabbing from it directly.

packages/client/wallets/smart-wallet/src/blockchain/wallets/account/signer.ts
```diff
@@ -0,0 +1,38 @@
+import type { EOASignerData, PasskeyDisplay, PasskeySignerData, SignerData, SignerDisplay } from "../../../types/API";
+
+export interface SignerConfig {
```
devlyn37, line 3: Wrapping the data in a class let's us keep data specific helpers close w/ actual data, and avoid big switch statements.

Added so I could use `display` in a more type safe and ergonomic way.

packages/client/wallets/smart-wallet/src/blockchain/wallets/service.ts
```diff
@@ -99,101 +79,6 @@ export class SmartWalletService {
             smartAccountClient: kernelAccountClient,
         });
 
-        return new EVMSmartWallet(this.crossmintWalletService, smartAccountClient, publicClient, chain);
-    }
-
-    private async fetchConfig(
```
devlyn37, line 105: Everything dropped here was moved to `AccountConfigFacade` in `config.ts`

bryan-eastwood3 (approved):

PR #662, opened 2024-08-13
SCW React demo: multiple auth providers to play around with

guiwoda:
## Description
This changes allow us to implement multiple authentication providers easily and test them out in the demo. I added Privy and Auth0 (with my project ID) but refactored around some code to make it easier to add more.

## Test plan

Run locally, make sure your project is configured with the corresponding `jwtConfig` pointing to the provider you are testing and modify the client api key accordingly.


apps/wallets/smart-wallet/react/src/app/App.tsx
```diff
@@ -14,4 +15,4 @@ function App() {
     );
 }
 
-export default App;
+export default withProviders(App);
```
jmderby, line 18: Overall the PR looks good. I'm not certain though that we should use a providers HOC here. I'm wondering if we should add them separately like how we're adding the <AppProvider> :thinking: 

I'm not sure what kind of implication that will have on your changes though. What do you think?

guiwoda: My reasoning was: I wanted a way to inject multiple auth providers, and wrapping the App around custom Provider components seemed to be the pattern for most of these React sdks.
I don't understand much of the implications tho!

jmderby: Totally fair and that makes sense. I'm not apposed to this I'm just concerned it's more complex than it needs to be. Lemme pull it in on my end and see what I can do.

jmderby: posting my commit here for visibility: https://github.com/Crossmint/crossmint-sdk/pull/662/commits/b2e1484356e2b9fd60539a6734a8a03cd2bf9455

devlyn37:
Closing since this is stale, pls re-open later if you want to land

PR #663, opened 2024-08-13
Smart Wallets: Client Caching Pt. 2 Runtime Wallet Config Validation

devlyn37:
## Description

I initially prototyped client caching in another branch, but the changes we're too big, so I'm breaking it down into a few steps. The second adds runtime smart wallet config validation, initially just to the response from `sdk/smart-wallet/config`, but we'll use this for the cached smart wallet config.

## Test plan

Ran the same tests as in https://github.com/Crossmint/crossmint-sdk/pull/661
Next PR I'll add unit tests that exercise the new schema


packages/client/wallets/smart-wallet/src/SmartWalletSDK.test.ts
```diff
@@ -1,5 +1,5 @@
 import { SmartWalletSDK } from "./SmartWalletSDK";
-import { SmartWalletSDKInitParams } from "./types/Config";
```
devlyn37, line 2: Initially I just changed the case to `types/config`, but that messed up the build, so rather than dive into what was going on there I just renamed the file XD.

packages/client/wallets/smart-wallet/src/api/CrossmintWalletService.ts
```diff
@@ -1,19 +1,20 @@
...
+import { SmartWalletConfigSchema } from "../types/schema";
+import type { SmartWalletConfig, StoreSmartWalletParams } from "../types/service";
 import { bigintsToHex, parseBigintAPIResponse } from "../utils/api";
+import { API_VERSION } from "../utils/constants";
 import { BaseCrossmintService } from "./BaseCrossmintService";
 
 export class CrossmintWalletService extends BaseCrossmintService {
-    async idempotentCreateSmartWallet(user: UserParams, input: StoreSmartWalletParams) {
-        return this.fetchCrossmintAPI(
+    async idempotentCreateSmartWallet(user: UserParams, input: StoreSmartWalletParams): Promise<void> {
```
devlyn37, line 16: tiny thing to make sure we're not relying on the data from here until there's validation

packages/client/wallets/smart-wallet/src/api/CrossmintWalletService.ts
```diff
@@ -37,22 +37,22 @@ export class CrossmintWalletService extends BaseCrossmintService {
...
             "Error getting smart wallet version configuration. Please contact support",
             user.jwt
         );
+
+        const result = SmartWalletConfigSchema.safeParse(data);
+        if (result.success) {
+            return result.data;
+        }
+
+        throw new CrossmintServiceError(
```
devlyn37, line 53: This error will only happen when something is broken with the crossmint service, it's not an expected error case, so surfacing the more raw zod error, which is the most useful for us to debug, is appropriate here.

packages/client/wallets/smart-wallet/src/blockchain/wallets/account/config.ts
```diff
@@ -32,22 +24,6 @@ export class AccountConfigFacade {
...
-            );
-        }
-
-        if (!isSupportedEntryPointVersion(entryPointVersion)) {
-            throw new SmartWalletError(
-                `Unsupported entry point version. Supported versions: ${SUPPORTED_ENTRYPOINT_VERSIONS.join(
-                    ", "
-                )}. Version used: ${entryPointVersion}. Please contact support`
-            );
-        }
```
devlyn37, line 49: validation replaced in the schema

packages/client/wallets/smart-wallet/src/blockchain/wallets/account/config.ts
```diff
@@ -58,28 +34,20 @@ export class AccountConfigFacade {
...
 
-    private getSigner(signers: any[]): SignerConfig | undefined {
+    private getSigner(signers: SignerData[]): SignerConfig | undefined {
         if (signers.length === 0) {
             return undefined;
         }
 
-        if (signers.length > 1) {
-            throw new CrossmintServiceError("Invalid wallet signer configuration. Please contact support");
-        }
```
devlyn37, line 80: replaced in schema

packages/client/wallets/smart-wallet/src/types/params.ts
```diff
@@ -31,5 +28,3 @@ export type EOASigner = EIP1193Provider | ViemAccount;
 export interface WalletParams {
     signer: EOASigner | PasskeySigner;
 }
-
-export type EntryPointDetails = { version: SupportedEntryPointVersion; address: EntryPoint };
```
devlyn37, line 35: Not needed anymore

packages/client/wallets/smart-wallet/src/api/CrossmintWalletService.ts
bryan-eastwood3, line 53: Wonder if it would make sense to make a utility function that calls safeParse and throws a CrossmintServiceError if it fails

bryan-eastwood3 (approved):

packages/client/wallets/smart-wallet/src/api/CrossmintWalletService.ts
devlyn37: I think so, but let's wait until there's a couple more instances of this pattern to be sure

PR #665, opened 2024-08-14
AA: Simple `callContract` function

bryan-eastwood3:
## Description

Implements a method according to the spec in [WAL-2560](https://linear.app/crossmint/issue/WAL-2560/simple-transaction-function).

Code is currently not tested and intended to be in a state for higher-level design feedback. Code is annotated with more information.

## Test plan

Tested locally the successful path and the revert path, unit tests for all of the error types


packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -0,0 +1,137 @@
...
+                return await this.handleReceipt(receipt, request);
+            } catch (e) {
+                if (e instanceof WaitForTransactionReceiptTimeoutError) {
+                    continue;
+                }
+                throw e;
+            }
+        }
+        // TODO discuss
+        throw new Error("Retries exceeded");
```
bryan-eastwood3, line 86: I'm open to input on how we should handle this case, the most orthodox way would be to just export an error `SendTransactionRetriesExceededError extends CrossmintSDKError`

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
```diff
@@ -112,4 +125,39 @@ export class EVMSmartWallet {
...
+            TAbi,
+            "nonpayable" | "payable",
+            TFunctionName
+        >
+    >({
+        address,
+        abi,
+        functionName,
+        args,
+        resendCallback,
```
bryan-eastwood3, line 145: This allows a dev using the SDK to e.g. make it so a popup appears that says "We need a new signature from your passkey" when we need a new signature to send a transaction. This may become less useful when we implement session keys but nice to have

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    private modifyGas(attempt: number, gas: EstimateFeesPerGasReturnType) {
+        if (gas.maxPriorityFeePerGas != null) {
+            gas.maxFeePerGas *= BigInt(this.gasIncreaseFactor * (attempt + 1));
+            gas.maxPriorityFeePerGas *= BigInt(this.gasIncreaseFactor * (attempt + 1));
+        }
+        if (gas.gasPrice != null) {
+            gas.gasPrice *= BigInt(this.gasIncreaseFactor * (attempt + 1));
+        }
+        return gas;
```
bryan-eastwood3, line 112: We should be fine without this but adding it for extra resilience

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -0,0 +1,137 @@
...
+export class SendTransactionExecutionRevertedError extends CrossmintSDKError {
+    constructor(
+        message: string,
+        public readonly viemError: BaseError,
+        public readonly revertError: ContractFunctionRevertedError,
+        public txId: string | undefined
+    ) {
+        super(message, CrossmintErrors.SEND_TRANSACTION_FAILED);
+    }
+}
```
bryan-eastwood3, line 44: The reason this error configuration was chosen is that there are others reasons that a transaction can't be sent, e.g.:
- Out of funds
- Nonce issues
- Gas issues

This approach allows us to surface those errors while also:
- Making it easy to handle execution reverted errors specifically
- Surfacing full information for execution reverted errors

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
```diff
@@ -112,4 +125,39 @@ export class EVMSmartWallet {
...
+        TArgs extends ContractFunctionArgs<TAbi, "nonpayable" | "payable", TFunctionName> = ContractFunctionArgs<
+            TAbi,
+            "nonpayable" | "payable",
+            TFunctionName
+        >
+    >({
+        address,
+        abi,
+        functionName,
+        args,
```
bryan-eastwood3, line 144: These 4 args are identical to the viem `writeContract` method which is used to call a function on a contract

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
```diff
@@ -112,4 +125,39 @@ export class EVMSmartWallet {
     public async nfts() {
         return this.crossmintService.fetchNFTs(this.address, this.chain);
     }
+
+    public async sendTransaction<
```
alfonso-paella, line 129: Can you add a tsdoc?

alfonso-paella: It's relevant for the review and we also get docs later for free

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
```diff
@@ -112,4 +125,39 @@ export class EVMSmartWallet {
...
+            "nonpayable" | "payable"
+        >,
+        TArgs extends ContractFunctionArgs<TAbi, "nonpayable" | "payable", TFunctionName> = ContractFunctionArgs<
+            TAbi,
+            "nonpayable" | "payable",
+            TFunctionName
+        >
+    >({
+        address,
+        abi,
```
alfonso-paella, line 142: What does `abi` need to have? Is it just the function, the entire contract, or something else?

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
```diff
@@ -112,4 +125,39 @@ export class EVMSmartWallet {
...
+        TFunctionName extends ContractFunctionName<TAbi, "nonpayable" | "payable"> = ContractFunctionName<
+            TAbi,
+            "nonpayable" | "payable"
+        >,
+        TArgs extends ContractFunctionArgs<TAbi, "nonpayable" | "payable", TFunctionName> = ContractFunctionArgs<
+            TAbi,
+            "nonpayable" | "payable",
+            TFunctionName
+        >
+    >({
```
alfonso-paella, line 140: I'm thinking we add:

`simulateBeforeSending: boolean = true` -- Simulates the transaction before sending it, and stops it from being sent if simulation failed
`retryConfig: { retryIfFailed: boolean, maxRetryAttempts?: int }` -- an object specifying some retry policy if the transaction failed. I made up the params, we shold put there whatever is relevant

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
```diff
@@ -112,4 +125,39 @@ export class EVMSmartWallet {
...
+        abi,
+        functionName,
+        args,
+        resendCallback,
+    }: Omit<
+        WriteContractParameters<TAbi, TFunctionName, TArgs, typeof this.accountClient.chain>,
+        "chain" | "account"
+    > & {
+        resendCallback?: TransactionServiceResendCallback;
+    }): Promise<Hex> {
```
alfonso-paella, line 151: As may be seen in comments above, we were thinking that this function would simplify devx greatly if it can do simulation and confirmation all in one call

If we do this, though, we need a more complex return type, where we can return failure errors of each of the stages. Midn thinking about that?

bryan-eastwood3: I think this function already does all of those. As for the return type, the idiom we're following in this code is to yield error information through typed exceptions. I think the two (very similar) options for the return type here are:
- Just the tx hash
- The transaction receipt

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
devlyn37, line 1: I think we're doing too much here now. We'd be better off starting minimal here and improving over time as we get feedback from clients, otherwise we're expanding our API surface and increasing scope quite a bit.

Main things on my mind:
- Do we really need the retries here? Maybe we can add them serverside later instead?
- Whenever possible let's surface `viem`, for reverts, and etc.

The main value imo that we get from this function is:
- guiding devs towards including their ABI and simulating against to reduce support load.
- Giving devs a single function they can call, rather than the three they'd otherwise need to call using `viem` clients directly.

Something that looks like:

```TS
sendTransaction(...){
    await this.client.public.simulate(...)
    const hash = await this.client.wallet.sendTransaction(...);
    return this.client.public.waitForTransactionReceipt({hash});
}
```

Could accomplish those goals in a simple way that doesn't give us much extra to maintain

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
devlyn37, line 129: We should probably be able to send eth with this function too right?

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
devlyn37: I'd argue that if a dev wants tweak any of these they'd be better of using the `viem` client

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
devlyn37: At a minimum:
- The function we're calling
- Any custom errors on the contract

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
alfonso-paella: Ok, that may work, but we should review the errors either here or in another PR. I would do it in this one (maybe in the TSDoc?)

packages/client/base/src/types/errors.ts
```diff
@@ -54,6 +54,8 @@ export const CrossmintErrors = {
     JWT_DECRYPTION: "not-authorized.jwt-decryption",
     JWT_IDENTIFIER: "not-authorized.jwt-identifier",
     OUT_OF_CREDITS: "out-of-credits",
+    SEND_TRANSACTION_FAILED: "send-transaction-failed",
+    SEND_TRANSACTION_EXECUTION_REVERTED: "send-transaction-execution-reverted",
```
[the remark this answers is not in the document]
bryan-eastwood3: Discussed in slack and addressed

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
bryan-eastwood3: @alfonso-paella can you share your thoughts?

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
bryan-eastwood3: Current errors are in SendTransactionService.ts

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
devlyn37: I'm not sure if bundlers enforce that same rule. But in either case I'd still advocate for not having retries atm.
Consider how complicated a single function call, involving multiple passkeys signatures, would look like!

devlyn37: > Can you see a way to improve this vs the current implementation? My understanding is that by making the ABI required in this call viem will automatically decode the errors

The current impl hits this requirement!

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
bryan-eastwood3: I'd agree with Devlyn here on the first two. The retry settings are specified in `resendCallback`

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
alfonso-paella: It's hard for me to tell from the PR. What validation / errors do we return if they botched it? 

This seems like a probably place to screw up

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
alfonso-paella: Hmm we can add them later i guess, so nbd

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
bryan-eastwood3: Right now it'll return a `SendTransactionError` containing the viem error if they don't specify the ABI entry corresponding to the function name they are calling. It's unfortunately not really possible for us to check that they are including the ABI entries for errors, but any modern tool that exports ABIs should include them

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
bryan-eastwood3: Ok, bottom line is I'm ok with removing the retry logic if @alfonso-paella agrees

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
alfonso-paella: Actually what is the confirmation behavior currently? That one does seem worth exposing to me as i dont think there's a one size fits all solution

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
alfonso-paella: > @alfonso-paella can you share your thoughts?

On retries -- I think it's ok to start with no retries, and add it later, as it seems non trivial e.g. handling passkeys, max gas thresholds, max timeouts, and so on. And i think we can add it later without breaking api changes

On errors - its hard for me to tell by looking at this PR. Is it possible to put on a google doc or here inline an example of the dev experience alternatives we're considering here?

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
```diff
@@ -126,6 +126,16 @@ export class EVMSmartWallet {
...
 
+    /**
+     * Sends a contract call transaction and returns the hash of a confirmed transaction.
+     * @param address the address of the contract to be called
+     * @param abi the ABI of the contract - ***should be defined as a typed variable*** to enable type checking of the contract arguments, see https://viem.sh/docs/typescript#type-inference for guidance
+     * @param functionName the name of the function to be called
+     * @param args the arguments to be passed to the function
+     * @param resendCallback a callback function that will be called if the transaction needs to be resent. Returns a value that controls transaction rebroadcasting See the type of `TransactionServiceResendCallback` for more information.
+     * @returns The transaction hash.
+     * @throws `SendTransactionError` if the transaction fails to send. Will be subclass `SendTransactionExecutionRevertedError` if the transaction reverted either during simulation or on-chain.
```
alfonso-paella, line 137: Can we expand this into two @throws lines?

SendTransactionExecutionRevertedError: if the transaction reverted during simulation or on chain.
SendTransactionError: ???

also dont we have network errors, and other reasons why a TX may not be sent?

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
```diff
@@ -126,6 +126,16 @@ export class EVMSmartWallet {
         return this.crossmintService.fetchNFTs(this.address, this.chain);
     }
 
+    /**
+     * Sends a contract call transaction and returns the hash of a confirmed transaction.
+     * @param address the address of the contract to be called
+     * @param abi the ABI of the contract - ***should be defined as a typed variable*** to enable type checking of the contract arguments, see https://viem.sh/docs/typescript#type-inference for guidance
+     * @param functionName the name of the function to be called
```
alfonso-paella, line 133:
```suggestion
     * @param functionName the name of the smart contract function to be called
```

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
alfonso-paella: > Out of funds
> Nonce issues
> Gas issues

How do these look? Are these SendTransactionErrors or specific subclasses? I think we should document all in the TSDoc of send transaction if they have different handlings

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
```diff
@@ -126,6 +126,16 @@ export class EVMSmartWallet {
...
+     * @param args the arguments to be passed to the function
```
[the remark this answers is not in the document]
bryan-eastwood3: No

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
bryan-eastwood3: > also dont we have network errors, and other reasons why a TX may not be sent?
This is the benefit of wrapping the errors, we can choose which errors (like execution reverted) we want to have special handling for (e.g. adding tenderly links in the future) while also not having to write a custom error for everything that can fail when sending a transaction, which is an extremely difficult task

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
bryan-eastwood3: I added confirmation options

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
bryan-eastwood3: See https://github.com/Crossmint/crossmint-sdk/pull/665/files#r1717524133. We can't hope to write custom errors for everything that can go wrong when sending a transaction, so we include the viem error in any error thrown by this function while also including subclasses that represent common failure scenarios

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
bryan-eastwood3: Honestly reading the jsdoc for `sendTransaction` + the jsdoc for the two errors in `SendTransactionService.ts` should be sufficient to understand, as well as [this comment](https://github.com/Crossmint/crossmint-sdk/pull/665/files#r1717524133) for motivation. If it's not clear from these jsdocs then we should address that

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -0,0 +1,142 @@
...
+    address: Address;
+    abi: Abi;
+    functionName: string;
+    args: any;
+    value?: bigint;
+};
+
+/**
+ * Error thrown when a transaction fails to send.
+ * @param viemError The error thrown by the viem client.
```
alfonso-paella, line 27: Can we link to viem docs here?

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -0,0 +1,142 @@
...
+ */
+export class SendTransactionError extends CrossmintSDKError {
+    constructor(message: string, public readonly viemError: BaseError, code = WalletErrorCode.SEND_TRANSACTION_FAILED) {
+        super(message, code);
+    }
+}
+
+/**
+ * Error thrown when a transaction fails due to a contract execution error.
```
alfonso-paella, line 36: Handling execution reverted issues tends to be the biggest pain point for devs in checkout and it may be here too

I would make instructions in this TSDoc as easy as possible

For example we could explain: when to use the viemError vs revertError, and show an example error object maybe?

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -0,0 +1,142 @@
...
+    abi: Abi;
+    functionName: string;
+    args: any;
+    value?: bigint;
+};
+
+/**
+ * Error thrown when a transaction fails to send.
+ * @param viemError The error thrown by the viem client.
+ */
```
alfonso-paella, line 28: Would it be clear from these errors whether the error originated during simulation, sending or confirming? If not, it may make sense to add that as a parameter in the error

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -0,0 +1,142 @@
...
+export class SendTransactionError extends CrossmintSDKError {
```
alfonso-paella, line 29: The wallets SDK is going to support solana and other chains in the future, that don't use viem

As these errors are coupled to viem (whcih is fine), should we maybe include EVM somewhere in the error class name? 

e.g. EVMSendTransactionError

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
bryan-eastwood3: I added two subclasses, `SendTransactionSimulationError` and `SendTransactionConfirmError` for errors that are thrown in those respective stages

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
bryan-eastwood3: I added some field descriptions and an example of how to catch a revert error and interpret it

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
bryan-eastwood3: Changed

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -24,20 +25,64 @@ export type TransactionServiceTransactionRequest = {
...
+ *   });
+ * } catch (e) {
+ *   if (e instanceof SendTransactionExecutionRevertedError) {
+ *     alert(`Transaction reverted: ${e.revertError.reason}`);
+ *   }
+ *   throw e;
+ * }
  */
-export class SendTransactionExecutionRevertedError extends SendTransactionError {
+export class SendTransactionExecutionRevertedError extends EVMSendTransactionSimulationError {
```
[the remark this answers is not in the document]
alfonso-paella: ok. i would triple check w devlyn if you want another opinion but im ok either way

alfonso-paella (approved):

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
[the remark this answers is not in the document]
bryan-eastwood3: @devlyn37 feel free to chime in

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.test.ts
devlyn37, line 1: Main is using vitest for this SDK and already has `vitest-mock-extended`, so should be easy to move things over. Was updated [here](https://github.com/Crossmint/crossmint-sdk/pull/671)

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
devlyn37: Why wouldn't someone who cares about that just use the viem client though?

It's only a few LOC more to use and gives them all the options they could ever want.

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
```diff
@@ -112,4 +125,68 @@ export class EVMSmartWallet {
...
         return this.crossmintService.fetchNFTs(this.address, this.chain);
     }
+
+    /**
+     * Sends a contract call transaction and returns the hash of a confirmed transaction.
+     * @param address the address of the contract to be called
+     * @param abi the ABI of the contract - ***should be defined as a typed variable*** to enable type checking of the contract arguments, see https://viem.sh/docs/typescript#type-inference for guidance
+     * @param functionName the name of the smart contract function to be called
+     * @param args the arguments to be passed to the function
+     * @param resendCallback a callback function that will be called if the transaction needs to be resent. Returns a value that controls transaction rebroadcasting See the type of `TransactionServiceResendCallback` for more information.
```
devlyn37, line 135: Not used anymore

devlyn37 (commented):
Quick initial review before I do the full thing

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
```diff
@@ -112,4 +125,68 @@ export class EVMSmartWallet {
...
+     * }] as const;
+     *
+     * await wallet.sendTransaction({
+     *   address: contractAddress,
+     *   abi,
+     *   functionName: "mintNFT",
+     *   args: [recipientAddress],
+     * });
+     */
+    public async sendTransaction<
```
devlyn37, line 163: afaik this doesn't support sending eth to another address, it should right?

Maybe something like:

```TS
account.sendTransaction({
            contractWrite: {
                abi: CollectionABI,
                functionName: "mintToCustomErrorTester",
                args: [account.address, BigInt(98)],
            }
            address: "0x5c030a01e9d2c4bb78212d06f88b7724b494b755",
            value: BigInt(5),
});
```

With :
- `contractWrite` as an optional property of the function parameters, AND
- all properties of `contractWrite` as non optional.

That way clients/dev must always include the ABI when interacting with a contract

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -0,0 +1,195 @@
...
+
+/**
+ * Error thrown when a transaction fails to send.
+ * @param viemError The error thrown by the viem client. See https://viem.sh/docs/glossary/errors.html
+ */
+export class EVMSendTransactionError extends CrossmintSDKError {
+    constructor(message: string, public readonly viemError: BaseError, code = WalletErrorCode.SEND_TRANSACTION_FAILED) {
+        super(message, code);
+    }
+}
```
devlyn37, line 34: @bryan-eastwood3 what do you see as the main value prop for having these extra errors, compared to letting the viem ones bubble up?

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -0,0 +1,195 @@
...
+        return {
+            ...this.defaultSendTransactionOptions,
+            ...config,
+        };
+    }
+
+    private async handleReceipt(receipt: TransactionReceipt, request: TransactionServiceTransactionRequest) {
+        if (receipt.status === "reverted") {
+            // This should revert and throw the full reason
+            await this.simulateCall(request, receipt.transactionHash);
```
devlyn37, line 160: good idea

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
devlyn37: Is there any comparisons somewhere?

bryan-eastwood3: Two main things:

- If we just bubble up the raw viem error, we can't add any functionality, e.g. if we wanted to always include a tenderly link in the error
- We generally have more ability to optimize the experience for long tail devs, e.g. by including the viem revert error as a field in our revert error a dev won't have to call `walk` and get it themselves (obviously very basic but we want to make this flow very frictionless). Another example would be us wrapping errors in different subclasses that indicate at which point the transaction failed (`EVMSendTransactionConfirmationError` for confirmation stage, `EVMSendTransactionSimulationError` for simulation)

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -0,0 +1,195 @@
...
+ *     alert(`Transaction reverted: ${e.revertError.reason}`);
+ *   }
+ *   throw e;
+ * }
+ */
+export class EVMSendTransactionExecutionRevertedError extends EVMSendTransactionSimulationError {
+    constructor(
+        message: string,
+        public readonly viemError: BaseError,
+        public readonly revertError: ContractFunctionRevertedError,
```
devlyn37, line 89: Rather than revertError here, I'd consider directly putting in relevant properties from it instead
  
```TS
data?: DecodeErrorResultReturnType | undefined
reason?: string | undefined
signature?: Hex | undefined
```

They
- are easier to access
- clear up any confusion from having two full viem errors, that could very well be the same.

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -0,0 +1,195 @@
...
+    ): Promise<Hex> {
+        const { confirmations, transactionConfirmationTimeout, awaitConfirmation } = this.getConfig(config);
+        try {
+            await this.simulateCall(request, undefined);
+            const hash = await client.writeContract({
+                ...request,
+                account: client.account,
+                chain: client.chain,
+            });
+            if (awaitConfirmation) {
```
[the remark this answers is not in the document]
devlyn37: If there's any other config we never actually use I'd also consider dropping!

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
devlyn37: Also applies to `EVMSendTransactionSimulationError`

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -0,0 +1,195 @@
...
+ *     args: [recipientAddress],
+ *   });
+ * } catch (e) {
+ *   if (e instanceof SendTransactionExecutionRevertedError) {
+ *     alert(`Transaction reverted: ${e.revertError.reason}`);
+ *   }
+ *   throw e;
+ * }
+ */
+export class EVMSendTransactionExecutionRevertedError extends EVMSendTransactionSimulationError {
```
devlyn37, line 85: Maybe we should extend `EVMSendTransactionError` here instead, failing on simulation and having a tx fail are pretty different cases. I'd worry about devs checking `e instanceof EVMSendTransactionSimulationError` and handling this in an inappropriate way!

devlyn37 (commented):
Left a few comments and suggestions!

Will also need to see that testing section filled out to approve here! Would be good to run some local tests and include example outputs and errors, in addition to the unit tests you added.

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
bryan-eastwood3: We can't make it extend EVMSendTransaction obviously, because then it would be incorrect in the simulation case. I just added a variable "stage" that indicates at what stage the error came from as per Alfonso's suggestion

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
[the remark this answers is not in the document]
bryan-eastwood3: Alfonso asked for it to be included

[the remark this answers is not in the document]
devlyn37: We're not exposing it through the API though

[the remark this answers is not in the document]
devlyn37: So seems like we should either delete the functionality, or expose the option through `callContract`

@alfonso-paella I think we should skip the confirmations for now, and wait until someone asks for it. I suspect anyone who's sophisticated enough to care will use the viem clients.

Also, it's not a breaking change to add this later. The reverse is not true.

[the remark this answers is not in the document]
bryan-eastwood3: Whoops you're right. @alfonso-paella I'd tend to agree that we shouldn't expose the option to not wait for confirmations, anyone who is just interested in landing a transaction from a smart wallet would want it enabled and anyone doing something that would require disabling it (e.g. their own transaction management logic) would almost certainly be using the viem client

packages/client/base/src/types/errors.ts
```diff
@@ -44,9 +44,18 @@ export const SmartWalletErrorCode = {
 } as const;
 export type SmartWalletErrorCode = (typeof SmartWalletErrorCode)[keyof typeof SmartWalletErrorCode];
 
+export const WalletErrorCode = {
+    SEND_TRANSACTION_FAILED: "wallet:send-transaction-failed",
+    SEND_TRANSACTION_EXECUTION_REVERTED: "wallet:send-transaction-execution-reverted",
+    SEND_TRANSACTION_CONFIRMATION_FAILED: "wallet:send-transaction-confirmation-failed",
+    SEND_TRANSACTION_SIMULATION_FAILED: "wallet:send-transaction-simulation-failed",
+};
+export type WalletErrorCode = (typeof WalletErrorCode)[keyof typeof WalletErrorCode];
```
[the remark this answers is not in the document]
bryan-eastwood3: I think so, this is in the base package

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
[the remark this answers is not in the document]
alfonso-paella: I forgot but I think what I'd suggested in the end is we allow a customizable number of blocks, not that we allow turning awaiting confirmation off

Can we do that?

Because some use cases may prefer low latency, others may prefer low chance of reorg

Is that hard to do?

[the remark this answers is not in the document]
bryan-eastwood3: I added two options to the SDK call:
- Transaction confirmations
- Confirmation timeout (how long to wait before throwing an error)

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -0,0 +1,191 @@
...
+    ) {}
+
+    async sendTransaction(
+        request: TransactionServiceTransactionRequest,
+        client: SmartAccountClient<EntryPoint, Transport, Chain, SmartAccount<EntryPoint>>,
+        config: Partial<SendTransactionOptions> = {}
+    ): Promise<Hex> {
+        const { confirmations, transactionConfirmationTimeout } = this.getConfig(config);
+
+        // Simulate
```
devlyn37, line 106: nit: this comment may be redundant aha

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -0,0 +1,191 @@
...
+                timeout: transactionConfirmationTimeout,
+            });
+            return await this.handleReceipt(receipt, request);
+        } catch (e) {
+            if (e instanceof BaseError) {
+                throw new EVMSendTransactionError(e.message, e, "confirmation");
+            }
+            throw e;
+        }
+    }
```
devlyn37, line 138: nit: are these comments needed?

`simulateCall`, `writeContract`, and `waitForTransactionReceipt` are probably descriptive enough

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
```diff
@@ -110,4 +123,71 @@ export class EVMSmartWallet {
     public async nfts() {
         return this.crossmintService.fetchNFTs(this.address, this.chain);
     }
+
+    /**
+     * Sends a contract call transaction and returns the hash of a confirmed transaction.
```
devlyn37, line 128: Need to update naming in the jsdoc still!

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
```diff
@@ -110,4 +123,71 @@ export class EVMSmartWallet {
...
+     * }] as const;
+     *
+     * await wallet.sendTransaction({
+     *   address: contractAddress,
+     *   abi,
+     *   functionName: "mintNFT",
+     *   args: [recipientAddress],
+     * });
+     */
+    public async callContract<
```
devlyn37, line 160: `Call` might to overloaded, it has different meanings in both
[viem](https://docs.ethers.org/v5/api/providers/provider/#Provider-call) and [ethers](https://docs.ethers.org/v5/api/providers/provider/#Provider-call)

Claude suggested `sendContractTransaction`, I think `writeContract` from `viem` also makes a lot of sense imo

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -0,0 +1,191 @@
...
+ *     args: [recipientAddress],
+ *   });
+ * } catch (e) {
+ *   if (e instanceof SendTransactionExecutionRevertedError) {
+ *     alert(`Transaction reverted: ${e.revertError.reason}`);
+ *   }
+ *   throw e;
+ * }
+ */
+export class EVMSendTransactionExecutionRevertedError extends EVMSendTransactionError {
```
devlyn37, line 64: Any downside to having something more concise, maybe `EVMTransactionRevertedError` instead?

devlyn37: Base error could be `EVMTransactionError`

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -0,0 +1,191 @@
...
+    constructor(
+        message: string,
+        public readonly viemError: BaseError,
+        revertError: ContractFunctionRevertedError,
```
devlyn37, line 68: Now that we've exposed it's useful properties here, I'm not sure there's a need for `revertError` to be included.

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -0,0 +1,191 @@
...
+            // Otherwise, throw a generic error (this should practically never happen)
+            throw new EVMSendTransactionExecutionRevertedError(
+                "Transaction reverted but unable to detect the reason",
+                new ContractFunctionRevertedError({ abi: request.abi as Abi, functionName: request.functionName }),
+                new ContractFunctionRevertedError({ abi: request.abi as Abi, functionName: request.functionName }),
+                receipt.transactionHash,
+                "execution"
+            );
+        }
+        return receipt.transactionHash;
```
devlyn37, line 160: Would be good to test this locally if you haven't already

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.test.ts
```diff
@@ -0,0 +1,140 @@
...
+                {
+                    address: zeroAddress,
+                    abi: [],
+                    functionName: "mockFunction",
+                    args: [],
+                },
+                mockAccountClient
+            )
+        ).rejects.toThrow(EVMSendTransactionExecutionRevertedError);
+    });
```
devlyn37, line 75: You could make these more robust by also checking the stage like you do in the test below

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.test.ts
devlyn37, line 1: We should test that our txn revert errors include:
- Decoded errors when applicable
- Revert string when applicable
Since it's such an important piece business wise for the new function

devlyn37 (commented):
One more round XD

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
bryan-eastwood3: Good catch

packages/client/wallets/smart-wallet/src/blockchain/wallets/EVMSmartWallet.ts
bryan-eastwood3: I'm mostly ambivalent but I think to the average developer callContract is clearer than writeContract at least. If you think we should change it could you write a bit more about why in the GC with Alfonso and I?

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
bryan-eastwood3: It's practically impossible to do so under AA since the bundler etc simulate the transaction before it lands, but I tested all of the components that get called here (e.g. that waitForTransaction doesn't throw an error on revert and returns the status "reverted") and unit tested that they work together in this flow

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
bryan-eastwood3: My personal opinion is that clarity is more important than brevity, worth mentioning that viem seems to follow them same philosophy with their errors: https://viem.sh/docs/glossary/errors.html

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
bryan-eastwood3: It's not included, it's best to pass it to the constructor so that the reassignment logic can be contained in the error

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.test.ts
bryan-eastwood3: - Decoded errors are returned in the "data" field copied from the viem error
- Revert strings are returned in the "reason" field

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
devlyn37: Ahh, ofc. misread this

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
devlyn37: I'm not sure the `send` or `execution` add more clarity here, but I could just be too used to hearing all the terminology often.

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
devlyn37: ok!

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
```diff
@@ -0,0 +1,186 @@
...
+}
+
+/**
+ * Error thrown when a transaction fails due to a contract execution error.
+ * @param viemError The error thrown by the viem client. See https://viem.sh/docs/glossary/errors.html
+ * @param data The name and arguments of the revert error from the provided ABI
+ * @param reason The revert error if it is a string and not an ABI error
+ * @example
+ * try {
+ *   await wallet.sendTransaction({
```
devlyn37, line 50: This still needs to be updated.

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.test.ts
devlyn37: Yeah I saw how the code works, imo worth having cursor write you a regression test, but will not block PR on it.

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
devlyn37: If it's practically impossible to exercise this code, is it worth having?

devlyn37 (approved):
LGTM, but let's get alfonso's opinion on the function name again before landing

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.ts
bryan-eastwood3: I think it's good to have a branch that handles this case so we never yield a tx hash that contains a reverted transaction, even if we can't simulate it ourselves under ideal conditions

packages/client/wallets/smart-wallet/src/blockchain/wallets/SendTransactionService.test.ts
bryan-eastwood3: Added a test

PR #667, opened 2024-08-14
Smart Wallets: Client Caching Pt. 3 Actually Do The Caching

devlyn37:
## Description

This PR finishes adding client side caching for smart wallet config, which:
- Makes most calls to `SDK.getOrCreateWallet` not require a roundtrip to the Crossmint services.
- Reduces execution time for `SDK.getOrCreateWallet` by ~ 50% when the cache active. (At least during local development, will get a more accurate picture once this lands)
- Opens the door to sub 100 millisecond `SDK.getOrCreateWallet` calls, but this'll require us to make a couple small updates to the ZeroDev SDK.

The PR also revises to the format of data we pass around within the SDK a little bit.

I've annotated the diff to make reviewing easier.
Don't be scared by the line count here, the majority are from lock file changes & unit tests!

## Test plan

Local / Manual:
- Went through the wallet creation flow, minted, refreshed the page, then minted again, confirming:
    - both mints work as expected
    - we're pulling config from the cache
    
- Unit:
    - Added unit tests for the key classes for these changes `AccountConfigService`, and `AccountConfigCache`.


apps/wallets/smart-wallet/react/src/app/auth/FirebaseAuthManager.ts
```diff
@@ -59,7 +59,7 @@ export const checkAuthState = (): Promise<string | undefined> => {
         onAuthStateChanged(auth, async (user) => {
             try {
                 if (user) {
-                    const jwt = await user.getIdToken(true);
```
devlyn37, line 62: This `true` is an optional property that forces a refreshed jwt every call. We don't need that!

packages/client/wallets/smart-wallet/src/SmartWalletSDK.ts
```diff
@@ -44,10 +46,11 @@ export class SmartWalletSDK {
             new EOACreationStrategy(),
             new PasskeyCreationStrategy(crossmintService.getPasskeyServerUrl(), clientApiKey)
         );
+        const accountCache = new AccountConfigCache(`smart-wallet-${SDK_VERSION}`);
```
devlyn37, line 49: Include the SDK version to invalid cache entries for older versions of the sdk in case there's been updates to how the data is stored.

packages/client/wallets/smart-wallet/src/blockchain/wallets/account/cache.ts
```diff
@@ -0,0 +1,43 @@
...
+            const key = localStorage.key(i);
+            if (key && key.startsWith(this.keyPrefix)) {
+                localStorage.removeItem(key);
+                i--; // Decrement i since we've removed an item
+            }
+        }
+    }
+
+    private key(user: UserParams) {
+        return `${this.keyPrefix}-${keccak256(toHex(user.jwt))}`;
```
devlyn37, line 41: Has the jwt is included in the key to automatically invalid the cache at the same cadence JWTs are refreshed in an external developers app.

packages/client/wallets/smart-wallet/src/blockchain/wallets/account/config.ts
```diff
@@ -1,29 +1,72 @@
...
-        userId: string;
-        existingSignerConfig?: SignerConfig;
-        smartContractWalletAddress?: Address;
-    }> {
-        const { entryPointVersion, kernelVersion, signers, smartContractWalletAddress, userId } =
-            await this.crossmintService.getSmartWalletConfig(user, chain);
+        user: UserParams & { id: string };
+        existing: PreExistingWalletProperties;
+    }) {
+        this.configCache.clear();
```
devlyn37, line 56: Clearing the cache here is the simplest way to keep it tidy, at the expense of not caching wallets for multiple users wallets.

Since that's not a use case we expect atm, opting for simplicity.

packages/client/wallets/smart-wallet/src/blockchain/wallets/account/config.ts
devlyn37, line 1: Major changes here:
- Added caching.
- Enforce that both the signer & address must either be null or defined, and updated the data being passed around to reflect that.

packages/client/wallets/smart-wallet/src/blockchain/wallets/account/creator.ts
devlyn37, line 1: Just some renamed vars and functions here

packages/client/wallets/smart-wallet/src/blockchain/wallets/account/eoa.ts
devlyn37, line 1: Updates here are only adapting to the new format data is passed around in

packages/client/wallets/smart-wallet/src/blockchain/wallets/account/passkey.ts
devlyn37, line 1: Updates here are only adapting to the new format data is passed around in

packages/client/wallets/smart-wallet/src/types/internal.ts
devlyn37, line 1: A couple minor changes here to better represent the data we're passing around

packages/client/wallets/smart-wallet/src/blockchain/wallets/account/cache.ts
```diff
@@ -0,0 +1,43 @@
+import { keccak256, toHex } from "viem";
+
+import { UserParams } from "../../../types/params";
+import { SmartWalletConfigSchema } from "../../../types/schema";
+import type { SmartWalletConfig } from "../../../types/service";
+
+export class AccountConfigCache {
```
bryan-eastwood3, line 7: I wonder if we should generalize this to be usable by other objects besides account configs. You'd have a better sense on how useful this would be

bryan-eastwood3 (approved):

packages/client/wallets/smart-wallet/src/blockchain/wallets/account/cache.ts
devlyn37: Next time we need to cache something I think this would make a lot of sense, but this PR brings network requests that are within our control (ie. not within the ZD SDK) to 0 for the average request, so not needed yet!

PR #676, opened 2024-08-16
Move BaseCrossmintService & logging to base package

bryan-eastwood3:
## Description

Changes in this PR:

- Move `BaseCrossmintService` to base package
  - In order to facilitate this, refactor `APIErrorService` to take package-specific errors in the constructor so we don't import them into the base package
- Change our logging to use an injected class, which was a requirement to move `BaseCrossmintService` to the base package and still retain package-specific log tags
  - Create `SDKLogger` in base package which takes the service name as a constructor parameter, which can be instantiated once per package and injected into all classes that emit logs

## Test plan

Tested both base package level error and smart wallet specific error flows and confirmed they still work


packages/client/base/src/utils/SDKLogger.ts
```diff
@@ -0,0 +1,53 @@
+import { getBrowserLogger } from "../services/logging";
+
+export class SDKLogger {
```
devlyn37, line 3: If SDKLogger was an abstract class, with abstract methods: `log`, `warn`, and `error`. With two implementations: `DatadogSDKLogger` & `ConsoleSDKLogger`, this would be a little tighter. There's probably no need to have two different logging interfaces / classes (SDKLogger & BrowserLoggerInterface).

Non blocking, either way I think injecting this class is nice improvement.

packages/client/wallets/smart-wallet/src/SmartWalletSDK.ts
```diff
@@ -39,7 +39,7 @@ export class SmartWalletSDK {
         }
 
         const crossmintService = new CrossmintWalletService(clientApiKey);
-        const errorProcessor = new ErrorProcessor(new DatadogProvider());
+        const errorProcessor = new ErrorProcessor(scwLogger);
```
devlyn37, line 42: The ErrorProcessor class logs every SDK errors, we want it always to be a datadog logger so we can see what issues devs run into most during development.

packages/client/base/package.json
```diff
@@ -24,6 +24,7 @@
     },
     "dependencies": {
         "@crossmint/common-sdk-base": "workspace:*",
+        "@datadog/browser-logs": "4.42.2",
```
devlyn37, line 27: Think you missed deleting this from the smart wallet SDK package.json

packages/client/base/src/services/api/BaseCrossmintService.ts
```diff
@@ -1,14 +1,16 @@
...
-import { logPerformance } from "../utils/log";
+import { CROSSMINT_DEV_URL, CROSSMINT_PROD_URL, CROSSMINT_STG_URL } from "../../consts";
 import { APIErrorService } from "./APIErrorService";
 
 export abstract class BaseCrossmintService {
     public crossmintAPIHeaders: Record<string, string>;
     protected crossmintBaseUrl: string;
-    protected apiErrorService: APIErrorService;
+    protected abstract apiErrorService: APIErrorService<string>;
+    protected abstract logger: SDKLogger;
```
devlyn37, line 13: You consider injecting these two instead? I don't think there's a downside, and then you can use this w/o defining a subclass, which would be useful for SDKs that only hit a single endpoint or two, or that don't really need any special error handling.

packages/client/wallets/smart-wallet/src/utils/log.ts
```diff
@@ -1,47 +1,10 @@
...
-        }
-        if (typeof json[prop] == "function") {
-            continue;
-        }
-        simpleObject[prop] = json[prop];
-    }
-    return JSON.stringify(simpleObject, null, 2); // returns cleaned up JSON
-}
-
-export function errorToJSON(error: Error | unknown) {
```
devlyn37, line 40: This function is redundant, feel free to delete it too if you'd like.

Only used in one spot, and the spot it's used would be better served by `error.message`

packages/client/wallets/smart-wallet/src/services/index.ts
devlyn37, line 1: Needed, doesn't look you're always importing from `*/services` anyways

devlyn37 (commented):
Left some comments and questions, only two things I'd consider blocking:
- Datadog logger in error processor comment.
- Deleting the unused dep from `/smart-wallet/package.json` comment

packages/client/base/src/services/api/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./APIErrorService";
```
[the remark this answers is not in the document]
bryan-eastwood3: Yes, so product-specific packages can use it

packages/client/wallets/smart-wallet/src/SmartWalletSDK.ts
bryan-eastwood3: Fixed

packages/client/base/src/services/api/BaseCrossmintService.ts
bryan-eastwood3: I played around with this a bit and I think abstract properties is a bit cleaner here. Dependency injection gets messy fast when not using a DI framework. Also keep in mind that BaseCrossmintService is abstract!

packages/client/wallets/smart-wallet/src/services/index.ts
bryan-eastwood3: Sorry can you explain?

devlyn37: *Needed?

devlyn37 (approved):
Approved assuming you fix CI and have tested latest changes.

packages/client/wallets/smart-wallet/src/services/index.ts
bryan-eastwood3: Probably no reason not to keep it around in case we add more stuff to that directory

PR #677, opened 2024-08-16
Smart Wallets: Export Types From `react-ui` Package + Cleanup

devlyn37:
## Description

This PR:
- Removes uneeded deps from the new smart wallet demo
- Exports a couple smart wallet types from `@crossmint/client-sdk-react-ui` that we're needed as a result.

## Test plan

Went through the demo flow, logged in, minted an NFT.


apps/wallets/smart-wallet/next/src/env.ts
devlyn37, line 1: This is overkill for our single env var

packages/client/ui/react-ui/src/index.ts
```diff
@@ -3,4 +3,5 @@ export * from "./hooks";
 export * from "./providers";
 
 export { CrossmintEvents, useCrossmintEvents } from "@crossmint/client-sdk-base";
+export type { EVMSmartWallet, Chain } from "@crossmint/client-sdk-smart-wallet";
```
devlyn37, line 6: Removing `@crossmint/client-sdk-smart-wallet` as a dependency from the demo package revealed we need to export these.

jmderby (approved):
Nice, cleaned up quite a bit

apps/wallets/smart-wallet/next/src/env.ts
jmderby: good call

PR #687, opened 2024-08-20
Auth: use jwks endpoint to verify the token 

MatiNavCrossmint:
## Description

*Blocked by base release*

Auth: use jwks endpoint to verify the token 

## Test plan

Tested locally


packages/client/auth-core/src/services/CrossmintService.ts
```diff
@@ -17,36 +10,24 @@ export class CrossmintService {
...
-        } catch (error) {
-            console.error("Error fetching Crossmint API: ", error);
-            throw new Error(`Error fetching Crossmint API: ${error}`);
-        }
+        this.crossmintBaseUrl =
+            getEnvironmentForKey(apiKey) === "production"
+                ? this.getUrlFromEnv("production")
+                : this.getUrlFromEnv("staging");
+
+        this.issuer =
```
guiwoda, line 27: with my PR, this should always equal the `crossmintBaseUrl`

packages/client/auth-core/src/utils/jwt.ts
```diff
@@ -1,25 +1,59 @@
...
+export async function verifyCrossmintSessionToken(apiKey: string, token: string) {
+    const crossmintService = new CrossmintService(apiKey, token);
     try {
-        response = await crossmintService.fetchCrossmintAPI({
-            endpoint: "/sdk/auth/me",
-            options: {
-                method: "POST",
-            },
-        });
+        return await verifyJWTWithPublicKey(crossmintService, "cached");
```
guiwoda, line 9: this reads weird. Are you handling cache misses by catching an Error?

packages/client/auth-core/src/utils/tokenAuth/publicKey.ts
```diff
@@ -0,0 +1,40 @@
...
+        return publicKey;
+    }
+
+    return getFetchedPublicKey(token, jwksUri);
+}
+
+async function getSigningKey(jwksUri: string, kid?: string) {
+    const client = new JwksClient({
+        jwksUri,
+        cache: true,
```
guiwoda, line 31: If jwks-rsa is using cache, why do we need our own in-memory cache?

packages/client/auth-core/src/services/CrossmintService.ts
MatiNavCrossmint: hmm now that you mentioned, is that really what we want? I believe this should have two values, the crossmint or the STG one. And when developing locally or using the demo, we could use the staging one. 

If not, using the crossmintBaseUrl could lead to tokens with random issuers when developing locally, which I don't see the gain on that

packages/client/auth-core/src/utils/jwt.ts
MatiNavCrossmint: hmm Im not following will ask you on DM about this

packages/client/auth-core/src/utils/tokenAuth/publicKey.ts
MatiNavCrossmint: Now that you mentioned, reading the JWKsClient doc I found this

```
Caching
In order to prevent a call to be made each time a signing key needs to be retrieved you can also configure a cache as follows. If a signing key matching the kid is found, this will be cached and the next time this kid is requested the signing key will be served from the cache instead of calling back to the JWKS endpoint.
```

And now I understand the kid purpose. So agree we shouldn't handle the cache, but we should make the kid dynamic.

I assume the rotation of keys will come with the refresh proposal, so I'll remove the caching strategy from here

packages/client/auth-core/src/services/CrossmintService.ts
guiwoda: The point of the issuer is that the signer of the JWT matches the place where you use it.
The only place where this would break is if you take a local JWT and try to use it in STG / PROD or any other env.

guiwoda: Leaving the comment here for visibility:
We'll have 3 issuers going forward:
- https://www.crossmint.com/ PROD
- https://staging.crossmint.com/ STG
- https://dev.crossmint.com/ LOCAL - PREVIEW

The SDK may not validate issuer anyway, as the different environments will never share keys, so public key validation should suffice.

MatiNavCrossmint: Perfect!

packages/client/auth-core/src/utils/jwt.ts
```diff
@@ -1,25 +1,52 @@
...
+            (err.message.includes("invalid signature") || err.message.includes("invalid algorithm"))
+        ) {
+            throw new Error(err.message);
+        }
+
+        throw new Error("Invalid token");
+    }
+}
+
+async function verifyJWTWithPublicKey(crossmintService: CrossmintService) {
```
guiwoda, line 44: Same here, receiving the `token: string` instead of the `CrossmintService` makes this a simpler method and doesn't get involved in how / where we store the token.

guiwoda (approved):
LGTM! Left some design comments

PR #689, opened 2024-08-20
use Alchemy RPC for publicClient

afeight:
this is recommended by ZeroDev

## Description

We got blocked for some simple RPC calls that led to wallets failing to create
This is ZeroDev's recommendation

## Test plan

Haven't yet

### DIFF
--- a/packages/client/wallets/smart-wallet/src/blockchain/chains.ts
+++ b/packages/client/wallets/smart-wallet/src/blockchain/chains.ts
@@ -61,6 +61,6 @@ export const viemNetworks: Record<SmartWalletChain, Chain> = {
     "arbitrum-sepolia": arbitrumSepolia,
 };
 
-export const getBundlerRPC = (chain: SmartWalletChain) => {
+export function getBundlerRPC(chain: SmartWalletChain) {
     return BUNDLER_RPC + zerodevProjects[chain];
-};
+}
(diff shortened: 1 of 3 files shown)


packages/client/wallets/smart-wallet/src/blockchain/chains.ts
```diff
@@ -61,6 +61,6 @@ export const viemNetworks: Record<SmartWalletChain, Chain> = {
     "arbitrum-sepolia": arbitrumSepolia,
 };
 
-export const getBundlerRPC = (chain: SmartWalletChain) => {
+export function getBundlerRPC(chain: SmartWalletChain) {
```
afeight, line 64: let's be consistent–if we use `function` at all, let's make it so all functions use it rather than `const + arrow`

packages/client/wallets/smart-wallet/src/blockchain/rpc.ts
```diff
@@ -0,0 +1,18 @@
+import { SmartWalletChain } from "./chains";
+
+const ALCHEMY_API_KEY = "-7M6vRDBDknwvMxnqah_jbcieWg0qad9";
```
afeight, line 3: I hate this, but I don't think we can get around it trivially

packages/client/wallets/smart-wallet/src/blockchain/chains.ts
devlyn37: sgtm

packages/client/wallets/smart-wallet/src/blockchain/rpc.ts
devlyn37: Wonder if we can do project and ip specific rate limits here

devlyn37 (approved):
Nice, I went through the minting flow and it's a bit faster now too.

PR #691, opened 2024-08-20
Auth: Add lax on 3p cookie

MatiNavCrossmint:
## Description

Auth: Add lax on 3p cookie

## Test plan

Tested locally


guiwoda (approved):

alfonso-paella:
Can you add in the test plan what tests did you run exactly?

MatiNavCrossmint:
> Can you add in the test plan what tests did you run exactly?

Sure, but basically doing a full auth flow and checking it works fine. It's worth highlighting that this is a change that is expected to not have an impact on the flow as the 3p cookie is used by the external domain only

@alfonso-paella

PR #695, opened 2024-08-21
Window: Fix issue about never returning on popup init

MatiNavCrossmint:
## Description

Window: Fix issue about never returning on popup init

## Test plan

Tested through the auth SDK flow

### DIFF
--- a/packages/client/window/src/windows/Popup.ts
+++ b/packages/client/window/src/windows/Popup.ts
@@ -51,7 +51,7 @@ function createPopup(url: string, options: PopupWindowOptions) {
         return _window;
     }
 
-    return new Promise((resolve, reject) => {
+    return new Promise<Window>((resolve, reject) => {
         _window.onload = () => resolve(_window);
         _window.onerror = () => reject("Failed to load popup window");
     });
--- a/packages/client/window/src/windows/Popup.ts
+++ b/packages/client/window/src/windows/Popup.ts
@@ -37,7 +37,7 @@ export class PopupWindow<IncomingEvents extends EventMap, OutgoingEvents extends
     }
 }
 
-async function createPopup(url: string, options: PopupWindowOptions): Promise<Window> {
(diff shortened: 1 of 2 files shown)


packages/client/window/src/windows/Popup.ts
```diff
@@ -46,6 +47,10 @@ async function createPopup(url: string, options: PopupWindowOptions): Promise<Wi
         throw new Error("Failed to open popup window");
     }
 
+    if (options.awaitToLoad === false) {
```
guiwoda, line 50: I'm not sure about this, but have you tried moving the `const _window` assignment inside the Promise?
This looks like a race condition: the _window's `onload` event might be firing before the `Promise` gets the chance  to hook the Promise resolution.

MatiNavCrossmint: Hmm I'm 99.9% sure that isn't the case, the reason is that the whole code here is sync and the window loading is an async action. While there might be a remote case that an async action takes less than sync ones (that's why I'm not 100% sure), this doesn't apply IMO

guiwoda: I'm aware of that, otherwise I wouldn't talk about race conditions.
I think we are both saying the same thing! I'm assuming that `_window` is not a promise, but a sync object that fires off async operations in the background. Is there any way to initialize the `_window` with the hooks as parameters? That way you make sure they always get called.

guiwoda: I also don't understand how is typescript allowing you to `return _window` when that's clearly not a `Promise<Window>`. Unless `Window extends Promise`???

afeight (approved):
we try it and cross fingers 🤞

packages/client/window/src/windows/Popup.ts
MatiNavCrossmint: Resolving this to allow merging

PR #700, opened 2024-08-23
Smart Wallets + Auth: Use Crossmint Provider (Resolves WAL-2737)

devlyn37:
## Description

This PR:
- Adds required usage of the `CrossmintProvider` as defined within the [PRD for Smart Wallets + Auth.](https://docs.google.com/document/d/1HXvDK3tM64aSfUB6QUp1Ed59kEA1lxxM59j1B4l6HAI/edit#heading=h.5rk2bhp8fiz4)
- Treats the `jwt` within the context of `CrossmintProvider` as the single source of truth.
    - The `client-sdk-auth-core` package, no longer has jwt state, instead it sets it within the `CrossmintProvider` context
    - The `CrossmintAuthProvider` now uses jwt from the `CrossmintProvider` state to call `getOrCreateWallet`.

Motivations here:
- Keep everything a client SDK might need to operate in `CrossmintProvider` context, including a JWT which can either come from an external provider, or our auth product.
- Avoid sync issues w/ state in `CrossmintProvider` and  `useAuthCore`.

I've included changes in the Smart Wallet + Auth demo to help w/ review, will remove them after this PR is approved.

## Test plan

Locally:
- Confirmed that when using `CrossmintAuthProvider` outside of a `CrossmintProvider`, a nice error is thrown. ✅
- Through the demo flow, confirmed auth functionality works as expected:
    - Login  ✅
    - Logout ✅
    - Using returned wallet ✅


packages/client/auth-core/src/contexts/authlogin.tsx
```diff
@@ -4,46 +4,44 @@ import AuthModal from "@/components/AuthModal";
...
 
-    useEffect(() => {
-        const crossmintSession = document.cookie.split("; ").find((row) => row.startsWith("crossmint-session"));
-        const crossmintSessionCookie = crossmintSession ? crossmintSession.split("=")[1] : null;
-        if (crossmintSessionCookie) {
-            setJwtToken(crossmintSessionCookie);
-        }
-    }, []);
+    const crossmintService = useMemo(
+        () => CrossmintServiceFactory.create(crossmint.apiKey, crossmint.jwt),
```
devlyn37, line 39: This needs to be replaced with the new base service added by Bryan!

packages/client/ui/react-ui/src/hooks/useCrossmint.tsx
```diff
@@ -45,10 +46,10 @@ export function CrossmintProvider({
     return <CrossmintContext.Provider value={value}>{children}</CrossmintContext.Provider>;
 }
 
-export function useCrossmint() {
+export function useCrossmint(missingContextMessage?: string) {
```
devlyn37, line 49: So that when this is used within a specific package, we can throw a package specific error. Example: `'CrossmintMintingProvider' must be used within the base 'CrossmintProvider', see xxx docs`

packages/client/auth-core/src/contexts/authlogin.tsx
```diff
@@ -4,46 +4,44 @@ import AuthModal from "@/components/AuthModal";
...
+    }
+    const crossmintSession = document.cookie.split("; ").find((row) => row.startsWith("crossmint-session"));
+    return crossmintSession ? crossmintSession.split("=")[1] : undefined;
+}
+
 type AuthContextType = {
     login: () => void;
     logout: () => void;
-    jwt: string | null;
+    jwt: string | undefined;
```
devlyn37, line 20: To align with types defined in `CrossmintProvider`

packages/client/ui/react-ui/src/hooks/useCrossmint.tsx
```diff
@@ -16,14 +17,17 @@ export function CrossmintProvider({
...
-        new Proxy<Crossmint>(createCrossmint(createCrossmintParams), {
-            set(target, prop, value) {
-                if (prop === "jwt" && target.jwt !== value) {
-                    setVersion((v) => v + 1);
-                }
-                return Reflect.set(target, prop, value);
-            },
-        })
+        new Proxy<Crossmint>(
+            createCrossmint({ ...createCrossmintParams, jwt: createCrossmintParams.jwt ?? getCachedJwt() }),
```
devlyn37, line 21: So that the JWT is included in the first render

devlyn37: If a client using external auth passes in a jwt, that will be used. Otherwise, we we check if there's an existing session on the browser for crossmint auth, otherwise jwt is undefined initially.

packages/client/auth-core/src/contexts/authlogin.tsx
```diff
@@ -4,46 +4,44 @@ import AuthModal from "@/components/AuthModal";
 import { CrossmintServiceFactory } from "@/services/CrossmintService";
 import { type ReactNode, createContext, useContext, useEffect, useMemo, useState } from "react";
 
+import { Crossmint } from "@crossmint/common-sdk-base";
+
+export function getCachedJwt(): string | undefined {
+    if (typeof document === "undefined") {
+        return undefined; // Check if we're on the client-side
```
alfonso-paella, line 11: To consider: should we drop a log here if so? Or even consider throwing an error if we're certain thats omeone calling this on the backend is screwing up?

packages/client/auth-core/src/contexts/authlogin.tsx
```diff
@@ -4,46 +4,44 @@ import AuthModal from "@/components/AuthModal";
...
+    }
+    const crossmintSession = document.cookie.split("; ").find((row) => row.startsWith("crossmint-session"));
```
alfonso-paella, line 13: is `crossmint-session` a const we should extract out so it's easier to refactor later?

devlyn37: fixed

packages/client/auth-core/src/contexts/authlogin.tsx
devlyn37: Leaving for now, tracking [here](https://linear.app/crossmint/issue/WAL-2742/investigate-and-handle-devs-accidentally-running-on-server)

jmderby (approved):

PR #705, opened 2024-08-24
Smart Wallets + Auth: `CrossmintWalletProvider` and `useWallet` hook (Resolves WAL-2656)

devlyn37:
## Description

This PR:
- Implements the `useWallet` hook as defined within the [smart wallets + auth PRD](https://docs.google.com/document/d/1HXvDK3tM64aSfUB6QUp1Ed59kEA1lxxM59j1B4l6HAI/edit#heading=h.5rk2bhp8fiz4)
- Adds `CrossmintWalletProvider`, which is used internally by `CrossmintAuthProvider` but may be used standalone for clients using external auth.
- Updates the smart wallet + auth demo to use the new useWallet hook.

Motivations:
- Give devs a way to handle smart wallet errors.
- Provide easy to use loading states.
- Expose a React sdk/provider for devs using wallets but not auth. 

## Test plan

Locally:
- Went through the smart wallet + auth demo flow:
    - Logged into an account without an existing wallet, went through the creation flow, then minted.
    - Logged into an account with an existing wallet, then minted.
Everything worked as expected.
    
Starting next week we'll begin to add e2e test coverage and improve testing more broadly. The product is still evolving pretty rapidly, so not investing too heavily into testing yet.


packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -0,0 +1,83 @@
...
+    const stack = error instanceof Error ? error.stack : undefined;
+    return { status: "loading-error", error: new SmartWalletError(`Unknown Wallet Error: ${message}`, stack) };
+}
+
+type WalletContext = {
+    status: WalletStatus;
+    wallet?: EVMSmartWallet;
+    error?: SmartWalletError;
+    getOrCreateWallet: () => Promise<void>;
+};
```
devlyn37, line 35: Exposing this rather than the type intersection defined in the PRD, in practice working with the type intersection sucked. Other libraries (wagmi, tanstack query, nextauth, etc), return similar data w/ their hooks and don't use intersections, likely for similar reasons.

We'll continue to use the type intersection internally so ts enforces only valid states though.

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -0,0 +1,83 @@
...
+            getOrCreateWallet();
+            return;
+        }
+
+        if (state.status === "loaded" && crossmint.jwt == null) {
+            console.log("Clearing wallet");
+            setState({ status: "not-loaded" });
+            return;
+        }
+    }, [crossmint.jwt, config.createOnLogin]);
```
devlyn37, line 80: was reading about best practices, som say to include everything in the effect inside the deps array, but seems wasteful. We know stuff in this effect really runs if the jwt changes.

cc @jmderby wdty?

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -0,0 +1,83 @@
...
+    const getOrCreateWallet = async () => {
+        if (!crossmint.jwt) {
+            console.error("JWT is not available");
+            return;
+        }
+
+        if (state.status === "in-progress" || state.status === "loaded") {
+            console.error("Wallet is already loaded or being created");
+            return;
+        }
```
devlyn37, line 56: Need these extra checks since we expose this function to devs and they can call whenever they want.

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -0,0 +1,83 @@
+import { useCrossmint } from "@/hooks";
+import { ReactNode, createContext, useEffect, useMemo, useState } from "react";
+
+import { EVMSmartWallet, SmartWalletError, SmartWalletSDK } from "@crossmint/client-sdk-smart-wallet";
+import { Crossmint } from "@crossmint/common-sdk-base";
+
+export type CrossmintWalletConfig = {
+    type: "evm-smart-wallet";
+    defaultChain: "polygon-amoy" | "base-sepolia" | "optimism-sepolia" | "arbitrum-sepolia";
```
devlyn37, line 9: we still gotta use the full chain list here XD, will be a follow up.

devlyn37: These are all the chains available on staging atm, so it's fine for now

apps/wallets/smart-wallet/next/src/components/header.tsx
```diff
@@ -6,101 +6,104 @@ import { Copy, Image as ImageIcon, User, WalletMinimal } from "lucide-react";
...
 
-export const Header = () => {
-    const { jwt, logout, wallet } = useAuth();
+function formatWalletAddress(address: string, startLength: number, endLength: number): string {
+    return `${address.substring(0, startLength)}...${address.substring(address.length - endLength)}`;
+}
+
+export const Header: React.FC = () => {
+    const { logout } = useAuth();
+    const { wallet, status: walletStatus } = useWallet();
```
devlyn37, line 22: I rename status every time I use it, maybe that's a good sign to just rename status => walletStatus

cc @jmderby what do other libs normally do, wdyt?

jmderby: Are we using `status` from `useAuth` anywhere within the demo? If not, I think it's acceptable to keep the name status. My thinking is that in the uncommon case where useAuth and useWallet are being used in a component and the status of each is being used, it would be reasonable for the developer to rename the status at that point. Additionally, in scenarios where the wallet is loading and we're not relying on any status from the useAuth hook, using status seems appropriate since the page is primarily concerned with the wallet's status.

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
jmderby: I think it's to prevent unwanted side effects within useEffects. I used to work with really strict eslint rules that would yell at me anytime i didn't exhaust my dependencies. It won't hurt performance to include them all afaik.

packages/client/ui/react-ui/src/hooks/useWallet.ts
```diff
@@ -0,0 +1,12 @@
...
+import { useContext } from "react";
+
+export function useWallet() {
+    const walletContext = useContext(WalletContext);
+
+    if (!walletContext) {
+        throw new Error("useWallet must be used within CrossmintAuthProvider or CrossmintWalletProvider");
+    }
+
+    return { ...walletContext };
```
jmderby, line 11: nit can just `return walletContext`

devlyn37: lol ty for catching

apps/wallets/smart-wallet/next/src/components/header.tsx
devlyn37: ok sgtm

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
devlyn37: Ok, fixed

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -0,0 +1,82 @@
...
+};
+
+export const WalletContext = createContext<WalletContext>({
+    status: "not-loaded",
+    getOrCreateWallet: async () => {},
+});
+
+export function CrossmintWalletProvider({ children, config }: { config: CrossmintWalletConfig; children: ReactNode }) {
+    const { crossmint } = useCrossmint("CrossmintWalletProvider must be used within CrossmintProvider");
+    const [state, setState] = useState<ValidWalletState>({ status: "not-loaded" });
```
jmderby, line 49: i'm wondering if we should default this to "in-progress" so the initial render will be loading

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -1,66 +1,21 @@
 import { useCrossmint } from "@/hooks";
-import { ReactNode, createContext, useEffect, useMemo, useState } from "react";
+import { ReactNode } from "react";
```
jmderby, line 2: nit; could be type

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -1,66 +1,21 @@
...
-import { Crossmint } from "@crossmint/common-sdk-base";
+import { AuthProvider as AuthCoreProvider } from "@crossmint/client-sdk-auth-core/client";
 
-type AuthWalletProviderParams = {
-    embeddedWallets: {
-        type: "evm-smart-wallet";
-        defaultChain: "polygon-amoy" | "base-sepolia";
-        createOnLogin: "all-users" | "off";
-    };
+import { CrossmintWalletConfig, CrossmintWalletProvider } from "./CrossmintWalletProvider";
```
jmderby, line 6: same with CrossmintWalletConfig

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -0,0 +1,82 @@
+import { useCrossmint } from "@/hooks";
+import { kMaxLength } from "buffer";
```
jmderby, line 2: can remove - not used

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { ReactNode, createContext, useEffect, useMemo, useState } from "react";
+
+import { EVMSmartWallet, SmartWalletError, SmartWalletSDK } from "@crossmint/client-sdk-smart-wallet";
```
jmderby, line 5: EVMSmartWallet is type as well

jmderby (approved):

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
devlyn37: discussed in person

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
devlyn37: done

PR #707, opened 2024-08-24
React UI Lib: Convert Testing Suite To Vitest

devlyn37:
## Description

Jest (or at least how we've configured it) often struggles to build things to test, and when it fails does not provide useful error messages. This has caused me to get waste hours multiple times, bit the bullet here and converted the package to use vitest, which is much better and provides error messages that let you actually fix any build issues that come up!

Newer packages all use vitest, this change pushes the repo further in that direction.

This PR also converts a few imports that look like this `@/x/y` into `../../x/y` to fix some build issues (that vitest reported, maybe there's some config we could set to enable, but less drama to just have the regular imports instead.)

## Test plan

Existing tests run and pass.

### DIFF
--- a/packages/client/ui/react-ui/src/components/CrossmintCollectionView.test.tsx
+++ b/packages/client/ui/react-ui/src/components/CrossmintCollectionView.test.tsx
@@ -1,34 +1,40 @@
-import "@testing-library/jest-dom";
-import { render, screen } from "@testing-library/react";
+import { cleanup, render, screen } from "@testing-library/react";
+import { beforeEach, describe, expect, test } from "vitest";
 
 import { CrossmintNFTCollectionView } from "./CrossmintNFTCollectionView";
 
 const wallets = [{ chain: "solana", publicKey: "12345" }];
 
-describe("when only passing mandatory fields", () => {
-    test("should add them to the iframe query params", () => {
-        render(<CrossmintNFTCollectionView wallets={wallets} />);
-        const iframe = screen.getByRole("nft-collection-view");
-        const src = iframe.getAttribute("src");
-        expect(src).toContain("wallets=%5B%7B%22chain%22%3A%22solana%22%2C%22publicKey%22%3A%2212345%22%7D%5D");
-        expect(src).toContain("clientVersion=");
+describe("CrossmintNFTCollectionView", () => {
+    beforeEach(() => {
+        cleanup();
     });
-});
 
-describe("when not setting any environment", () => {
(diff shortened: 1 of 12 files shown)


packages/client/ui/react-ui/vitest.config.ts
```diff
@@ -0,0 +1,9 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+    test: {
+        environment: "jsdom",
```
jmderby, line 5: Jest automatically injects a globally `afterEach` into the testing env. The reason we have to add `cleanup()` in each test is because Vitest isn't configured this way by default. After searching online, I found this github issue that explains how we can add the same behaviour by default without needing to call cleanup every test. I'll also link the api docs around this config option:

Fix: 
https://github.com/testing-library/vue-testing-library/issues/296#issue-1577197825

Vitest config option:
https://vitest.dev/config/#globals

packages/client/ui/react-ui/src/components/CrossmintCollectionView.test.tsx
```diff
@@ -1,34 +1,40 @@
...
-describe("when only passing mandatory fields", () => {
-    test("should add them to the iframe query params", () => {
-        render(<CrossmintNFTCollectionView wallets={wallets} />);
-        const iframe = screen.getByRole("nft-collection-view");
-        const src = iframe.getAttribute("src");
-        expect(src).toContain("wallets=%5B%7B%22chain%22%3A%22solana%22%2C%22publicKey%22%3A%2212345%22%7D%5D");
-        expect(src).toContain("clientVersion=");
+describe("CrossmintNFTCollectionView", () => {
+    beforeEach(() => {
+        cleanup();
```
jmderby, line 10: see comment below (not marking as a blocker just a suggestion and maybe even reduce these calls)

jmderby (approved):
approving w/ comment

packages/client/ui/react-ui/src/components/CrossmintCollectionView.test.tsx
devlyn37: fixed

packages/client/ui/react-ui/vitest.config.ts
devlyn37: nice, ty for figuring this out!

devlyn37: I added to the config and removed all the calls to cleanup

PR #708, opened 2024-08-26
docs: add line to fix links during generation

dmulvi:
## Description

problem:
- typedoc generates markdown style links that lack the `./` preceding them
- this confuses mintlify and the links open in a new tab.

solution:
this adds code to the custom plugin file for typedoc that searches for all occurrences of `](` not followed by http or `../` (to avoid breaking correct external links and relative path) and replaces it with `](./`

## Test plan

- tested locally by running `npx typedoc` and examining the generated docs directory
- previously 304 occurrences matching regex: `\]\((?!http|\.{2}\/)` and now zero (when adding negative look ahead for `./` also


afeight (approved):
cool!

PR #715, opened 2024-08-28
Smart Wallets: Fix Error Loop Issue

devlyn37:
## Description

This PR fixes an issue where, in the event of an error in `SmartWalletSdk.getOrCreateWallet`, we'd call it again in repeat. @jmderby, who'll be reviewing this PR, and I are in the same room and we've shared more context in person.

## Test plan

On the Smart Wallet + auth  Demo:
- Happy path, login and mint flow.
- Confirmed we're no longer repeating requests in the event of an error.
- Confirmed that the `getOrCreateWallet` function we expose from `useWallet` can be used to successfully recover from the error state.

### DIFF
--- a/.changeset/three-pots-rescue.md
+++ b/.changeset/three-pots-rescue.md
@@ -1,0 +1,5 @@
+---
+"@crossmint/client-sdk-react-ui": patch
+---
+
+fix error handling issue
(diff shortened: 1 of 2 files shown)


jmderby (approved):
collab in-person, lgt(merge)

PR #719, opened 2024-08-28
VC: nft metadata

vmc-crossmint:
## Description
Returning also the nfts metadata when returning credentials nfts

### DIFF
--- a/packages/client/verifiable-credentials/src/verifiableCredentialsSDK/types/nft.ts
+++ b/packages/client/verifiable-credentials/src/verifiableCredentialsSDK/types/nft.ts
@@ -4,6 +4,7 @@ export interface Nft {
     chain: VCChain;
     contractAddress: string;
     tokenId: string;
+    metadata?: any;
 }
 
 export interface NftWithMetadata extends Nft {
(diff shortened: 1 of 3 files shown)


packages/client/verifiable-credentials/src/verifiableCredentialsSDK/types/nft.ts
```diff
@@ -4,6 +4,7 @@ export interface Nft {
     chain: VCChain;
     contractAddress: string;
     tokenId: string;
+    metadata?: any;
```
alberto-crossmint, line 7: Can we type this a bit mre specifically? Looking at the code in the NFTProjection type in crossmint-main this is the type we're returning:

```
export interface NFTMetadata {
    image: string;
    collection?: NFTMetadataCollection;
    name?: string;
    description?: string;
    attributes?: NFTMetadataAttribute[];
    animation_url?: string;
    external_url?: string;
    uri?: string;
    properties?: NFTPropertiesAttribute;
    credentialId?: string;
}
```

alberto-crossmint (approved):

packages/client/verifiable-credentials/src/verifiableCredentialsSDK/types/nft.ts
vmc-crossmint: I dont think makes much sense because most of that stuff is specific to our implementation and inner working, this interface should work for any nft metadata

PR #725, opened 2024-08-31
Auth: Use `BaseCrossmintService`

bryan-eastwood3:
## Description

Use BaseCrossmintService for auth, delete a bunch of old code

It's worth mentioning that this shouldn't change any functionality. The only API call inside of this service is to our jwks.json file which should never throw an error, so the enhanced error handling in this class won't have any effect. 

## Test plan

Tested signing into the smart wallets nextjs app according to @devlyn37's instructions


packages/client/auth-core/src/utils/jwt.ts
```diff
@@ -1,20 +1,20 @@
-import { CrossmintServiceFactory, type CrossmintServiceWithToken } from "@/services/CrossmintService";
+import { CrossmintAuthService } from "@/services/CrossmintAuthService";
 import { JsonWebTokenError, TokenExpiredError, verify } from "jsonwebtoken";
 
 import { getPublicKey } from "./tokenAuth/publicKey";
 
 export async function verifyCrossmintSessionToken(apiKey: string, token: string) {
-    const crossmintService = CrossmintServiceFactory.create(apiKey, token);
+    const crossmintService = new CrossmintAuthService(apiKey);
```
bryan-eastwood3, line 7: Since the JWT token is mutable I don't think it makes sense for it to be a constructor parameter

guiwoda (approved):

PR #727, opened 2024-09-03
Fix hydration error and client side error from auth provider

jmderby:
## Description

tldr; we were receiving a error when using the auth provider in a nextjs app and required a janky workaround for rendering it client side only. This PR fixes that and also cleans up the demo by removing the janky workaround. 
Additional details can be found in this linear issue: https://linear.app/crossmint/issue/WAL-2742/investigate-and-handle-devs-accidentally-running-on-server

This PR identifies a potential issue with Next.js, specifically regarding rendering problems when using context providers in server/client components in conjunction with Tailwind CSS. We have reported this bug on the Next.js GitHub board for further investigation and tracking.
Link to issue: https://github.com/vercel/next.js/issues/69682

## Test plan

Bug is fixed and regular auth login flow still works

## Package updates

- @crossmint/client-sdk-smart-wallet
- @crossmint/client-sdk-react-ui
- @crossmint/client-sdk-auth-core

Caveat: I've decided to set our react versions for both react-ui and auth-core to the same of version ^18.3.0. The carat will update to latest patch version but not exceed version 18.3


apps/wallets/smart-wallet/next/src/app/_lib/providers.tsx
```diff
@@ -1,9 +1,45 @@
...
...814 15.3173 25.2625 16.5588 28.7803C17.8004 32.298 18.4812 34.394 19.6928 35.2083C20.9043 36.0226 23.3073 35.6089 24.0716 34.7812' stroke='%23602C1B' stroke-width='2.5' stroke-linecap='round' stroke-linejoin='round'%3E%3C/path%3E%3Cpath d='M16.5491 23.792C15.7326 22.5723 14.0598 20.439 12.7716 16.9352C11.8009 14.2953 9.73159 13.4674 8.07616 15.5367C6.42074 17.606 8.01611 21.0801 9.49463 23.792C10.9732 26.5039 13.6633 30.2286 14.698 31.2632C15.3187 34.781 14.5083 37.4718 15.9545 39.9509C18.0087 43.472 22.1474 43.8858 26.079 42.8512C28.7691 42.4374 31.0453 44.0928 34.77 43.8858C38.4947 43.6789 39.7363 39.7473 41.3917 37.4711C43.0471 35.1949 45.9441 32.5048 47.1857 31.0563C48.4272 29.6078 48.0134 27.5385 46.5649 26.9178C45.1164 26.297 43.8582 26.5151 42.2194 28.1593C40.6806 29.7032 39.3224 31.884 39.3224 31.884C39.3224 31.884 36.8393 31.4702 34.5631 32.2979C32.0052 33.228 30.0106 34.3672 28.3552 36.8503' stroke='%23602C1B' stroke-width='2.5' stroke-linecap='round' stroke-linejoin='round'%3E%3C/path%3E%3C/g%3E%3C/g%3E%3Cdefs%3E%3CclipPath id='clip0_943_5959'%3E%3Crect width='54' height='54' fill='white' transform='translate(0.660156)'%3E%3C/rect%3E%3C/clipPath%3E%3C/defs%3E%3C/svg%3E",
```
devlyn37, line 23: appLogo and appName should be dropped here!

packages/client/auth-core/src/contexts/authlogin.tsx
```diff
@@ -78,18 +77,23 @@ export function AuthProvider({ children, crossmint, setJwtToken, appearance }: A
         }
     }, [crossmint.jwt]);
 
+    const value = useMemo(() => ({ login, logout, jwt: crossmint.jwt }), [login, logout]);
```
devlyn37, line 80: Should crossmint be a dep here?

packages/client/auth-core/src/contexts/authlogin.tsx
```diff
@@ -78,18 +77,23 @@ export function AuthProvider({ children, crossmint, setJwtToken, appearance }: A
...
-                <AuthModal
-                    baseUrl={crossmintService.crossmintBaseUrl}
-                    setModalOpen={setModalOpen}
-                    setJwtToken={setJwtToken}
-                    apiKey={crossmint.apiKey}
-                    appearance={appearance}
-                />
-            )}
+            {modalOpen
+                ? createPortal(
```
devlyn37, line 86: mind dropping some context on what a portal is and why we need it?

packages/client/wallets/smart-wallet/src/SmartWalletSDK.ts
```diff
@@ -72,6 +68,10 @@ export class SmartWalletSDK {
         chain: SmartWalletChain,
         walletParams: WalletParams = { signer: { type: "PASSKEY" } }
     ): Promise<EVMSmartWallet> {
+        if (!isClient()) {
```
devlyn37, line 71: ty for fixing!

apps/wallets/smart-wallet/next/src/app/_lib/providers.tsx
jmderby: good catch, until the main repo PR is up on staging, the auth modal will look like so
is that fine for now? 
Screenshot 2024-09-03 at 4 24 38 PM

packages/client/auth-core/src/contexts/authlogin.tsx
jmderby: Good catch, added!

packages/client/auth-core/src/contexts/authlogin.tsx
jmderby, line 86: By using a react portal, we can avoid Next.js hydration errors. By doing this, it ensures that the modal is handled separately from the rest of the app's content. This prevents problems that happen when the HTML generated on the server doesn't match what React expects on the client, which caused our hydration errors. The portal lets the modal live in its own place in the DOM, keeping everything in sync between the server and client.

in my opinion, we really should have used a portal from the get go

jmderby: done lmk if it covers enough

apps/wallets/smart-wallet/next/src/app/_lib/providers.tsx
devlyn37: yeah

packages/client/auth-core/src/contexts/authlogin.tsx
devlyn37: forget to commit the change here?

devlyn37: Also, why `useMemo` here? This doesn't seem v/ resource intensive operation.

apps/wallets/smart-wallet/next/src/app/template.tsx
```diff
@@ -2,61 +2,13 @@
...
-    const [isMounted, setIsMounted] = useState(false);
-    useEffect(() => {
-        setIsMounted(true);
-    }, []);
-
-    // This is a workaround to prevent a hydration error when using the CrossmintAuthProvider.
-    // In theory, this auth provider should be added to the providers.tsx file, but due to that file running initially server-side, it
-    // causes a hydration error when the auth modal is opened for whatever reason.
-    // Checking for mount here also prevents the auth provider from throwing a "Smart Wallet SDK should only be used client side." error.
-    return !isMounted ? (
```
devlyn37, line 24: beautiful and much needed, ty

packages/client/auth-core/src/contexts/authlogin.tsx
jmderby: removed

PR #728, opened 2024-09-03
Smart Wallet: Configure Signer via `CrossmintWalletProvider` ( Resolves WAL-2836 )

devlyn37:
## Description

This PR adds the ability to configure the signer in both `CrossmintWalletProvider`.

### Context

Currently `CrossmintWalletProvider` requires the `createOnInit` prop, which:
- when set to `all-users`, will cause the provider, when the jwt is set in `CrossmintProvider`, to begin creating a wallet automatically.
- when set to `off`, relies on the developer to call the `getOrCreateWallet` function exposed by the provider via the `useWallet` hook.

### SDK Surface Changes for `CrossmintWalletProvider`

**Removed `createOnInit` prop**

With `createOnInit`, wallet signer config needs to be passed in as a prop. There we're too many cases to think about here, and many did not have obvious behavior we should implement. I simplified the SDK surface here by only allowing wallet creation through the `getOrCreateWallet` method exposed through the `useWallet` hook.

**Added optional parameter `config` to `getOrCreateWallet` exposed from the `useWallet` hook**

Calling `getOrCreateWallet()` will continue to create a wallet with a passkey based signer, but now calling `getOrCreateWallet({type: "evm-smart-wallet", signer: {...}}).` is an option. This allows external signers, such as privy embedded wallets, to be used.

**Added** exposed a `clearWallet` function via the `useWallet` hook.

Removing `createOnInit` required moving code to react to changes in the jwt from `CrossmintProvider` from `CreateWalletProvider` into `CrossmintAuthProvider`. This required the `CrossmintWalletProvider` to expose some way to clear it's wallet state. Considering it was useful for integrating with our authentication, I suspect it'll be useful for devs integrating with external authentication as well.

### How to specify the signer using `CrossmintAuthProvider`

`createOnLogin` must be `false`, and the developer needs to call `getOrCreateWallet` themselves. This enables:
- A simple wallet creation flow by default (with passkeys, and the wallet created on login), comparable with competing products.
- flexibility for more complicated signer flows. (ex. using an external wallet, that won't be defined initially, etc).

## Test plan

Locally: 
- Tested with privy embedded wallets as the signer [demonstrated here](https://www.loom.com/share/dc419a8e0b77477bb18dba2d95d369a5) (with latest code changes).
- Confirmed the Auth + Smart Demo still works as expected

Unit:
- Added extensive unit tests for both `CrossmintAuthProvider` and `CrossmintWalletProvider`, which make up most of the line count for this PR.

## Package updates

Patch `@crossmint/client-sdk-react-ui`: "Add option to specify signer in CrossmintAuthProvider and CrossmintWalletProvider"


packages/client/ui/react-ui/src/hooks/useWallet.ts
```diff
@@ -1,6 +1,7 @@
-import { WalletContext } from "@/providers/CrossmintWalletProvider";
 import { useContext } from "react";
 
+import { WalletContext } from "../providers/CrossmintWalletProvider";
```
devlyn37, line 3: this change was to appease a vitest build thing

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -1,23 +1,61 @@
...
+            getOrCreateWallet({
+                type: embeddedWallets.type,
+                ...(embeddedWallets.config ?? { signer: { type: "PASSKEY" } }),
+            });
+        }
+
+        if (status === "loaded" && crossmint.jwt == null) {
+            clearWallet();
+        }
+    }, [crossmint.jwt, status]);
```
devlyn37, line 58: Logic here moved over from `CrossmintWalletProvider`

packages/client/ui/react-ui/src/testUtils.ts
```diff
@@ -0,0 +1,5 @@
+export const MOCK_API_KEY = "sk_development_12341234";
+export const waitForSettledState = async (callback: () => void) => {
```
devlyn37, line 2: This feels really naive to me, but I didn't see anything better in the docs and claude couldn't think of anything better either.

jmderby: We'll look later if necessary to make testing the status more robust, but good for now I think

jmderby (approved):
discussed PR in person - nice work!

PR #730, opened 2024-09-03
Smart Wallets + Auth: Enable All Smart Wallet Chains For `CrossmintAuthProvider` & `CrossmintWalletProvider`

devlyn37:
## Description

We temporarily only had staging chains enabled for these providers, this PR updates the SDK to enable the use of all Smart Wallet supported chains.

## Test plan

Sanity checked that the auth + smart wallet demo flow still works.

## Package updates

This PR patches both `@crossmint/client-sdk-smart-wallet` and `@crossmint/client-sdk-react-ui`

### DIFF
--- a/packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
+++ b/packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
@@ -1,11 +1,16 @@
 import { useCrossmint } from "@/hooks";
 import { ReactNode, createContext, useEffect, useMemo, useState } from "react";
 
-import { EVMSmartWallet, SmartWalletChain, SmartWalletError, SmartWalletSDK } from "@crossmint/client-sdk-smart-wallet";
+import {
+    EVMSmartWallet,
+    EVMSmartWalletChain,
+    SmartWalletError,
+    SmartWalletSDK,
+} from "@crossmint/client-sdk-smart-wallet";
 
 export type CrossmintWalletConfig = {
     type: "evm-smart-wallet";
-    defaultChain: SmartWalletChain;
+    defaultChain: EVMSmartWalletChain;
     createOnLogin: "all-users" | "off";
 };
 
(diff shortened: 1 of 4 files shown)


jmderby (approved):
lgtm

packages/client/wallets/smart-wallet/src/index.ts
```diff
@@ -11,6 +11,8 @@ export type {
     WalletParams,
 } from "./types/params";
 
+export type { SmartWalletChain } from "./blockchain/chains";
```
alfonso-paella, line 14: I wonder if this type should be EVMSmartWalletChain. I think in the future we may benefit from disallowing one to pass "type": "solana-smart-wallet", "defaultChain": "polygon-amoy"

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -1,11 +1,11 @@
 import { useCrossmint } from "@/hooks";
 import { ReactNode, createContext, useEffect, useMemo, useState } from "react";
 
-import { EVMSmartWallet, SmartWalletError, SmartWalletSDK } from "@crossmint/client-sdk-smart-wallet";
+import { EVMSmartWallet, SmartWalletChain, SmartWalletError, SmartWalletSDK } from "@crossmint/client-sdk-smart-wallet";
 
 export type CrossmintWalletConfig = {
     type: "evm-smart-wallet";
-    defaultChain: "polygon-amoy" | "base-sepolia" | "optimism-sepolia" | "arbitrum-sepolia";
+    defaultChain: SmartWalletChain;
```
alfonso-paella, line 8: Do we have a good error if you pass a chain that's not compatible with the environment your API key is set to?

packages/client/wallets/smart-wallet/src/index.ts
devlyn37: makes sense, renamed.

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
devlyn37: No, mind if I add that as a follow up?

devlyn37: will do so unless you object!

PR #731, opened 2024-09-03
Smart Wallet: Chain <> Env Validation

devlyn37:
## Description

This PR adds chain <> env validation, so if a dev provides a staging API key and a mainnet chain (or vice versa), they'll get an error message that looks like this:

Screenshot 2024-09-03 at 2 52 51 PM

## Test plan

Locally, within the smart wallet SDK playground:
- Confirmed the error case works as expected, as pictured above, for both staging & production keys.
- Sanity checked that the happy path still works.

## Package updates

patch `@crossmint/client-sdk-smart-wallet`

### DIFF
--- a/packages/client/wallets/smart-wallet/src/SmartWalletSDK.ts
+++ b/packages/client/wallets/smart-wallet/src/SmartWalletSDK.ts
@@ -1,9 +1,15 @@
 import { stringify } from "viem";
 
-import { validateAPIKey } from "@crossmint/common-sdk-base";
+import { APIKeyEnvironmentPrefix, validateAPIKey } from "@crossmint/common-sdk-base";
 
 import { CrossmintWalletService } from "./api/CrossmintWalletService";
-import type { SmartWalletChain } from "./blockchain/chains";
+import {
+    SMART_WALLET_MAINNETS,
+    SMART_WALLET_TESTNETS,
+    type SmartWalletChain,
+    isMainnetChain,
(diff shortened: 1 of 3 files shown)


packages/client/wallets/smart-wallet/src/SmartWalletSDK.ts
```diff
@@ -83,4 +92,23 @@ export class SmartWalletSDK {
             }
         });
     }
+
+    private assertValidChain(chain: SmartWalletChain) {
+        if (!this.validChain(chain)) {
+            const validChains = this.crossmintEnv === "production" ? SMART_WALLET_MAINNETS : SMART_WALLET_TESTNETS;
+            const formattedChains = validChains.map((chain) => `"${chain}"`).join(", ");
+            throw new SmartWalletError(
+                `Invalid chain "${chain}" for environment "${this.crossmintEnv}" specified by API key.\n` +
```
alfonso-paella, line 101: `The selected chain "${chain}" is not available in "${this.crossmintEnv}". Either set a valid chain or check you're using an API key for the environment you're trying to target.`

alfonso-paella (approved):

packages/client/wallets/smart-wallet/src/SmartWalletSDK.ts
devlyn37: updated

PR #751, opened 2024-09-11
Crossmint Provider: Hydration Fix (Resolves WAL-2880)

devlyn37:
## Description

Right now, within the `CrossmintProvider` component, we read from cookies for Crossmint Auth sessions to populate the initial JWT. Since Next.js generates a static page when possible, the first render, which is that static page, would not have a JWT value. The subsequent hydration would read in the cookies, and cause an error because the static page and first client render would be different, which is not allowed.

This PR moves reading the cookie to a `useEffect` that only happens once on render client side, fixing the hydration error.

## Test plan

There's pretty solid unit test coverage over `CrossmintProvider`, `CrossmintAuthProvider`, which should continue to pass here.

Locally, I ran through the demo flow and confirmed everything worked as expected, w/ no hydration issues.

### DIFF
--- a/packages/client/ui/react-ui/src/hooks/useCrossmint.tsx
+++ b/packages/client/ui/react-ui/src/hooks/useCrossmint.tsx
@@ -2,8 +2,6 @@ import { ReactNode, createContext, useCallback, useContext, useMemo, useRef, use
 
 import { Crossmint, createCrossmint } from "@crossmint/common-sdk-base";
 
-import { getCachedJwt } from "../utils";
-
 export interface CrossmintContext {
     crossmint: Crossmint;
     setJwt: (jwt: string | undefined) => void;
@@ -18,17 +16,14 @@ export function CrossmintProvider({
     const [version, setVersion] = useState(0);
 
     const crossmintRef = useRef<Crossmint>(
-        new Proxy<Crossmint>(
-            createCrossmint({ ...createCrossmintParams, jwt: createCrossmintParams.jwt ?? getCachedJwt() }),
-            {
-                set(target, prop, value) {
-                    if (prop === "jwt" && target.jwt !== value) {
-                        setVersion((v) => v + 1);
-                    }
-                    return Reflect.set(target, prop, value);
-                },
-            }
-        )
+        new Proxy<Crossmint>(createCrossmint(createCrossmintParams), {
+            set(target, prop, value) {
+                if (prop === "jwt" && target.jwt !== value) {
+                    setVersion((v) => v + 1);
+                }
+                return Reflect.set(target, prop, value);
+            },
+        })
     );
 
     const setJwt = useCallback((jwt: string | undefined) => {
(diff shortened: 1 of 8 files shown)


packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.test.tsx
```diff
@@ -88,9 +87,15 @@ describe("CrossmintAuthProvider", () => {
...
         };
+
+        Object.defineProperty(document, "cookie", {
+            writable: true,
+            value: "",
+        });
     });
 
     test("Happy path", async () => {
+        document.cookie = `${SESSION_PREFIX}=mock-jwt; path=/;SameSite=Lax;`;
```
devlyn37, line 98: Moved to mocking cookies to get closer to real world behavior.

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -41,6 +42,13 @@ export function CrossmintAuthProvider({ embeddedWallets, children, appearance }:
     const crossmintBaseUrl = validateApiKeyAndGetCrossmintBaseUrl(crossmint.apiKey);
     const [modalOpen, setModalOpen] = useState(false);
 
+    useEffect(() => {
+        const session = sessionFromClient();
+        if (session != null) {
+            setJwt(session);
+        }
+    }, []);
```
devlyn37, line 50: This product specific logic belongs here, not in `CrossmintProvider`

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.test.tsx
```diff
@@ -61,11 +69,14 @@ describe("CrossmintWalletProvider", () => {
 
     beforeEach(() => {
         vi.resetAllMocks();
-
-        vi.mocked(createCrossmint).mockImplementation(() => ({
-            apiKey: MOCK_API_KEY,
-            jwt: "mock-jwt",
-        }));
+        vi.mocked(createCrossmint).mockImplementation(() => ({} as any));
+        vi.mocked(useCrossmint).mockReturnValue({
```
devlyn37, line 73: mock the hook instead to get closer to real world behavior.

packages/client/ui/react-ui/src/hooks/useCrossmint.tsx
```diff
@@ -18,17 +16,14 @@ export function CrossmintProvider({
...
-            }
-        )
+        new Proxy<Crossmint>(createCrossmint(createCrossmintParams), {
+            set(target, prop, value) {
+                if (prop === "jwt" && target.jwt !== value) {
+                    setVersion((v) => v + 1);
+                }
+                return Reflect.set(target, prop, value);
+            },
+        })
```
devlyn37, line 26: @jmderby Leaving everything as is here, since the unit tests defines a requirement that we can't hit w/o. The updating jwt from SDK piece.

packages/client/ui/react-ui/src/hooks/useCrossmint.tsx
```diff
@@ -18,17 +16,14 @@ export function CrossmintProvider({
...
-            {
-                set(target, prop, value) {
-                    if (prop === "jwt" && target.jwt !== value) {
-                        setVersion((v) => v + 1);
-                    }
-                    return Reflect.set(target, prop, value);
-                },
-            }
-        )
+        new Proxy<Crossmint>(createCrossmint(createCrossmintParams), {
```
jmderby, line 19: jwt could be removed from createCrossmint() but is a non-blocker

jmderby (approved):
Discussed in person the changes, tests pass and hydration error is no longer. Lets ship

PR #752, opened 2024-09-11
Crossmint Provider: Drop `jwt` prop (Resolves WAL-2884)

devlyn37:
## Description

atm we let developers set jwt from the props, or via setJwt, but there's no clear way they can co-exist. For example, what if a developer uses the jwt prop, but then calls setJwt? Do the jwt prop and actual jwt differ in value? etc.

This PR removes the `jwt` prop. I considered removing `setJwt` instead, but that's not possible atm since `CrossmintAuthProvider` relies on it.

## Test plan

Existing unit tests.

Locally, I went through the auth + smart wallet demo flow just in case XD.

## Package updates

patch `@crossmint/client-sdk-react-ui`

### DIFF
--- a/packages/client/ui/react-ui/src/hooks/useCrossmint.tsx
+++ b/packages/client/ui/react-ui/src/hooks/useCrossmint.tsx
@@ -11,12 +11,15 @@ const CrossmintContext = createContext<CrossmintContext | null>(null);
 
 export function CrossmintProvider({
     children,
-    ...createCrossmintParams
-}: { children: ReactNode } & Parameters<typeof createCrossmint>[0]) {
+    apiKey,
+    overrideBaseUrl,
+}: Omit<Crossmint, "jwt"> & {
+    children: ReactNode;
+}) {
     const [version, setVersion] = useState(0);
 
     const crossmintRef = useRef<Crossmint>(
-        new Proxy<Crossmint>(createCrossmint(createCrossmintParams), {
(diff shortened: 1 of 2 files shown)


packages/client/ui/react-ui/src/hooks/useCrossmint.tsx
```diff
@@ -11,12 +11,15 @@ const CrossmintContext = createContext<CrossmintContext | null>(null);
...
+    apiKey,
+    overrideBaseUrl,
+}: Omit<Crossmint, "jwt"> & {
+    children: ReactNode;
+}) {
     const [version, setVersion] = useState(0);
 
     const crossmintRef = useRef<Crossmint>(
-        new Proxy<Crossmint>(createCrossmint(createCrossmintParams), {
+        new Proxy<Crossmint>(createCrossmint({ apiKey, overrideBaseUrl }), {
```
jmderby, line 22: it makes sense to keep `jwt` in the Crossmint type as the provider return type is Crossmint and returns a jwt, but shouldn't we also remove `jwt` from the `createCrossmint` helper fn?

jmderby: or maybe if other devs are using the function from common-sdk-base then maybe we should keep it 🤔

jmderby (approved):
approved with comments

PR #759, opened 2024-09-17
Smart Wallets: Rename `EOASigner` => `ExternalSigner`

devlyn37:
## Description

This PR renames `EOASigner` to `ExternalSigner` for a couple reasons:
- It's not accurate, `1193Provider` a valid `EOASigner` is a generic interface for wallets and can even represent AA wallets.
- It's confusing, if I wanted to use a privy, dynamic, or web3auth signer `EOASigner` is not obvious, but `ExternalSigner` is!

Ofc open to further naming improvements.

## Test plan

N/A

## Package updates

patch `@crossmint/client-sdk-smart-wallet` (XD will respect semver after the public announcement)


packages/client/wallets/smart-wallet/src/types/internal.ts
```diff
@@ -40,7 +40,7 @@ export interface PasskeyCreationContext extends WalletCreationContext {
 }
 
 export interface EOACreationContext extends WalletCreationContext {
```
devlyn37, line 42: Can rename this stuff later as needed, want to ship this API change quickly before Racino starts migrating and we do the public announcement.

nischayu95 (approved):
SGTM!

PR #764, opened 2024-09-18
@crossmint/client-sdk-react-ui: Twind (Tailwind-in-js)

jmderby:
## Description

Twind is a small, fast, and feature-rich CSS(tailwind)-in-JS library. It allows us to use Tailwind classes in our react-ui package.
tldr; It's a runtime CSS-in-JS solution with Tailwind syntax using a lightweight compiler 

https://twind.style/

## Test plan

I've added a `cn` utility funciton for ease of class merging, and replaced our auth modal iframe styles from inline to tailwind classes for the sake of testing it out to make sure it works. 
tldr; I've tested build mode, dev mode and the auth flow with default styles and custom styles and it all works as expected. I've also configured twind to use the default tailwind preset `presetTailwind` so that we get ootb tailwind presets.

## Package updates

### DIFF
--- a/packages/client/ui/react-ui/src/lib/utils.ts
+++ b/packages/client/ui/react-ui/src/lib/utils.ts
@@ -1,0 +1,12 @@
+import { type ClassValue, clsx } from "clsx";
+import { twMerge } from "tailwind-merge";
+
+/**
+ * Combines multiple class names into a single string.
+ *
+ * @param inputs - The class names to combine.
+ * @returns The combined class names as a string.
+ */
+export function cn(...inputs: ClassValue[]) {
+    return twMerge(clsx(inputs));
+}
--- a/.changeset/selfish-cows-fix.md
+++ b/.changeset/selfish-cows-fix.md
@@ -1,0 +1,5 @@
+---
+"@crossmint/client-sdk-react-ui": patch
+---
+
+Added Twind, a lightweight css-in-js tailwind compiler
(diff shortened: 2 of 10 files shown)


packages/client/ui/react-ui/src/lib/utils.ts
```diff
@@ -0,0 +1,12 @@
+import { type ClassValue, clsx } from "clsx";
+import { twMerge } from "tailwind-merge";
+
+/**
+ * Combines multiple class names into a single string.
+ *
+ * @param inputs - The class names to combine.
+ * @returns The combined class names as a string.
+ */
+export function cn(...inputs: ClassValue[]) {
```
mPaella, line 10: use full name `classNames` to match crossmint-main

packages/client/ui/react-ui/src/lib/utils.ts
```diff
@@ -0,0 +1,12 @@
+import { type ClassValue, clsx } from "clsx";
```
mPaella, line 1: i would make this path `/utils/classNames.ts`, a single "utils" file is an antipattern imo

mPaella (approved):

packages/client/ui/react-ui/src/lib/utils.ts
jmderby: fixed 👍

packages/client/ui/react-ui/src/lib/utils.ts
jmderby: good catch, fixed

PR #765, opened 2024-09-18
Vitest config: allow imports that use @ alias

jmderby:
## Description

When files get auto imported we use the @ alias and vitest didn't like that and wanted relative imports instead. This pr adds @ alias to the vitest config to allow that. 

## Test plan

Importing with relative or @ alias works.

## Package updates

null

### DIFF
--- a/.changeset/gorgeous-dancers-sing.md
+++ b/.changeset/gorgeous-dancers-sing.md
@@ -1,5 +1,0 @@
----
-"@crossmint/client-sdk-react-ui": patch
----
-
-Added alias to the vitest config
--- a/.changeset/gorgeous-dancers-sing.md
+++ b/.changeset/gorgeous-dancers-sing.md
@@ -1,0 +1,5 @@
+---
+"@crossmint/client-sdk-react-ui": patch
+---
+
+Added alias to the vitest config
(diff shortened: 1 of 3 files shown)


.changeset/gorgeous-dancers-sing.md
devlyn37, line 1: pls remove

packages/client/ui/react-ui/src/consts/version.ts
devlyn37, line 1: Can't wait to delete these lol

devlyn37 (approved):

packages/client/ui/react-ui/src/consts/version.ts
jmderby: me too

packages/client/ui/react-ui/vitest.config.ts
```diff
@@ -6,5 +7,8 @@ export default defineConfig({
         include: ["**/*.test.{ts,tsx}"],
         exclude: ["node_modules"],
         globals: true,
+        // This is needed because we are using the @ symbol to import from the src folder.
+        // Otherwise, Vitest will yell at us.
+        alias: [{ find: "@", replacement: resolve(__dirname, "./src") }],
```
mPaella, line 12: "@/", no?

PR #766, opened 2024-09-18
Auth: Makes sure refresh token is received from Auth frame

AlbertoElias:
## Description

We were creating the refresh token server side, but weren't using it in the client. This is the first step by receiving it from the auth frame.

This also stores the refresh token as a cookie with the correct expiry closing https://linear.app/crossmint/issue/WAL-2905/%5Brefresh-token%5D-store-refresh-token-in-the-client

And it uses the refresh token to call `/refresh` on page load and before jwt expires to always have up to date session token: https://linear.app/crossmint/issue/WAL-2931/%5Brefresh-token%5D-use-refresh-token-to-fetch-jwt

## Test plan

Tested manually in combo with this [pr](https://github.com/Paella-Labs/crossbit-main/pull/14734) and JWT is still fetched correctly and now we also receive the refresh token

## Package updates

`@crossmint/client-sdk-react-ui@1.3.20` 
`@crossmint/common-sdk-base@1.2.7`


packages/client/ui/react-ui/src/hooks/useCrossmint.tsx
```diff
@@ -35,14 +39,21 @@ export function CrossmintProvider({
         }
     }, []);
 
+    const setRefreshToken = useCallback((refreshToken: string | undefined) => {
```
MatiNavCrossmint, line 42: Although I see us already using useCallback throughout this file, so this comment should apply to all cases or to none, I see it as an overhead using useCallback here. 

The purpose of useCallback is to prevent a function being created on each render, making it more performant. However, I don't see that criteria applying for a function that performs a simple if check.

But this is just something that took my attention, it's not a comment to change anything

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -149,3 +170,8 @@ function sessionFromClient(): string | undefined {
     const crossmintSession = document.cookie.split("; ").find((row) => row.startsWith(SESSION_PREFIX));
     return crossmintSession ? crossmintSession.split("=")[1] : undefined;
 }
+
+function refreshTokenFromClient(): string | undefined {
```
MatiNavCrossmint, line 174: I would change this and the upper method to `getRefreshTokenFromCookie` and `getSessionTokenFromCookie`

MatiNavCrossmint (commented):
Amazing PR, I left just some small comments 👍

packages/client/ui/react-ui/src/hooks/useCrossmint.tsx
AlbertoElias: Yeah, you're right. I can remove in this file at least

AlbertoElias:
> Amazing PR, I left just some small comments 👍

Thanks! Replied to your comments!

packages/client/ui/react-ui/src/hooks/useCrossmint.tsx
```diff
@@ -33,17 +33,17 @@ export function CrossmintProvider({
         })
     );
 
-    const setJwt = (jwt: string | undefined) => {
+    const setJwt = useCallback((jwt: string | undefined) => {
```
AlbertoElias, line 36: @MatiNavCrossmint had to add them back because when `useCrossmint` is re-rendered, `setJWT` would change and wherever `setJWT` is used as a dependency would be re-executed.

packages/client/auth-core/server/index.ts
```diff
@@ -1 +0,0 @@
-export { verifyCrossmintSessionToken } from "../src/utils";
```
MatiNavCrossmint, line 1: I know we will focus on the client for this iterations, but I don't get why we remove this functions, unless there is already exported somehow in another place, or there's something I'm not seeing

packages/client/ui/react-ui/src/hooks/useRefreshToken.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export function useRefreshToken({ crossmintAuthService, setAuthMaterial }: UseAuthTokenRefreshProps) {
+    const timeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+
+    const refreshAuthMaterial = useCallback(() => {
+        const refreshToken = getCookie(REFRESH_TOKEN_PREFIX);
+        if (refreshToken != null) {
+            crossmintAuthService
+                .refreshAuthMaterial(refreshToken)
+                .then((result) => {
```
MatiNavCrossmint, line 31: I'm curious about using then catch instead of async await. I prefer us using async/await to keep consistency with the codebase, although there may be a reason behind using it with this way

packages/client/auth-core/server/index.ts
AlbertoElias: This was only re-exporting a util in the library for server code, but I was attempting to share the code for both instances and re-configured the exports so this would work in the same way without breaking anything. But now I noticed we'll need to keep them separate as `jsonwebtoken` doesn't work browser-side so I'll have to rethink this part

packages/client/ui/react-ui/src/hooks/useRefreshToken.ts
AlbertoElias: My bad, I was doing it like that in a `useEffect` and forgot to change to async/await upon refactoring

MatiNavCrossmint (approved):
LGTM!

PR #768, opened 2024-09-18
Remove auto generating version scripts

jmderby:
## Description

You know those annoying files that get auto generated and appear in your version control? well this PR allows you to never have to worry about those again XD 

## Test plan

Ensure this doesn't break anything.

### DIFF
--- a/packages/client/ui/vanilla-ui/src/consts/version.ts
+++ b/packages/client/ui/vanilla-ui/src/consts/version.ts
@@ -1,1 +1,3 @@
-export const LIB_VERSION = "1.1.15";
+import { version } from "../../package.json";
+
+export const LIB_VERSION = version;
--- a/packages/client/ui/react-ui/package.json
+++ b/packages/client/ui/react-ui/package.json
@@ -19,12 +19,9 @@
         "LICENSE"
     ],
     "scripts": {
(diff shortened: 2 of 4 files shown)


packages/client/ui/vanilla-ui/src/consts/version.ts
```diff
@@ -1 +1,3 @@
-export const LIB_VERSION = "1.1.15";
+import { version } from "../../package.json";
```
devlyn37, line 1: Doing this includes package.json in the client bundle, which exposes the dependancies we use, and is problematic if we ship a package version w/ a vulnerable dep.

We may need to do something at build time instead here!

jmderby: > While the package.json content might be included in the source maps, for an open-source project this doesn't significantly alter the security landscape since the information is already public. However, if you're concerned about the source maps, you can configure your build process to exclude them from the production bundle or to generate source maps without including content from package.json.

devlyn37: True, and since we include the version in the bundle it's not like the attacker couldn't trace back the version to the source code on npm or internet archives etc.

devlyn37 (approved):

PR #776, opened 2024-09-20
SDK Internal DX: Prune Pre-Commit Hook & Unused Dev Deps

devlyn37:
## Description

This PR deletes our pre-commit hooks, I've noticed it causes trouble (not working w/ different node installations, weird failure states myself and other eng have gotten into). We already run the lint & format check in CI, plus have the workspace setup to format on save, which should be more than enough.

I've already been operating w/o the pre-commit hook locally for the past few months since I started working on the SDK and haven't had any issues.

## Test plan

n/a

## Package updates

n/a

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,0 @@
-pnpm lint-staged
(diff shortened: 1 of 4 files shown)


jmderby (approved):
lgtm ✅

PR #779, opened 2024-09-24
Passkey Prompts: Custom Appearance prop (Resolves WAL-2922)

jmderby:
## Description

This PR adds `appearance` prop (exactly the same as our auth modal). It makes necessary changes in order to facilitate the styles from that prop while also leaving default styles applied when `appearance` prop is vacant.

## Test plan

Manual tested login flow, minting flow, and ensured styles work both applied and default as well as setting `showWalletModals: false` and ensuring the modals do NOT apply. 

## Package updates

n/a


AlbertoElias (approved):
LGTM

aigustin:
lgtm!

Small note, should we have the defaults in one place? Right now they seem all over the place

jmderby:
> lgtm!
> 
> Small note, should we have the defaults in one place? Right now they seem all over the place

Great point! We currently don't have a theme setup in this package but wouldn't be a bad idea to make a ticket for this. It's the same with the auth modal currently. Having a theme would be nice for @crossmint/client-sdk-react-ui

PR #785, opened 2024-09-25
Auth: Implements one time use refresh token to fetch auth material

AlbertoElias:
## Description

Makes SDK listen to initial one-time refresh token to then fetch all auth material.

Also removes unused popup/iframe events. Please test locally to double check it's working.

## Test plan

Delete cookies => Log in with both OTP and OAuth

## Package updates

@crossmint/client-sdk-react-ui v1.5.0

### DIFF
--- a/.changeset/brown-rice-push.md
+++ b/.changeset/brown-rice-push.md
@@ -1,0 +1,5 @@
+---
+"@crossmint/client-sdk-react-ui": minor
+---
+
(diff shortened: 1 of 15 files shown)


MatiNavCrossmint (approved):
LGTM! 

Once merged, we should also check that the flow works when run on an incognito window, as we are asked to sign in our google accounts which was an edge case on the communication flow

PR #786, opened 2024-09-25
Auth: Makes sure refreshAuthMaterial is only ran once at a time

AlbertoElias:
## Description

Currently, React Strict Mode triggers hooks twice, which ends up calling /refresh twice, and the second time fails logging users out. This forces that it only runs once at a time, even in dev mode.

## Test plan

Run smart wallets demo app and check in networks tab in inspector that /refresh is only triggered once.

## Package updates

@crossmint/client-sdk-react-ui patch

### DIFF
--- a/packages/client/ui/react-ui/src/hooks/useRefreshToken.ts
+++ b/packages/client/ui/react-ui/src/hooks/useRefreshToken.ts
@@ -23,31 +23,41 @@ type UseAuthTokenRefreshProps = {
     logout: () => void;
 };
 
+let refreshPromise: Promise<void> | null = null;
+
 export function useRefreshToken({ crossmintAuthService, setAuthMaterial, logout }: UseAuthTokenRefreshProps) {
     const refreshTaskRef = useRef<CancellableTask | null>(null);
 
-    const refreshAuthMaterial = useCallback(async () => {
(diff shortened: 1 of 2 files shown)


packages/client/ui/react-ui/src/hooks/useRefreshToken.ts
jmderby, line 1: I'm not sure how that works. The promise doesn't ever resolve, is this more of a hack to have the function only ever execute once? 🤔

AlbertoElias: It's to make sure that everything inside the async IIFE has finished running before it can be called again. The actual promise just holds that IIFE until it has finished running and it's then set to null. Open to suggestions to cleaner ways to hack around this too

jmderby: I can't think of any better way, so if this fixes the issue i'm ok with it.

jmderby: It'd be nice if that description you wrote was a comment inside that function though

jmderby (approved):
approving with suggestion to add explanation block comment inside function, and fix CI lint

PR #788, opened 2024-09-26
infra: make most pkgs tree shakable

mPaella:
## Description
making most of the packages tree shakable

## Test plan

u can see the bundle sizes of our apps has been decreased (new is on the right)
image

tested payments demo nextjs app

### DIFF
--- a/packages/common/base/package.json
+++ b/packages/common/base/package.json
@@ -15,7 +15,7 @@
     "types": "./dist/index.d.ts",
     "files": ["dist", "src", "LICENSE"],
     "scripts": {
-        "build": "tsup --minify",
+        "build": "tsup",
         "dev": "tsup --watch",
         "test": "vitest run"
     },
--- a/packages/common/base/package.json
+++ b/packages/common/base/package.json
@@ -15,8 +15,8 @@
     "types": "./dist/index.d.ts",
     "files": ["dist", "src", "LICENSE"],
     "scripts": {
-        "build": "tsup src/index.ts --clean --format esm,cjs --outDir ./dist --minify --dts --sourcemap",
-        "dev": "tsup src/index.ts --clean --format esm,cjs --outDir ./dist --dts --sourcemap --watch",
+        "build": "tsup --minify",
+        "dev": "tsup --watch",
         "test": "vitest run"
     },
     "dependencies": {
--- a/tsup.config.base.ts
+++ b/tsup.config.base.ts
@@ -11,4 +11,5 @@ export const treeShakableConfig: Options = {
     outDir: "dist",
     entry: ["src/**/*.(ts|tsx)", "!src/**/*.test.(ts|tsx)"],
     shims: true,
+    sourcemap: process.env.NODE_ENV !== "production",
 };
--- a/tsup.config.base.ts
+++ b/tsup.config.base.ts
@@ -1,0 +1,14 @@
+import type { Options } from "tsup";
+
+export const treeShakableConfig: Options = {
+    splitting: true,
+    clean: true,
+    dts: true,
+    format: ["esm", "cjs"],
+    bundle: true,
+    skipNodeModulesBundle: true,
+    watch: false,
+    outDir: "dist",
+    entry: ["src/**/*.(ts|tsx)", "!src/**/*.test.(ts|tsx)"],
+    shims: true,
+};
--- a/tsup.config.base.ts
+++ b/tsup.config.base.ts
@@ -11,5 +11,6 @@ export const treeShakableConfig: Options = {
     outDir: "dist",
     entry: ["src/**/*.(ts|tsx)", "!src/**/*.test.(ts|tsx)"],
     shims: true,
+    minify: process.env.NODE_ENV === "production",
     sourcemap: process.env.NODE_ENV !== "production",
 };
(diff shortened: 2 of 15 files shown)


AlbertoElias:
If the config is the same, would it be possible to have one in the root and import it from each package?

alfonso-paella:
For @AlbertoElias to consider during review -- is this test plan sufficient? Could tree-shaking fail in unexpected ways we must test? (e.g. run a couple demo sites or something)

mPaella:
> If the config is the same, would it be possible to have one in the root and import it from each package?

done

packages/common/base/package.json
```diff
@@ -15,8 +15,8 @@
     "types": "./dist/index.d.ts",
     "files": ["dist", "src", "LICENSE"],
     "scripts": {
-        "build": "tsup src/index.ts --clean --format esm,cjs --outDir ./dist --minify --dts --sourcemap",
-        "dev": "tsup src/index.ts --clean --format esm,cjs --outDir ./dist --dts --sourcemap --watch",
+        "build": "tsup --minify",
+        "dev": "tsup --watch",
```
AlbertoElias, line 19: should we keep generating sourcemaps in `dev` for ease of development?

tsup.config.base.ts
```diff
@@ -0,0 +1,14 @@
...
+    splitting: true,
+    clean: true,
+    dts: true,
+    format: ["esm", "cjs"],
+    bundle: true,
+    skipNodeModulesBundle: true,
+    watch: false,
+    outDir: "dist",
+    entry: ["src/**/*.(ts|tsx)", "!src/**/*.test.(ts|tsx)"],
+    shims: true,
```
AlbertoElias, line 13: have we agreed at crossmint to ship node.js env shims? Do we def need them or could we force us to use browser-only code and libraries to reduce bundle size?

AlbertoElias (commented):
Couple comments. Also, shouldn't we release new versions with these new builds?

AlbertoElias:
I'm testing it with other apps. So far, in Smarter Wallet, there are some styles that are not being built correctly so I'm looking into it.

jmderby:
> I'm testing it with other apps. So far, in Smarter Wallet, there are some styles that are not being built correctly so I'm looking into it.
> 
> @jmderby could you chime in with how `twind` works to create classes on the fly? The issue comes with classes that are set conditionally during execution such as:
> 
> ```js
> <iframe
>     ref={iframeRef}
>     src={iframeSrc}
>     onLoad={handleIframeLoaded}
>     title="Authentication Modal"
>     className={classNames(
>         "w-full h-[500px] border pt-12 pb-8",
>         appearance?.colors?.border
>             ? `border-[${appearance.colors.border}]`
>             : "border-[#D0D5DD]",
>         appearance?.borderRadius ? `rounded-[${appearance.borderRadius}]` : "rounded-2xl",
>         appearance?.colors?.background ? `bg-[${appearance.colors.background}]` : "bg-white"
>     )}
> />
> ```

Yeah, I might have to re-evaluate how we're handling the dynamic classes by using inline styles or something. It breaks tailwind rules 
https://tailwindcss.com/docs/content-configuration#dynamic-class-names:~:text=Don%E2%80%99t%20use%20props%20to%20build%20class%20names%20dynamically

mPaella:
yeah, classes like `border-[${appearance.colors.border}]` are known to not work/be flaky

better to use style object for those

jmderby:
@maxwellfortney @AlbertoElias do you guys want me to prioritize a PR that replaces dynam classes with inline?

mPaella:
not sure safelisting will even work, but i would just go for styles for simplicity + certainty it will work

>want me to prioritize a PR
assuming its currently broken, i think that makes sense

 in the context of this pr tho, i think we can say the conditional styles not working is not due to the tree shaking

AlbertoElias:
> in the context of this pr tho, i think we can say the conditional styles not working is not due to the tree shaking

Not due to the tree shaking, but something changes in the build process that breaks this while it does work in `main`. I'll finish testing other apps and report back

tsup.config.base.ts
mPaella: no agreement, i think the bundle size impact is pretty negligible tho and i'd rather not have to think about the compatibility

the docs are not very clear on what exactly it does https://tsup.egoist.dev/#inject-cjs-and-esm-shims

packages/common/base/package.json
mPaella: i can add them to `dev`, but we really need a proper solution - where during the build for release (during the ci/cd) we dont generate sourcemaps, but locally we do

i've already made this issue here to stop including sourcemaps in the prod deploy
https://linear.app/crossmint/issue/INF-625/sdk-repo-dont-publish-maps-to-final-dist

AlbertoElias: Agreed, but we should have them in dev, they're super useful for debugging in dev tools

mPaella: check [new commit](https://github.com/Crossmint/crossmint-sdk/pull/788/commits/7691a2b8a2c4ccf568f10cbd70423baecb15733f)

AlbertoElias:
Looking good. But we should wait to merge this until @jmderby fixes the twind stuff on `react-ui` or development and testing will be a pain

AlbertoElias (approved):

AlbertoElias:
@mPaella Implemented #805 so this can now be merged.

PR #791, opened 2024-10-01
Cleans up how refresh tokens are received to avoid multiple refreshes

AlbertoElias:
## Description

Due to how `iframeWindow` was being created and the fact that `React.StrictMode` runs all hooks twice, we were listening to the post message events on the iframe twice, and therefore calling `/refresh` twice with the same refresh secret which would sometime throw an error

## Test plan

Running the demo app in dev mode (`React.StrictMode = true`) and authenticating via email will now make a single call to `/refresh`

## Package updates

`@crossmint/client-sdk-react-ui`: patch

### DIFF
--- a/packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
+++ b/packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
@@ -1,5 +1,5 @@
 import { REFRESH_TOKEN_PREFIX, SESSION_PREFIX, deleteCookie, getCookie, setCookie } from "@/utils/authCookies";
-import { type ReactNode, createContext, useEffect, useState } from "react";
+import { type ReactNode, createContext, useCallback, useEffect, useState } from "react";
 import { createPortal } from "react-dom";
 
 import { CrossmintAuthService } from "@crossmint/client-sdk-auth-core/client";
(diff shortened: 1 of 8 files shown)


packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -97,11 +100,14 @@ export function CrossmintAuthProvider({ embeddedWallets, children, appearance }:
...
-        setAuthMaterial(authMaterial);
-        return authMaterial;
-    };
+    const fetchAuthMaterial = useCallback(
+        async (refreshToken: string): Promise<AuthMaterial> => {
+            const authMaterial = await crossmintAuthService.refreshAuthMaterial(refreshToken);
+            setAuthMaterial(authMaterial);
+            return authMaterial;
+        },
+        [crossmintAuthService, setAuthMaterial]
```
afeight, line 109: do we need these as dependencies? May be required by linter, but it's doesn't seem to be the case that if these vars change, we need to re-run the function

afeight: rather, is there a benefit of using `useCallback` here instead of what we previously had? 
Is it preventing some memory leaks to hold references to the service / `set` method?

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -47,12 +47,15 @@ export function CrossmintAuthProvider({ embeddedWallets, children, appearance }:
     const crossmintBaseUrl = validateApiKeyAndGetCrossmintBaseUrl(crossmint.apiKey);
     const [modalOpen, setModalOpen] = useState(false);
 
-    const setAuthMaterial = (authMaterial: AuthMaterial) => {
-        setCookie(SESSION_PREFIX, authMaterial.jwtToken);
-        setCookie(REFRESH_TOKEN_PREFIX, authMaterial.refreshToken.secret, authMaterial.refreshToken.expiresAt);
-        setJwt(authMaterial.jwtToken);
-        setRefreshToken(authMaterial.refreshToken.secret);
-    };
+    const setAuthMaterial = useCallback(
```
afeight, line 50: same here, more for my own info
React is a bit rusty to understand the benefit of using `useCallback` here. Mind explaining? (DM best so I see faster)

afeight (approved):
Overall it seems to check out, but my React knowledge is a bit at it's limit to 100% verify that I understand what all is happening here 😅

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
AlbertoElias: Yup, better to remove the deps here. Will do

PR #792, opened 2024-10-03
Auth: User Profile

MatiNavCrossmint:
## Description

Implement
```typescript
const {user, getUser} = useAuth()
```

## Test plan

Run signin flow wiht google and Farcaster


packages/client/auth-core/src/services/CrossmintAuthService.ts
```diff
@@ -20,6 +20,22 @@ export class CrossmintAuthService extends BaseCrossmintService {
         return {
             jwtToken: result.jwt,
             refreshToken: result.refresh,
+            user: result.user,
         };
     }
+
+    async getUserFromServer(externalUserId: string) {
+        const result = await this.fetchCrossmintAPI(
+            `sdk/auth/user/${externalUserId}`,
```
AlbertoElias, line 29: should do go in a versioned route directly?

packages/client/auth-core/src/services/CrossmintAuthService.ts
```diff
@@ -20,6 +20,22 @@ export class CrossmintAuthService extends BaseCrossmintService {
...
+            { method: "GET" },
+            "Error fetching user."
+        );
+
+        return result.user;
+    }
+    async getUserFromClient(jwt: string) {
+        const result = await this.fetchCrossmintAPI("sdk/auth/user", { method: "GET" }, "Error fetching user.", jwt);
```
AlbertoElias, line 37: how about `return await this...`

packages/common/base/src/types/Crossmint.ts
```diff
@@ -1,5 +1,22 @@
 import { type ValidateAPIKeyPrefixExpectations, validateAPIKey } from "@/apiKey";
 
+export type FarcasterMetadata = {
```
AlbertoElias, line 3: As these are Auth specific things, shouldn't they go in a separate file? Would also be good to have all Auth types in a single place shared between SDK and crossmint-nextjs

AlbertoElias (approved):
looks good, left some minor comments

packages/client/auth-core/src/services/CrossmintAuthService.ts
MatiNavCrossmint: It should go, but making the change will break the current user flow. So, we should duplicate the controller, and I prefer delaying that until needed. As this adds a new endpoint, it's not needed for this iteration

AlbertoElias: okay, though just for my knowledge, what unversioned endpoints are we calling from the sdks right now that would break?

PR #793, opened 2024-10-03
Auth modal: Responsive height

jmderby:
## Description

Now that we're adding more loginOptions, we need the modal to respond to the different heights. I was able to achieve this simply by using our existing strategy from embedded checkouts. This ticket relys on a PR from main repo

### Photos to showcase different permutations/combos

Screenshot 2024-10-03 at 4 05 15 PM
Screenshot 2024-10-03 at 4 06 56 PM
Screenshot 2024-10-03 at 4 07 16 PM
Screenshot 2024-10-03 at 4 08 58 PM
Screenshot 2024-10-03 at 4 09 31 PM

## Test plan

Tested through flow and everything still works, also tested different combinations.

## Package updates

@crossmint/client-sdk-react-ui

PR #794, opened 2024-10-03
Auth modal: Login Options + Responsive height

jmderby:
## Description

This PR adds the new `loginOptions` prop that allows us to pass different sign in buttons. 
(main repo PR: https://github.com/Paella-Labs/crossbit-main/pull/14945)

### Caveat of this PR
I needed to combine https://github.com/Crossmint/crossmint-sdk/pull/793 into this PR due to needing to dynamically calculate the initial modal height based on the `loginOptions`. Specifically Farcaster (i explain further below)

### Photos to showcase different permutations/combos

Screenshot 2024-10-03 at 4 06 56 PM
Screenshot 2024-10-03 at 4 07 16 PM
Screenshot 2024-10-03 at 4 08 58 PM
Screenshot 2024-10-03 at 4 09 31 PM

**Farcaster requires a large modal height**, so we're defaulting the minHeight of the authmodal to 500px when farcaster is in the `loginOptions` array. For now this quick and dirty sln should suffice. 

Screenshot 2024-10-09 at 11 16 00 AM
Screenshot 2024-10-09 at 11 16 06 AM

## Test plan

All flows work correctly without issues.

## Package updates


packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -17,10 +17,13 @@ export type CrossmintAuthWalletConfig = {
     showWalletModals?: boolean;
 };
 
+export type LoginMethod = "email" | "google" | "farcaster";
```
AlbertoElias, line 20: As SDKs are used in crossbit-main, how about having the type centralized in one place here in the sdks and the fetching that in crossbit-main?

jmderby: Good call, I can add it my main concern right now is if we wanted to add a auth method quickly from main repo we'd be bottlenecked by having to add the type in SDK. I'll make a ticket to follow this up later 👍

packages/client/ui/react-ui/src/components/auth/AuthModal.tsx
```diff
@@ -76,11 +82,24 @@ export default function AuthModal({
                 outgoingEvents: {},
             });
 
+            console.log({ initIframe });
```
AlbertoElias, line 91: leftover log?

packages/client/ui/react-ui/src/components/auth/AuthModal.tsx
```diff
@@ -56,6 +56,12 @@ export default function AuthModal({
     const iframePaddingTopPX = 48;
     const iframePaddingBottomPX = 32;
     const paddingOffset = iframePaddingTopPX + iframePaddingBottomPX;
+    let iFrameRenderMinHeightPX = 300;
```
AlbertoElias, line 59: for the sake of avoiding using `let`, how about a ternary?

AlbertoElias (approved):
Couple of nitpicks. I would also consider centering the farcaster button vertically as it looks awkward as is. Not important as only having Farcaster is an edge-case and it would also happen in the other repo.

PR #795, opened 2024-10-03
Rename showWalletModals to showPasskeyHelpers

jmderby:
## Description

We decided this naming was more clear.

we don't have documentation yet for this property so there's no need to update it anywhere.

## Test plan

na

## Package updates

BREAKING CHANGES 


AlbertoElias (approved):
Like this name more!

PR #802, opened 2024-10-09
Server SDK: Adds server-sdk and removes server-sdk-auth

AlbertoElias:
## Description

As per the decisions made for our server-sdk, we'll remove `server-sdk-auth` and use `server-sdk` instead. This creates a basic skeleton, keeping our auth features and it can be used like this:

```js
import { createCrossmint, CrossmintAuth } from "@crossmint/server-sdk";

        apiKey: process.env.SERVER_CROSSMINT_API_KEY || "",
});
const crossmintAuth = CrossmintAuth.from(crossmint);
```

## Test plan

Created an `ssr` example app.

## Package updates

@crossmint/server-sdk minor
@crossmint/server-sdk-next-starter patch


MatiNavCrossmint (approved):
LGTM!

PR #804, opened 2024-10-09
removes server-sdk-auth changeset

AlbertoElias:
## Description

We deleted the library before it publishes so don't need the change set.

## Test plan

None

## Package updates

None


MatiNavCrossmint (approved):
✅

PR #805, opened 2024-10-09
Removes dynamic tailwind classes as they break tree shaking build

AlbertoElias:
## Description

Unblocks #788 and we [shouldn't be using dynamic classes](https://tailwindcss.com/docs/content-configuration#:~:text=the%20corresponding%20CSS%3A-,Don%E2%80%99t%20construct%20class%20names%20dynamically,-%3Cdiv%20class) for tailwind anyway: 

## Test plan

Ran next demo with and without appearance object

## Package updates

`@crossmint/client-sdk-react-ui` patch

### DIFF
--- a/.changeset/long-bugs-remain.md
+++ b/.changeset/long-bugs-remain.md
@@ -1,0 +1,5 @@
+---
+"@crossmint/client-sdk-react-ui": patch
+---
+
+Removes dynamic tailwind classes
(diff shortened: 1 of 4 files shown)


jmderby:
@AlbertoElias we should also change the one spot in `PoweredByCrossmint.tsx`

jmderby (changes requested):
Looks good, just update the `PoweredByCrossmint.tsx` file and then were good to go 👍

jmderby (approved):
nice I tested everything, that it still works the same 👍

PR #808, opened 2024-10-10
Auth: add getUser on server sdk

MatiNavCrossmint:
## Description

Auth: add getUser on server sdk

## Test plan

- Created a new page on smarterWallets and use successfully the getUser function:
```typescript
import { CrossmintAuth, createCrossmint } from "@crossmint/server-sdk";

export default async function TestPage() {
        apiKey: "xxxxxxxx",
    });
    const auth = CrossmintAuth.from(crossmint);
    const user = await auth.getUser("76a4e894-0fe1-4a55-9250-7aab31373af1");
    return <div>{user.email}</div>;
}
```

## Package updates

`server-sdk: 0.2.0`

### DIFF
--- a/packages/server/src/auth/CrossmintAuth.ts
+++ b/packages/server/src/auth/CrossmintAuth.ts
@@ -1,7 +1,5 @@
 import { type Crossmint, CrossmintApiClient } from "@crossmint/common-sdk-base";
-import { SDK_NAME, SDK_VERSION } from "./utils/constants";
-
-const CROSSMINT_API_VERSION = "2024-09-26";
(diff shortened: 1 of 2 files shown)


packages/server/src/auth/CrossmintAuth.ts
```diff
@@ -1,6 +1,8 @@
 import { type Crossmint, CrossmintApiClient } from "@crossmint/common-sdk-base";
 import { SDK_NAME, SDK_VERSION } from "./utils/constants";
 
+const CROSSMINT_API_VERSION = "2024-09-26";
```
AlbertoElias, line 4: can you add this to the constants file?

AlbertoElias (approved):

PR #810, opened 2024-10-10
Auth SDK: Implements getSession in server sdk for ssr

AlbertoElias:
## Description

Implemented a getSession method in `CrossmintAuth` that will get the current session, and refresh it if necessary. Works with the Request object used by most node.js frameworks and also is flexible with other frameworks like next.js by allowing developers to fetch the tokens from the cookies themselves.

## Test plan

Added unit tests and also created a next.js example to make sure it works there.


packages/server/README.md
```diff
@@ -1 +1,51 @@
...
+
+To use the SDK in your application, follow these steps:
+
+1. Import the SDK into your project:
+
+```ts
+import { createCrossmint, CrossmintAuth } from "@crossmint/server-sdk";
+
+const crossmint = createCrossmint({
+    apiKey: process.env.SERVER_CROSSMINT_API_KEY || "",
```
afeight, line 23: nit: would probably go 
```suggestion
    apiKey: process.env.SERVER_CROSSMINT_API_KEY!,
```
so that it blows up if the value is missing rather than sending requests that always 403, which they might ask us for help about instead of debugging themselves 😅

packages/server/README.md
```diff
@@ -1 +1,51 @@
...
+});
+
+const crossmintAuth = CrossmintAuth.from(crossmint);
```
afeight, line 26: maybe this is in the doc, but is there a reason to prefer `CrossmintAuth.from(crossmint)` over `new CrossmintAuth(crossmint)`? Is this a secondary instantiation route?

packages/server/src/auth/utils/cookies.ts
```diff
@@ -0,0 +1,51 @@
...
+        cookieHeader = getCookieHeaderFromNodeRequest(request);
+    } else if (isFetchRequest(request)) {
+        cookieHeader = getCookieHeaderFromFetchRequest(request);
+    } else {
+        throw new CrossmintAuthenticationError("Unsupported request type");
+    }
+
+    const { [SESSION_PREFIX]: jwtToken, [REFRESH_TOKEN_PREFIX]: refreshToken } = parseCookieHeader(cookieHeader);
+
+    return { jwtToken, refreshToken };
```
afeight, line 23: when possible, I aim to keeping a method either fully declarative or fully imperative, here it's a bit of both where the first half w/ if-else is imperative, then we have a nice declarative `parseCookieHeader`
I'd be thinking:
```
const cookieHeader = getCookieHeader(request)
const { ... } = parseCookieHeader(...)
return { ... }
```

which has the added benefit of removing some repeated code between `getCookieHeaderFromNodeRequest` and `getCookieHeaderFromFetchRequest` because `getCookieHeader` would have to look like:
```
let cookieHeader;
if (isNodeRequest) {
  cookieHeader = ...
} else if (isFetchRequest) {
  cookieHeader = ...
} 

if (cookieHeader == null) {
  throw
}
return cookieHeader

File: packages/server/src/auth/utils/tokenAuth/publicKey.ts [hunk 4]
```diff
@@ -23,6 +23,7 @@ async function getSigningKey(jwksUri: string, kid?: string) {
         const signingKey = await client.getSigningKey(kid);
         return signingKey;
     } catch (error) {
+        console.error("Error fetching signing key", error);
```
afeight, line 26: 🙏

packages/server/tsconfig.json
```diff
@@ -6,7 +6,7 @@
         "paths": {
             "@/*": ["src/*"]
         },
-        "module": "commonjs"
+        "module": "ESNext"
```
afeight, line 9: what's the larger impact of this? Is it less compatible with servers running older versions? Or is the impact local?

packages/server/src/auth/CrossmintAuth.ts
```diff
@@ -20,4 +30,62 @@ export class CrossmintAuth {
     public static from(crossmint: Crossmint): CrossmintAuth {
         return new CrossmintAuth(crossmint);
     }
+
+    public getJWKSUri() {
+        return `${this.apiClient.baseUrl}/.well-known/jwks.json`;
+    }
```
afeight, line 36: is this something that needs to be public/expected to be used from a client? Or more meant as the helper for `verifyCrossmintJwtToken`?

packages/server/src/auth/CrossmintAuth.ts
```diff
@@ -20,4 +30,62 @@ export class CrossmintAuth {
...
+        return {
+            jwtToken: resultJson.jwt,
+            refreshToken: resultJson.refresh,
+            user: resultJson.user,
+        };
+    }
+
+    public async getSession(options: GenericRequest | AuthMaterialBasic): Promise<AuthSession> {
+        const { jwtToken, refreshToken } =
+            "refreshToken" in options ? options : getAuthCookies(options as GenericRequest);
```
afeight, line 58: I thought that TS would be able to infer that, if `"refreshToken"` is not there, then it is not of type `AuthMaterialBasic` so it must be `GenericRequest`. Does it still throw compiler errors if we remove the typecast?
```suggestion
            "refreshToken" in options ? options : getAuthCookies(options);
```

packages/server/src/auth/CrossmintAuth.ts
```diff
@@ -20,4 +30,62 @@ export class CrossmintAuth {
...
+    private async validateOrRefreshSession(jwtToken: string | undefined, refreshToken: string): Promise<AuthSession> {
+        if (jwtToken) {
+            try {
+                const decodedJwt = await this.verifyCrossmintJwtToken(jwtToken);
+                return {
+                    jwtToken,
+                    userId: decodedJwt.sub as string,
+                };
+            } catch (_) {
+                // The JWT is invalid, we need to refresh the session
```
afeight, line 81: thanks for the comment 🙂

packages/server/README.md
AlbertoElias: createCrossmint verifies the api key and throws an error if it's not valid. I think that error would be more descriptive. lmk what you

packages/server/src/auth/CrossmintAuth.ts
AlbertoElias: It's different levels of utilities to access it:

1. The actual endpoint
2. This function to easily get the endpoint
3. We then use it internally if they trust our implementation of the verification, but they can also verify themselves

packages/server/tsconfig.json
AlbertoElias: I'll create a separate build for commonjs. I want to ensure we have a good es module implementation for tree shaking

AlbertoElias: actually, this is unnecessary as we're using `tsup` and that generates both a .cjs and .js output so we're good.

apps/auth/ssr/src/app/page.tsx
```diff
@@ -1,22 +1,48 @@
...
-export default function Home() {
+export default async function Home() {
+    let isLoggedIn = false;
+    let userId: string | undefined;
+
     try {
         const crossmint = createCrossmint({
             apiKey: process.env.SERVER_CROSSMINT_API_KEY || "",
         });
         const crossmintAuth = CrossmintAuth.from(crossmint);
```
afeight, line 12: re: `.from` instantiation
It makes sense to have this as a static function because maybe one day we need to make it async that a normal constructor can't handle, so maybe we at least just make it `init` to match smart wallets
image

apps/auth/ssr/src/app/page.tsx
```diff
@@ -1,22 +1,48 @@
...
+                jwtToken: jwtCookie,
+                refreshToken: refreshCookie,
+            });
+            isLoggedIn = true;
+            userId = fetchedUserId;
+            console.log("jwt", jwtToken);
+            console.log("userId", userId);
+        } else {
+            console.log("user not logged in");
+        }
```
afeight, line 29: this is probably superfluous. As this will show up in client logs maybe we remove
```suggestion
        }
```

packages/common/auth/src/types/authmaterial.ts
```diff
@@ -1,10 +1,28 @@
 import type { SDKExternalUser } from "./user";
 
+export type AuthMaterialBasic = {
+    jwtToken?: string;
+    refreshToken: string;
+};
+
+interface refreshTokenObject {
```
afeight, line 8: two small things: 1. first letter uppercase, 2. "object" doesn't add meaning 
```suggestion
interface RefreshToken {
```

packages/common/auth/src/types/authmaterial.ts
```diff
@@ -1,10 +1,28 @@
...
+    expiresAt: string;
+}
+
 export type AuthMaterial = {
     jwtToken: string;
-    refreshToken: {
-        secret: string;
-        expiresAt: string;
-    };
+    refreshToken: refreshTokenObject;
```
afeight, line 15:
```suggestion
    refreshToken: RefreshToken;
```

packages/common/auth/src/types/authmaterial.ts
```diff
@@ -1,10 +1,28 @@
...
     user: SDKExternalUser;
 };
+
+export type AuthMaterialResponse = {
+    jwt: string;
+    refresh: refreshTokenObject;
```
afeight, line 21:
```suggestion
    refresh: RefreshToken;
```

packages/common/auth/src/types/authmaterial.ts
```diff
@@ -1,10 +1,28 @@
...
+    user: SDKExternalUser;
+};
+
+export type AuthSession = {
+    jwtToken: string;
```
afeight, line 26: `jwtToken` on the other hand I don't like haha. Isn't the `t` "token" anyway? I think just `jwt` here and other places for consistency
```suggestion
    jwt: string;
```

packages/common/auth/src/types/authmaterial.ts
```diff
@@ -1,10 +1,28 @@
 import type { SDKExternalUser } from "./user";
 
+export type AuthMaterialBasic = {
+    jwtToken?: string;
```
afeight, line 4:
```suggestion
    jwt?: string;
```

packages/common/auth/src/types/authmaterial.ts
AlbertoElias: This has to remain `refresh` because it's the schema used in the backend which we can't change without making a breaking change.

packages/server/src/auth/utils/cookies.ts
```diff
@@ -0,0 +1,55 @@
...
+    return cookieHeader;
+}
+
+function getCookieHeaderFromFetchRequest(request: Request): string {
+    const cookieHeader = request.headers.get("Cookie");
+    if (cookieHeader == null) {
+        throw new CrossmintAuthenticationError("Cookie header not found");
+    }
+    return cookieHeader;
+}
```
afeight, line 44: I think we've still got a little duplication here that's not needed
```suggestion
function getCookieHeader(request: GenericRequest): string {
    let cookieHeader: string;

    if (isNodeRequest(request)) {
        cookieHeader = request.headers.cookie;
    } else if (isFetchRequest(request)) {
        cookieHeader = request.headers.get("Cookie");
    } else {
        throw new CrossmintAuthenticationError("Unsupported request type");
    }

    if (cookieHeader == null) {
        throw new CrossmintAuthenticationError("Cookie header not found");
    }
    return cookieHeader;
}
```

packages/server/src/auth/utils/jwt.ts
```diff
@@ -1,12 +1,11 @@
-import { CrossmintAuthService } from "@crossmint/common-sdk-auth";
-import { JsonWebTokenError, TokenExpiredError, verify } from "jsonwebtoken";
+import jwt from "jsonwebtoken";
 
 import { getPublicKey } from "./tokenAuth/publicKey";
 
-export async function verifyCrossmintSessionToken(apiKey: string, token: string) {
-    const crossmintService = new CrossmintAuthService(apiKey);
+export async function verifyCrossmintJwtToken(token: string, jwksUri: string) {
```
afeight, line 5: nit: re `jwtToken`
```suggestion
export async function verifyCrossmintJWT(token: string, jwksUri: string) {
```

packages/server/src/auth/CrossmintAuth.ts
```diff
@@ -21,6 +31,63 @@ export class CrossmintAuth {
...
+            }
+        }
+
+        const refreshedAuthMaterial = await this.refreshAuthMaterial(refreshToken);
+        return {
+            jwt: refreshedAuthMaterial.jwt,
+            userId: refreshedAuthMaterial.user.id,
+        };
+    }
```
afeight, line 89: nit: keep public methods all first in the class, then private helpers. Should go:
1. `getSession` (public)
2. `getUser` (public)
3. `verifyCrossmintJWT` (public)
4. `getJWKSUri` (not sure yet)
5. `refreshAuthMaterial` (private)
6. `validateOrRefreshSession`

packages/server/src/auth/CrossmintAuth.ts
```diff
@@ -21,6 +31,63 @@ export class CrossmintAuth {
         return new CrossmintAuth(crossmint);
     }
 
+    public getJWKSUri() {
```
afeight, line 34: nit: Casing here is weird because of inconsistency. Either: `getJwksUri` or `getJWKSURI` plz

afeight (approved):
cool, feel free to ignore the nits, the only important Q is re: `getJwksUri` being public or private

apps/auth/ssr/src/app/page.tsx
AlbertoElias: we discussed that this was an old approach and we should go with `from`

PR #817, opened 2024-10-17
Auth Server SDK: Stores new auth material cookies

AlbertoElias:
## Description

Following the SSR doc, we needed a helper to store the cookies server-side. If a response is added to `getSession` this is done automatically. If not, we provide a `storeAuthMaterial` helper to do so.

## Test plan

Added tests and also tried in local node, express and remix apps.

## Package updates

react-ui: patch
common-auth: patch
server-sdk: minor


packages/server/src/auth/CrossmintAuth.ts
```diff
@@ -31,16 +32,28 @@ export class CrossmintAuth {
...
         }
 
         try {
-            return await this.validateOrRefreshSession(jwt, refreshToken);
+            return await this.validateOrRefreshSession(jwt, refreshToken, response);
         } catch (error) {
+            if (error instanceof CrossmintAuthenticationError && response != null) {
+                this.storeAuthMaterial(response, {
+                    jwt: "",
+                    refreshToken: {
```
AlbertoElias, line 51: I've added the `refreshToken` so it can be stored at will. It follows the object format from the response from `/refresh` to use the expiry date when configuring the cookie. An alternative would be to just pass around the string (simpler code) and set the cookie without an expiry and once it's invalid, it will throw errors anyway

packages/server/src/auth/utils/cookies.ts
```diff
@@ -39,3 +67,12 @@ function parseCookieHeader(cookieHeader: string): Record<string, string> {
...
     );
 }
+
+function createCookieString(cookieOptions: CookieOptions): string {
+    const expiresInUtc = cookieOptions.expiresAt ? new Date(cookieOptions.expiresAt).toUTCString() : "";
+    let cookieString = `${cookieOptions.name}=${cookieOptions.value}; path=/; SameSite=Lax;`;
+    if (expiresInUtc) {
+        cookieString += ` expires=${expiresInUtc};`;
+    }
+    return cookieString;
```
afeight, line 77: I know there's URLSearchParams that does this work for query params, there's nothing similar to use, right?

AlbertoElias: Sadly, not that I know of. The idea is to follow this pattern with other properties like `Secure` and `HttpOnly` once we also support more secure cookies

AlbertoElias: I could create my own util though in the httponly pr as we'll have more options to add there

afeight (approved):
cool!

PR #818, opened 2024-10-17
embed v3: crypto tx sending

mPaella:
## Description

sending crypto txs on sol and evm

## Test plan

tested both evm and sol

evm switching newtwork + sending working fine
sol sending is working fine, added a todo to work on network swtiching (its more complicated)

did e2e 
image

### DIFF
--- a/packages/client/ui/react-ui/src/components/embed/v3/crypto/utils/handleSendTransaction.ts
+++ b/packages/client/ui/react-ui/src/components/embed/v3/crypto/utils/handleSendTransaction.ts
@@ -1,0 +1,31 @@
+import type { Wallet } from "@dynamic-labs/sdk-react-core";
+
+import type { BlockchainIncludingTestnet } from "@crossmint/common-sdk-base";
+
+import { handleEvmTransaction } from "./handleEvmTransaction";
+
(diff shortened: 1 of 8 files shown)


packages/client/ui/react-ui/src/components/embed/v3/crypto/utils/handleSendTransaction.ts
```diff
@@ -0,0 +1,31 @@
...
+    const commonParams = {
+        chain,
+        serializedTransaction,
+        iframeClient,
+    };
+    if (isSolanaWallet(primaryWallet)) {
+        return await handleSolanaTransaction({
+            ...commonParams,
+            primaryWallet,
+        });
```
pgarciaegido, line 27: They reuse the entire object right? `commonParams` could include primaryWallet as well I think

pgarciaegido (approved):

packages/client/ui/react-ui/src/components/embed/v3/crypto/utils/handleSendTransaction.ts
mPaella: its a ts thing, cuz of the `isSolanaWallet` and `isEthereumWallet` so it has to be after commonParams

PR #828, opened 2024-10-23
Auth: Refactors auth sdks to bring in as much as possible to common and client

AlbertoElias:
## Description

We had too much of Auth, including managing refreshing tokens, in `react-ui.` This PR brings that into `client-sdk-auth`. Also brings into `common-sdk-auth` things in common between client and server. And adapts `client-sdk-auth` to new API surface as decided with `server-sdk`.

## Test plan

With Smarter Wallet and SSR demo apps. And adding/updating tests.

## Package updates

@crossmint/client-sdk-react-ui: minor
@crossmint/client-sdk-auth: major
@crossmint/common-sdk-auth: major
@crossmint/common-sdk-base: minor
@crossmint/server-sdk: minor


packages/common/base/src/types/Crossmint.ts
```diff
@@ -3,20 +3,18 @@ import { type ValidateAPIKeyPrefixExpectations, validateAPIKey } from "@/apiKey"
 export type Crossmint = {
     apiKey: string;
     jwt?: string;
-    refreshToken?: string;
```
afeight, line 6: mind explaining this? I lack context

afeight (approved):
lgtm

packages/client/auth/package.json
```diff
@@ -15,11 +15,15 @@
...
+        "dev": "tsup --watch",
+        "test": "vitest run"
     },
     "dependencies": {
+        "@crossmint/client-sdk-base": "workspace:*",
         "@crossmint/common-sdk-auth": "workspace:*",
-        "jwt-decode": "4.0.0"
+        "@crossmint/common-sdk-base": "workspace:*",
+        "jwt-decode": "4.0.0",
+        "@farcaster/auth-kit": "0.6.0"
```
guiwoda, line 27: Do we have to depend on farcaster on our SDKs?
We should be able to make it so that the developer brings farcaster in if they wish to login with it, right?

packages/client/auth/src/CrossmintAuthClient.test.ts
```diff
@@ -0,0 +1,275 @@
...
+            const mockCallback = vi.fn();
+            (crossmintAuthClient as any).callbacks.onTokenRefresh = mockCallback;
+
+            await crossmintAuthClient.handleRefreshToken(mockRefreshToken);
+
+            expect(mockCallback).toHaveBeenCalledWith(mockAuthMaterial);
+        });
+
+        it("should handle errors and call logout", async () => {
+            const mockError = new Error("Refresh failed");
```
guiwoda, line 127: nitpick: not a mock

packages/client/auth/src/CrossmintAuthClient.ts
```diff
@@ -0,0 +1,182 @@
...
+
+export class CrossmintAuthClient extends CrossmintAuth {
+    private callbacks: CrossmintAuthClientCallbacks;
+    private refreshTask: CancellableTask | null = null;
+    private isRefreshing = false;
+
+    private constructor(crossmint: Crossmint, callbacks: CrossmintAuthClientCallbacks = {}) {
+        super(crossmint);
+        this.callbacks = callbacks;
+        this.handleRefreshToken();
```
guiwoda, line 22: I would move this call to the `from` method, as this constructor is private and only called from there.
It's usually a bad practice to fire off processes from constructors, as the sole expected responsibility of them is to return an instance of the object.

packages/client/auth/src/CrossmintAuthClient.ts
```diff
@@ -0,0 +1,182 @@
...
+        setCookie(REFRESH_TOKEN_PREFIX, authMaterial.refreshToken.secret, authMaterial.refreshToken.expiresAt);
+    }
+
+    public logout() {
+        deleteCookie(REFRESH_TOKEN_PREFIX);
+        deleteCookie(SESSION_PREFIX);
+        this.callbacks.onLogout?.();
+    }
+
+    public async handleRefreshToken(refreshTokenSecret?: string): Promise<void> {
```
guiwoda, line 56: this method is `async`, but its promise is usually ignored.
I don't know the details of how memory management works client side, but is it safe to let a promise go out of scope without any references? In other ~more serious~ languages, a Garbage Collector would consider that object orphaned and clean it up.

packages/client/auth/src/CrossmintAuthClient.ts
```diff
@@ -0,0 +1,182 @@
...
+        this.handleRefreshToken();
+    }
+
+    public static from(crossmint: Crossmint, callbacks?: CrossmintAuthClientCallbacks): CrossmintAuthClient {
+        return new CrossmintAuthClient(crossmint, callbacks);
+    }
+
+    public getSession() {
+        this.handleRefreshToken();
+        return getCookie(SESSION_PREFIX);
```
guiwoda, line 31: Without an `await` in the previous line, we are not guaranteeing that the cookie is already set here.

packages/client/auth/src/CrossmintAuthClient.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    public storeAuthMaterial(authMaterial: AuthMaterialWithUser) {
+        setCookie(SESSION_PREFIX, authMaterial.jwt);
+        setCookie(REFRESH_TOKEN_PREFIX, authMaterial.refreshToken.secret, authMaterial.refreshToken.expiresAt);
+    }
+
+    public logout() {
+        deleteCookie(REFRESH_TOKEN_PREFIX);
+        deleteCookie(SESSION_PREFIX);
+        this.callbacks.onLogout?.();
```
guiwoda, line 53: Same comment as in the test: should we be revoking the refresh token here?

packages/client/auth/src/CrossmintAuthClient.ts
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    public async confirmEmailOtp(email: string, emailId: string, token: string) {
+        const queryParams = new URLSearchParams({
+            email,
+            signinAuthenticationMethod: "email",
+            token,
+            locale: "en",
+            state: emailId,
+            callbackUrl: `${this.apiClient.baseUrl}/api/${CROSSMINT_API_VERSION}/session/sdk/auth/we-dont-actually-use-this-anymore`,
```
guiwoda, line 109: Ouch, this looks awful hahaha!
Maybe change it to `dummy-url` or something?

packages/client/auth/src/CrossmintAuthClient.ts
```diff
@@ -0,0 +1,182 @@
...
+        });
+        const result = await this.apiClient.post(
+            `api/${CROSSMINT_API_VERSION}/session/sdk/auth/authenticate?${queryParams}`,
+            {
+                headers: {
+                    "Content-Type": "application/json",
```
guiwoda, line 115: nitpick, but it just crossed my mind: why isn't the `apiClient` adding this header for every request? we have non-json endpoints?

packages/client/ui/react-ui/src/components/auth/AuthForm.tsx
```diff
@@ -8,7 +8,7 @@ import { FarcasterProvider } from "../../providers/auth/FarcasterProvider";
 import { classNames } from "@/utils/classNames";
 
 export function AuthForm({ className }: { className?: string }) {
-    const { step, appearance, baseUrl, loginMethods } = useAuthForm();
+    const { step, appearance, loginMethods, baseUrl } = useAuthForm();
```
guiwoda, line 11: Unnecessary?

packages/client/ui/react-ui/src/components/auth/methods/email/EmailOTPInput.tsx
```diff
@@ -22,17 +22,13 @@ export function EmailOTPInput({
     const handleOnSubmit = async () => {
         setLoading(true);
         try {
-            const oneTimeSecret = await onConfirmEmailOtp(
+            const oneTimeSecret = await crossmintAuth?.confirmEmailOtp(
```
guiwoda, line 25: Why might `crossmintAuth` be `null` here?

packages/client/ui/react-ui/src/hooks/useCrossmintAuth.ts
```diff
@@ -0,0 +1,12 @@
+import { useContext } from "react";
+import { AuthContext } from "../providers/CrossmintAuthProvider";
+
+export function useCrossmintAuth() {
```
guiwoda, line 4: This looks like it is always truthy, we shouldn't need to validate on usage.

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -65,31 +59,30 @@ export function CrossmintAuthProvider({
...
     const [user, setUser] = useState<SDKExternalUser | undefined>(undefined);
-    const { crossmint, setJwt, setRefreshToken } = useCrossmint(
-        "CrossmintAuthProvider must be used within CrossmintProvider"
-    );
-    const crossmintAuthService = new CrossmintAuthService(crossmint.apiKey);
+    const { crossmint, setJwt } = useCrossmint("CrossmintAuthProvider must be used within CrossmintProvider");
+    // Only create the CrossmintAuth instance once, even in StrictMode, as the constructor calls /refresh
+    // It can only be called once to avoid race conditions
+    const crossmintAuthRef = useRef<CrossmintAuth | null>(null);
+    // biome-ignore lint/correctness/useExhaustiveDependencies: crossmint can't be a dependency because it updates with each jwt change
```
guiwoda, line 66: hmmm maybe we can drill down on that (in a future PR?)

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -65,31 +59,30 @@ export function CrossmintAuthProvider({
...
+    const crossmintAuth = useMemo(() => {
+        if (!crossmintAuthRef.current) {
+            crossmintAuthRef.current = CrossmintAuth.from(crossmint, {
```
guiwoda, line 69: Ooooh now I understand the need for hooks. I thought it was for the developer, but it's actually to let us hook here in react.
I wonder if that could also confuse developers in the future and they won't know what to do on their hooks.

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -65,31 +59,30 @@ export function CrossmintAuthProvider({
...
+                onLogout: () => {
+                    setJwt(undefined);
+                    setUser(undefined);
```
guiwoda, line 72: shouldn't we call `logout()` as well?

packages/common/auth/src/CrossmintAuth.ts
```diff
@@ -0,0 +1,49 @@
+import { type Crossmint, CrossmintApiClient } from "@crossmint/common-sdk-base";
+import { CROSSMINT_API_VERSION, SDK_NAME, SDK_VERSION } from "./utils/constants";
+import { type AuthMaterialWithUser, CrossmintAuthenticationError, type AuthMaterialResponse } from "./types";
+
+export class CrossmintAuth {
+    protected crossmint: Crossmint;
+    protected apiClient: CrossmintApiClient;
+
+    constructor(crossmint: Crossmint) {
```
guiwoda, line 9: if this is meant to be private, I'd also add `CrossmintApiClient` as a dependency and move creation of that object to the `from` method.

guiwoda: Maybe this was designed to be extended?
If so, I'd suggest redesigning to favor composition over inheritance. Let me know if I can help with that!

guiwoda (commented):
Hey, great job! Looks very thorough and well thought.
I left a bunch of comments, let's continue the conversation on each one and I'll review again!

packages/client/auth/package.json
guiwoda: After reading the entire PR, I think this comment doesn't make much sense from my part.
I do think we should be able to split up auth providers in the future, but for now I think this should be fine.

packages/client/ui/react-ui/src/components/auth/AuthForm.tsx
AlbertoElias: yeah, initially I removed baseUrl thinking I didn't need it, but it ended up being necessary in one place and I added it back and it stayed like this haha

packages/client/auth/src/CrossmintAuthClient.ts
AlbertoElias: I thought about this too, but didn't want to touch the `apiClient` just in case and followed the same pattern used elsewhere

packages/client/auth/src/CrossmintAuthClient.ts
AlbertoElias: This is so we can mimick the API decided for the server sdk but in most cases it won't be used like this. We only expect this to return the jwt. `handleRefreshToken` triggers a task to get the most recent auth material and set off a timer to re-check later on so it's more like a background process that we want to make sure it's being called on init. Right now we're sticking to triggering this when creating an instance in the constructor

packages/client/auth/src/CrossmintAuthClient.ts
AlbertoElias: copying from another comment:

> handleRefreshToken triggers a task to get the most recent auth material and set off a timer to re-check later on so it's more like a background process that we want to make sure it's being called on init.

That's why this doesn't return anything

packages/client/auth/src/CrossmintAuthClient.ts
AlbertoElias: Fair, I explained in other comments why we're calling this

packages/client/auth/src/CrossmintAuthClient.test.ts
```diff
@@ -0,0 +1,275 @@
...
+        it("should clear auth cookies and call onLogout callback", () => {
+            const mockCallbacks = { onLogout: vi.fn() };
+            crossmintAuthClient = CrossmintAuthClient.from(mockCrossmint as unknown as Crossmint, mockCallbacks);
+
+            crossmintAuthClient.logout();
+
+            expect(cookiesUtils.deleteCookie).toHaveBeenCalledWith("crossmint-refresh-token");
+            expect(cookiesUtils.deleteCookie).toHaveBeenCalledWith("crossmint-jwt");
+            expect(mockCallbacks.onLogout).toHaveBeenCalled();
+        });
```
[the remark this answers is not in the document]
AlbertoElias: We also need to delete the cookies. But what we're missing is calling the server when performing a logout

packages/client/ui/react-ui/src/hooks/useCrossmintAuth.ts
AlbertoElias: what do you mean? what is always truthy?

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
AlbertoElias: well, how the `Crossmint` object is managed is that a new jwt triggers a new version: https://github.com/Crossmint/crossmint-sdk/blob/main/packages/client/ui/react-ui/src/hooks/useCrossmint.tsx#L23 how would you suggest changing that?

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
AlbertoElias: not sure what you're referring to

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
AlbertoElias: this is a callback that is called after the react sdk calls logout

packages/common/auth/src/CrossmintAuth.ts
AlbertoElias: Open to suggestions. Here I implemented as per the approved PRD and we discussed multiple API surfaces and ended up going with this one. So following this `from` approach is a must. But if there's anything we can do to improve the implementation, let me know!

packages/client/auth/src/CrossmintAuthClient.ts
guiwoda: Yes, thank you for that!
I do still believe it'd be cleaner for this call to be in the `from` method instead of the constructor. I do understand it has to be triggered on auth client initialization, what I suggest is merely moving it out of the constructor to keep the constructor clean of logic that doesn't exclusively represent constructing the instance.

packages/client/auth/src/CrossmintAuthClient.ts
guiwoda: I understand it doesn't return anything, but the promise is still running in the background.
Given I'm not familiar with JS, I asked GPT and this is what I got:

**TL;DR: You can safely ignore the `Promise<void>` response, as JS is just.. JS.**

> No, JavaScript’s garbage collector will not stop a promise that is still pending or hasn’t resolved yet, even if it becomes unreachable.
> 
> In JavaScript, promises are designed to run to completion once they are created. This means that when a promise is created, the asynchronous operation it represents will proceed independently of whether there are references to the promise object. If the promise resolves (or rejects), it will do so regardless of whether any code is waiting for it or whether it has been stored in a variable.
> 
> Here’s what happens in two main cases:
> 
> 	1.	When a promise is unreachable but pending: JavaScript will not garbage collect the underlying asynchronous operation (such as a network request or a timeout) until it completes. So, the promise will still resolve or reject when that operation finishes.
> 	2.	When a promise is unreachable and resolved: Once a promise has resolved or rejected, and if there are no references to it, it can be garbage collected like any other object.
> 
> This approach ensures the reliability of asynchronous operations in JavaScript.

packages/client/auth/src/CrossmintAuthClient.ts
guiwoda: But what's the expected use if this returns `null` because the cookie is not set yet?
We've already triggered a `refresh` in the background, but the developer doesn't have any `Promise` to wait for it.

packages/client/ui/react-ui/src/components/auth/methods/email/EmailOTPInput.tsx
guiwoda: oooh I get it. Thanks!

packages/client/ui/react-ui/src/hooks/useCrossmintAuth.ts
guiwoda: Nevermind, it was related to the `undefined` question above, but I didn't know about the provider init stuff.

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
guiwoda: I have no clue what that means to React. It feels like this has something to do with component refresh mechanisms related to state change?
I wouldn't know how changing the jwt affects react components, but I do know that we shouldn't need to rebuild the entire service layer because of the jwt changing... it's even expected to change often because of behind-the-scenes refreshs.
Why can't we have a regular memoized value?

packages/common/auth/src/CrossmintAuth.ts
guiwoda: I'm not against the static `from` method. I'm just not sure why we'd have inheritance as the extension mechanism.

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
guiwoda: the `onLogout` and `onTokenRefresh` "hooks" are defined in the auth-sdk-client and then used here.
What's the expected usage of those methods if used outside of the `react-ui` sdk?

packages/client/auth/src/CrossmintAuthClient.ts
AlbertoElias: Yeah, for further context, all async tasks enter the event loop. It has different stacks that as it loops around picks up a task from each stack. Promises I believe enter the microTask stack and remain there until resolved or rejected. Sorry, wasn't sure what you were referring to.

P.D. In case you want to learn more, this video explains it best: https://www.youtube.com/watch?v=cCOL7MC4Pl0

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
AlbertoElias: Other apps that use the auth client sdk directly, instead of the react one, will want to do similar actions as the ones we take here in the react sdk. The idea of the client auth sdk is to be framework agnostic

guiwoda: Great! Now I understand the separation:
- CrossmintAuthClient will remove the cookies (cross-framework way of storing the material)
- React-UI will hook to handle component state change
  - Other frameworks might want to do the same
  - When using React-UI, developers are not expected to add their own hooks to `onLogout` or `onRefreshToken`.

packages/client/auth/src/CrossmintAuthClient.ts
AlbertoElias: I think I'll just delete it. This is a half-baked method just to be API compliant with server sdk. It returns the jwt if it's there for immediate use as we needed that behaviour in the react sdk, but I couldn't find a way to substitute calling `handleRefreshToken` on init and calling it here instead in a way that I could make it work sensibly with the current sdk. So I left the function half-baked, and maintained the other behaviour that uses callbacks. If we followed the way the server sdk does it, we wouldn't return the jwt until it has been validated and refreshed.

packages/client/auth/src/CrossmintAuthClient.ts
AlbertoElias: it still needs to be valid url, and this param is used in the oauth flow, just not used anymore in the rest of the flows. I think it's good to have an ugly name as a reminder that we need to refactor the server side

guiwoda (approved):
LGTM let's go!!

PR #834, opened 2024-10-25
Also send api key in header in oauth fetch to support backend change

AlbertoElias:
## Description

With [this PR](https://github.com/Paella-Labs/crossbit-main/pull/15388), we need all Auth requests to send the API key in the header instead of as a query parameter. We were missing this oauth fetch.

## Test plan

Opened smarter wallet and saw in the console the api key is now sent in the header.

## Package updates

react-ui: patch


MatiNavCrossmint (approved):
LGTM!

PR #837, opened 2024-10-25
Tree shaking: react-ui fix

jmderby:
## Description

This PR adds tree shaking capabilities back to the react-ui package
The solution around our twind issue was to add a TwindProvider that runs `install` inside a useEffect on mount.

## Test plan

Tested build and styles worked as expected.

## Package updates

### DIFF
--- a/packages/client/ui/react-ui/package.json
+++ b/packages/client/ui/react-ui/package.json
@@ -15,7 +15,7 @@
     "types": "./dist/index.d.ts",
     "files": ["dist", "src", "LICENSE"],
     "scripts": {
-        "build": "tsup",
+        "build": "NODE_OPTIONS='--max-old-space-size=8192' tsup",
         "dev": "tsup --watch",
         "test": "vitest run"
     },
--- a/packages/client/ui/react-ui/src/hooks/useCrossmint.tsx
+++ b/packages/client/ui/react-ui/src/hooks/useCrossmint.tsx
@@ -1,7 +1,6 @@
 import { type ReactNode, createContext, useCallback, useContext, useMemo, useRef, useState } from "react";
 
 import { type Crossmint, createCrossmint } from "@crossmint/common-sdk-base";
-import { TwindProvider } from "@/providers/TwindProvider";
 
 export interface CrossmintContext {
     crossmint: Crossmint;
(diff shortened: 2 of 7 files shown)


packages/client/ui/react-ui/package.json
```diff
@@ -15,7 +15,7 @@
     "types": "./dist/index.d.ts",
     "files": ["dist", "src", "LICENSE"],
     "scripts": {
-        "build": "tsup",
+        "build": "NODE_OPTIONS='--max-old-space-size=8192' tsup",
```
jmderby, line 18: building locally wouldn't work until i increased the old-space memory size.

packages/client/ui/react-ui/src/hooks/useCrossmint.tsx
```diff
@@ -56,7 +57,11 @@ export function CrossmintProvider({
         [setJwt, setRefreshToken, version]
     );
 
-    return <CrossmintContext.Provider value={value}>{children}</CrossmintContext.Provider>;
+    return (
+        <CrossmintContext.Provider value={value}>
```
mPaella, line 61: hmm i would go even deeper/more component specific than this

for ex, we wont be using twind for hosted or embedded checkout, so having it this high isnt optimal

mPaella: i'd be fine with it on CrossmintAuthProvider for now, or maybe just on the actual component

jmderby: definitely, i'll toss is lower 👍

mPaella (approved):

packages/client/ui/react-ui/src/providers/TwindProvider.tsx
```diff
@@ -0,0 +1,14 @@
+import type React from "react";
+import twindConfig from "@/twind.config";
+import { install } from "@twind/core";
+import { useEffect } from "react";
+
+export function TwindProvider({ children }: { children: React.ReactNode }) {
```
AlbertoElias, line 6: Shouldn't this also wrap all other providers exported by the library? Or maybe only the main Crossmint provider as that's a necessary import for everything else?

PR #841, opened 2024-10-29
Auth: UI style fixes + demo width fix

jmderby:
## Description

PR contains changes regarding broken styles once package was built and ran in a client app. 
This PR also fixes small bug with demo screen width. I removed the viewport override from the twind file.
I worked with Pablo UX in collaborating these changes.

Screenshot 2024-10-29 at 10 51 55 AM

## Test plan

Built app and ran from a completely separate client app running package via `yalc`. Tested custom styles as well and everything worked as expected.

## Package updates


packages/client/ui/react-ui/src/twind.config.ts
```diff
@@ -91,6 +93,12 @@ export default defineConfig({
                 "500": "500ms",
             },
         },
+        fontFamily: {
+            sans: ["Inter", "sans-serif"],
+        },
+    },
+    preflight: {
```
jmderby, line 100: we might not wanna preflight the fonts as it may interfere with client apps consuming this lib.

packages/client/ui/react-ui/src/twind.config.ts
```diff
@@ -4,18 +4,17 @@ import presetTailwind from "@twind/preset-tailwind";
 export default defineConfig({
     presets: [presetTailwind()],
     theme: {
-        screens: {
-            xs: "480px",
```
jmderby, line 8: adding this screen size was causing the client app (demo) to have broken widths. removing fixes it.

packages/client/ui/react-ui/src/twind.config.ts
AlbertoElias: Yeah, I wouldn't force a font either, I think we should inherit and make a suggestion at most

AlbertoElias (approved):
would also remove font preflight

pablo-paella:
The powered by Crossmint looks like its using a different font.
We should use Jakarta from Google Fonts.

packages/client/ui/react-ui/src/twind.config.ts
jmderby: I removed any font forcing and just use defaults.

AlbertoElias (approved):

PR #846, opened 2024-11-05
Auth: Supports setting custom routes for refrsh and configuring cookie options

AlbertoElias:
## Description

To support a more secure way to store authentication material, we'd like to support them being stored in Secure + HttpOnly cookies. To do so, they need to be set server-side. So the client can no longer try to refresh tokens from Crossmint's API and then set it in the client. Instead, we provide a utility for 3Ps to create their custom refresh route and configure the different cookie options. It's compatible with all JS backend environments based on either Fetch or Node.JS API.

## Test plan

Updated all tests and provided node.js, express, remix and next.js demo apps to test it.

## Package updates

`@crossmint/common-sdk-auth`: minor
`@crossmint/client-sdk-auth`: minor
`@crossmint/server-sdk`: minor
`@crossmint/client-sdk-react-ui`: minor


packages/client/auth/src/CrossmintAuthClient.ts
```diff
@@ -53,16 +62,22 @@ export class CrossmintAuthClient extends CrossmintAuth {
         this.callbacks.onLogout?.();
     }
 
-    public async handleRefreshToken(refreshTokenSecret?: string): Promise<void> {
+    public async handleRefreshAuthMaterial(refreshTokenSecret?: string): Promise<void> {
```
guiwoda, line 65: Nit: I feel this name change isn't that big of a deal and it made you rename all the references everywhere.
Is there anything I'm missing here?

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -25,26 +25,32 @@ export type CrossmintAuthProviderProps = {
...
     login: () => {},
     logout: () => {},
+    jwt: undefined,
+    user: undefined,
     status: "logged-out",
     getUser: () => {},
-});
+};
+
+export const AuthContext = createContext<AuthContextType>(defaultContextValue as AuthContextType);
```
guiwoda, line 53: Why is the `as AuthContextType` needed here? The variable is typed already.

packages/common/auth/src/CrossmintAuth.ts
```diff
@@ -16,34 +21,68 @@ export class CrossmintAuth {
...
+    public static from(crossmint: Crossmint, options: CrossmintAuthOptions = {}): CrossmintAuth {
+        return new CrossmintAuth(crossmint, options);
     }
 
     public getJwksUri() {
         return `${this.apiClient.baseUrl}/.well-known/jwks.json`;
     }
 
-    protected async refreshAuthMaterial(refreshToken: string): Promise<AuthMaterialWithUser> {
+    protected async refreshAuthMaterial(refreshToken?: string): Promise<AuthMaterialWithUser> {
```
guiwoda, line 35: Why would we allow custom refresh routes to work without a refresh token?

packages/server/README.md
```diff
@@ -49,3 +49,57 @@ const { jwt, userId } = await crossmintAuth.getSession({
...
+    cookieOptions: {
+        httpOnly: true,
+        sameSite: "Strict",
+        secure: true,
+        domain: ".example.com",
+    },
+});
+```
+
+## Advanced Usaged
```
guiwoda, line 95: Why would this section be advanced?
I thought the entire point of this was to allow a custom refresh route that can store the cookie as httpOnly

packages/server/src/auth/CrossmintAuthServer.ts
```diff
@@ -58,12 +70,78 @@ export class CrossmintAuthServer extends CrossmintAuth {
         return user;
     }
 
+    public async handleCustomRefresh(request: GenericRequest, response?: ServerResponse): Promise<GenericResponse> {
+        // If the request from the client includes a refresh token (like the OneTimeSecret after authenticating), use that
+        // Otherwise, try to get the refresh token from the cookies
+        const body = isFetchRequest(request) ? await request.json() : await getNodeRequestBody(request);
```
guiwoda, line 76: The `isFetch` / `isNode` conditionals are all over the place.
Maybe we can have two separate classes with a shared interface, and select the correct class once at the beginning of this method?
Something like 

```typescript
interface RequestAdapter {
  toJson();
  toResponseError(message);
  withHeaders(headers): RequestAdapter;
  withBody(body): RequestAdapter;
  // etc.
}

class FetchRequestAdapter implements RequestAdapter {
    // fetch-specific stuff here
}
class NodeRequestAdapter implements RequestAdapter {
     // node-specific stuff here
}

File: packages/server/src/auth/types/request.ts [hunk 6]
```diff
@@ -18,3 +18,50 @@ export function isNodeResponse(response: GenericResponse): response is ServerRes
...
+
+        req.on("data", (chunk) => {
+            body += chunk.toString();
+        });
+
+        req.on("end", () => {
+            try {
+                resolve(JSON.parse(body));
+            } catch (_) {
+                resolve(body);
```
guiwoda, line 39: I would add a comment that explains why this error is ignored.

```
// If it fails to parse the body as JSON, just return the body as a string.
```

packages/server/src/auth/utils/cookies.ts
```diff
@@ -21,13 +22,24 @@ export function getAuthCookies(request: GenericRequest): AuthMaterialBasic {
...
+export function setAuthCookies(response: GenericResponse, authMaterial: AuthMaterial, options?: CookieOptions) {
     const cookies = [
-        createCookieString({ name: SESSION_PREFIX, value: authMaterial.jwt }),
+        createCookieString({
+            name: SESSION_PREFIX,
+            value: authMaterial.jwt,
+            options: {
+                ...options,
+                // The JWT needs to be accessible to the browser for the client SDK to work
+                httpOnly: false,
```
guiwoda, line 33: I understand this, but should we document it in the README?
Otherwise it might be confusing for the developer to configure `httpOnly: true` on the `CrossmintAuth` options and then see it `false` in their debugging sessions.

packages/client/auth/README.md
```diff
@@ -45,15 +45,27 @@ You can provide callbacks for token refresh and logout events when creating the
...
+        },
     },
 });
 ```
 
+You can also provide a custom refresh route:
+
+```typescript
+const crossmintAuth = CrossmintAuthClient.from(crossmint, {
+    refreshRoute: "/api/refresh",
```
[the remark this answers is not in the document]
AlbertoElias: It actually depends on the environment. In next.js it's fine with just the path, in remix I had to provide the whole URL. will document it

packages/client/auth/src/CrossmintAuthClient.ts
AlbertoElias: I thought the original name was really bad, and as it hasn't shipped yet, it's not a breaking change. I think the new name is more consistent with the naming elsewhere

packages/common/auth/src/CrossmintAuth.ts
AlbertoElias: When HttpOnly is set, the client can't send a refresh token because it can't access the cookie, so it needs to be fetched from the cookie in the server

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
AlbertoElias: This was a leftover attempt at fixing a TypeScript issue. The actual fix was passing default values for `jwt`, `user` etc as undefined

packages/server/README.md
AlbertoElias: It's a more complicated setup, but I see what you mean, I'll change it to a more enticing title

packages/server/README.md
```diff
@@ -49,3 +49,57 @@ const { jwt, userId } = await crossmintAuth.getSession({
...
+
+You can also provide a custom refresh route:
+
+```typescript
+const crossmintAuth = CrossmintAuthClient.from(crossmint, {
+    refreshRoute: "/api/refresh",
+});
+```
+
+This way, the SDK will use the provided route to refresh the token instead of the default one and the authentication material can be stored in HttpOnly cookies that are tied to the domain of the provided route.
```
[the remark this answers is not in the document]
AlbertoElias: The idea behind `handleCustomeRefresh` is to return the Response object that 3P developers need to return in their endpoints to make it super easy for them and they can forget about the details involved. It also handles the appropriate errors to match the call to `refresh` if it wasn't via a custom route

packages/client/auth/README.md
```diff
@@ -39,12 +39,42 @@ const user = await crossmintAuth.getUser();
 crossmintAuth.logout();
 ```
 
+## Secure setup using HttpOnly cookies
+
+To secure the authentication material, you can set up a custom endpoint in your backend that will handle refreshing the authenticatino material and storing it in HttpOnly cookies. This way, the authentication material is not accessible to JavaScript running in the browser.
```
guiwoda, line 44: nit: typo in `authenticatino`

guiwoda (approved):
LGTM! 🎉

packages/server/src/auth/CrossmintAuthServer.ts
AlbertoElias: I actually only do it there and in the cookies util to fetch the header so I think this abstraction is not worth adding right now as I think the current code is pretty declarative to read. there is one other `isFetchRequest` but that is only to prepare the appropriate response. wdyt?

guiwoda: I don't know, I don't feel that this code is "pretty declarative" at all tbh.
There's 5 uses of either `isFetch` or `isNode` just in this method, and you're saying there's other uses outside of it...

packages/server/src/auth/types/request.ts
```diff
@@ -11,15 +12,101 @@ export function isFetchRequest(request: GenericRequest): request is Request {
...
+}
+
+export class NodeResponseAdapter implements ResponseAdapter {
+    constructor(private response?: ServerResponse) {}
+
+    setCookies(cookies: string[]) {
+        if (this.response == null) {
+            throw new Error("Response not found");
+        }
+        this.response.setHeader("Set-Cookie", cookies);
```
guiwoda, line 65: So Node supports sending an array of cookies here and it will handle setting multiple `Set-Cookie` headers, while fetch needs each cookie to be set separately, right?

packages/server/src/auth/types/request.ts
```diff
@@ -11,15 +12,101 @@ export function isFetchRequest(request: GenericRequest): request is Request {
...
+    setCookies(cookies: string[]) {
+        if (this.response == null) {
+            throw new Error("Response not found");
+        }
+
+        cookies.forEach((cookie) => this.response?.headers.append("Set-Cookie", cookie));
+    }
+
+    setAuthMaterial(authMaterial: AuthMaterialWithUser) {
+        return new Response(JSON.stringify(authMaterial), {
```
guiwoda, line 97: Isn't this method loosing state set to `this.response` before? i.e. cookies?

packages/server/src/auth/CrossmintAuthServer.ts
```diff
@@ -114,14 +108,15 @@ export class CrossmintAuthServer extends CrossmintAuth {
...
+            const errorResponse = isNodeResponse(response)
+                ? response
+                : responseAdapter.setError(401, errorResponseBody);
 
             this.logout(request, errorResponse);
 
             // We need to set the rest of the Node response AFTER setting headers like we do in logout
             if (isNodeResponse(errorResponse)) {
-                return setNodeResponseError(errorResponse, 401, errorResponseBody);
+                return responseAdapter.setError(401, errorResponseBody);
```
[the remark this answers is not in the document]
AlbertoElias: the response needs to be created in a different order for each type of response. As I explained in the comment, the body of the response, in Node's case, needs to happen after setting the headers, while with Fetch, it needs to be set in the constructor and the headers go after

packages/server/src/auth/types/request.ts
AlbertoElias: i create a different adapter for each instance. for cookies, I need to pass in a response, for `handleCustomRefresh` I need to create the response instance

PR #849, opened 2024-11-08
Better auth error handling and notifying end users

AlbertoElias:
## Description

Catch auth errors properly and in react sdk show them to end users.

## Test plan

Testing with smarter wallet

## Package updates

react-ui: patch
common-auth: patch
client-auth: minor

### DIFF
--- a/packages/client/ui/react-ui/src/providers/auth/AuthFormProvider.tsx
+++ b/packages/client/ui/react-ui/src/providers/auth/AuthFormProvider.tsx
@@ -14,12 +14,14 @@ const initialOAuthUrlMap: OAuthUrlMap = {
 };
 interface AuthFormContextType {
     step: AuthStep;
+    error: string | null;
     appearance?: UIConfig;
     loginMethods: LoginMethod[];
     oauthUrlMap: OAuthUrlMap;
     isLoadingOauthUrlMap: boolean;
     baseUrl: string;
     setStep: (step: AuthStep) => void;
+    setError: (error: string | null) => void;
     setDialogOpen: (open: boolean) => void;
 }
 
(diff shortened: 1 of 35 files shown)


packages/client/ui/react-ui/src/providers/auth/AuthFormProvider.tsx
```diff
@@ -47,6 +49,7 @@ export const AuthFormProvider = ({
 }: { children: ReactNode; initialState: ContextInitialStateProps }) => {
     const { crossmintAuth } = useCrossmintAuth();
     const [step, setStep] = useState<AuthStep>("initial");
```
jmderby, line 51: I'm thinking since we don't reset the error when they navigate to another connector in example: 
1. user enters invalid email, the error display at the top nicely
2. user clicks on farcaster login, error still displays from invalid email.

one way I can think to fix this would be to add a useEffect inside this provider here

```
useEffect(() => {
  if(error != null){
    setError(null)
  }
}, [step, error])
```
it's possible i'm missing something in that snippet code but we might want to fix this somehow. 

Screenshot 2024-11-12 at 11 01 24 AM

AlbertoElias: That useEffect is proving to be a bit more problematic. What I'm going to do is so that on all clicks to the `<AuthFormBackButton>` it'll reset the error

jmderby: You could do it that way too, you'll also have to reset the error when clicking on any provider

AlbertoElias: yup, did that too. more stable that way

jmderby: for sure, i think the explicit approach will be best to prevent future bugs too 👍

jmderby: hmm, it still appears to be broken on embedded. 
Screenshot 2024-11-12 at 4 18 43 PM

jmderby (approved):
Looking good, I'm going to approve under the following conditions: 
1. Pablo has approved the UI 
2. Small minor fix required when an error displaying and clicking into "Connect with a wallet" -> error still showing. Not a huge deal but would be nice if error disappeared.  (vid below)

https://github.com/user-attachments/assets/7c68e0da-9074-43a2-9a48-45d0fbbec6c1

AlbertoElias:
@jmderby fixed that last comment and yup, approved by Pablo

PR #862, opened 2024-11-13
window: new tab client

mPaella:
## Description

new tab window client

## Test plan

local, tested setting pay button to open in new tab

### DIFF
--- a/packages/client/base/src/services/hosted/v3/crossmintHostedCheckoutV3Service.ts
+++ b/packages/client/base/src/services/hosted/v3/crossmintHostedCheckoutV3Service.ts
@@ -1,6 +1,6 @@
 import type { CrossmintHostedCheckoutV3Props } from "@/types/hosted/v3/CrossmintHostedCheckoutV3Props";
 import { appendObjectToQueryParams } from "@/utils/appendObjectToQueryParams";
-import { PopupWindow } from "@crossmint/client-sdk-window";
+import { NewTabWindow, PopupWindow } from "@crossmint/client-sdk-window";
 import type { CrossmintApiClient } from "@crossmint/common-sdk-base";
 import { crossmintHostedCheckoutOverlayService } from "./crossmintHostedCheckoutOverlayService";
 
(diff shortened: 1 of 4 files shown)


packages/client/base/src/services/hosted/v3/crossmintHostedCheckoutV3Service.ts
```diff
@@ -27,37 +27,37 @@ export function crossmintHostedCheckoutV3Service({
...
     }
 
-    // TODO: Implement new tab client
-    function createNewTabClient(): ReturnType<typeof PopupWindow.initSync> {
-        throw new Error("Not implemented");
+    function createNewTabClient(url: string): ReturnType<typeof NewTabWindow.initSync> {
+        return NewTabWindow.initSync(url, {});
     }
 
     // TODO: Implement same tab client
```
pgarciaegido, line 42: can we remove todo?

packages/client/window/src/windows/NewTab.ts
```diff
@@ -0,0 +1,65 @@
...
+        const popup = createNewTabSync(url);
+        return new NewTabWindow<IncomingEvents, OutgoingEvents>(
+            popup,
+            options.targetOrigin || urlToOrigin(url),
+            options
+        );
+    }
+}
+
+function createNewTabSync(url: string) {
```
pgarciaegido, line 48: could this be private methods instead of just functions?

pgarciaegido (approved):

packages/client/base/src/services/hosted/v3/crossmintHostedCheckoutV3Service.ts
mPaella: thats on the next if, which isnt implemented yet

packages/client/window/src/windows/NewTab.ts
mPaella: ehh i dont think it really makes a difference, and this is the pattern for all of them...

PR #863, opened 2024-11-15
Makes it impossible to call refresh twice in the client

AlbertoElias:
## Description

Strong assurance that refresh won't be called twice

## Test plan

I created a test

## Package updates

client-auth: patch

### DIFF
--- a/packages/client/auth/src/CrossmintAuthClient.ts
+++ b/packages/client/auth/src/CrossmintAuthClient.ts
@@ -22,7 +22,6 @@ type CrossmintAuthClientConfig = CrossmintAuthOptions & {
 export class CrossmintAuthClient extends CrossmintAuth {
     private callbacks: CrossmintAuthClientCallbacks;
     private refreshTask: CancellableTask | null = null;
-    private isRefreshing = false;
     private refreshPromise: Promise<AuthMaterialWithUser> | null = null;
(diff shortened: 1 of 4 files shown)


packages/client/auth/src/CrossmintAuthClient.ts
```diff
@@ -178,7 +183,7 @@ export class CrossmintAuthClient extends CrossmintAuth {
         try {
             const queryParams = new URLSearchParams({
                 signinAuthenticationMethod: "farcaster",
-                callbackUrl: `${this.apiClient.baseUrl}/${AUTH_SDK_ROOT_ENDPOINT}/callback?isPopup=false`,
+                callbackUrl: `${this.apiClient.baseUrl}/${AUTH_SDK_ROOT_ENDPOINT}/callback`,
```
jmderby, line 186: nice!

packages/client/auth/src/CrossmintAuthClient.ts
```diff
@@ -46,6 +46,8 @@ export class CrossmintAuthClient extends CrossmintAuth {
                 headers: { "Content-Type": "application/json" },
             });
 
+            console.error("getUser", response);
```
jmderby, line 49: can prob remove XD

jmderby (approved):
lgtm

PR #890, opened 2024-12-16
Auth: Twitter

jmderby:
## Description

Add twitter auth to our auth adk

## Test plan

tested end to end flow, twitter users are now created and used in our system via stytch! 

## Package updates

@crossmint/common-sdk-auth

### DIFF
--- a/packages/client/ui/react-ui/src/components/auth/AuthForm.tsx
+++ b/packages/client/ui/react-ui/src/components/auth/AuthForm.tsx
@@ -9,7 +9,7 @@ import { PoweredByCrossmint } from "../common/PoweredByCrossmint";
 import { FarcasterProvider } from "../../providers/auth/FarcasterProvider";
 import { classNames } from "@/utils/classNames";
 import { AlertIcon } from "@/icons/alert";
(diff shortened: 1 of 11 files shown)


packages/client/ui/react-ui/src/components/auth/AuthForm.tsx
```diff
@@ -9,6 +9,7 @@ import { PoweredByCrossmint } from "../common/PoweredByCrossmint";
 import { FarcasterProvider } from "../../providers/auth/FarcasterProvider";
 import { classNames } from "@/utils/classNames";
 import { AlertIcon } from "@/icons/alert";
+import { XSignIn } from "./methods/x/XSignIn";
```
AlbertoElias, line 12: if we're using the name twitter everywhere, I would stick to that, or rename to X everywhere

AlbertoElias (approved):

packages/client/ui/react-ui/src/components/auth/AuthForm.tsx
jmderby: nice catch, i renamed everything to twitter except for the button itself

>Sign in with X

PR #912, opened 2025-01-29
Auth: Adapts server sdk to non-nodejs envs and adds example to next auth app

AlbertoElias:
## Description

I added an example with middleware to our nextjs-auth app. To make this work, we needed the server-sdk to run nicely in non-nodejs environments, so I also changed our dependencies to use `jose` which does work.

## Test plan

Added unit test and tested with example app, where I added a new page to test the flow where we refresh directly with Crossmint servers

## Package updates

server-sdk: minor
next-auth: minor


apps/auth/nextjs-ssr/src/hooks/auth.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    try {
+        const crossmint = createCrossmint({
+            apiKey: process.env.SERVER_CROSSMINT_API_KEY || "",
+        });
+        const crossmintAuth = CrossmintAuth.from(crossmint, {
+            refreshRoute,
+        });
+
+        const session = await crossmintAuth.getSession({
```
mPaella, line 21: this doesnt refresh the token, correct?

mPaella, line 21: just overall meta comment: it seems weird that u need both the refresh and the jwt to get the session, or is that wrong? i think with just the jwt you should be able to

AlbertoElias: Nope, if the jwt is valid, there's no need to refresh. And the refresh had just happened in the middleware so we're good

AlbertoElias: I see how the naming is confusing yeah, but really the core token is the refreshToken as the jwt is almost always expired.

PR #929, opened 2025-02-08
Auth: Optionally dont throw when cookies not found

AlbertoElias:
## Description

With a custom refresh route, we send the onetimesecret in the body, not in the cookies, but we check both. We need to make sure we dont throw an error if cookie not found in that case.

## Test plan

Tested with nextjs ssr auth example

## Package updates

common-sdk-auth: patch
server-sdk: patch


packages/server/src/auth/utils/cookies.ts
```diff
@@ -52,11 +55,11 @@ export function setAuthCookies(response: GenericResponse, authMaterial: AuthMate
     responseAdapter.setCookies(cookies);
 }
 
-function getCookieHeader(request: GenericRequest): string {
+function getCookieHeader(request: GenericRequest, throwError = true): string | null | undefined {
```
afeight, line 58: Do we need to add both `| null | undefined`?
What does `const cookieHeader = requestAdapter.getCookieHeader()` return?

packages/server/src/auth/CrossmintAuthServer.ts
```diff
@@ -135,8 +138,11 @@ export class CrossmintAuthServer extends CrossmintAuth {
         try {
             // It's not necessary to call the logout endpoint, but desirable
             if (request != null) {
-                const { refreshToken } = getAuthCookies(request);
-                await this.logoutFromDefaultRoute(refreshToken);
+                const authCookies = getAuthCookies(request);
+                if (authCookies != null) {
+                    const { refreshToken } = authCookies;
+                    await this.logoutFromDefaultRoute(refreshToken);
```
afeight, line 144: nit: can do this since it's not reused
```suggestion
                    await this.logoutFromDefaultRoute(authCookies.refreshToken);
```

afeight (approved):
couple comments but overall looks ok

packages/server/src/auth/utils/cookies.ts
AlbertoElias: the fetch adapters can return undefined and the node adapter can return null hahaha but i'll move the undefined to force a null

PR #935, opened 2025-02-13
Fix: stale JWT re-render bug

jmderby:
## Description

Fixed issue with React not re-rendering upon jwt invalidation and new one being set

## Test plan

1. repro bug
2. fix and remove ability to repro bug

## Package updates

### DIFF
--- a/packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
+++ b/packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
@@ -163,6 +163,7 @@ export function CrossmintAuthProvider({
                 }}
             >
                 <CrossmintWalletProvider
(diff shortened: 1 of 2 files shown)


packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -163,6 +163,7 @@ export function CrossmintAuthProvider({
                 }}
             >
                 <CrossmintWalletProvider
+                    key={crossmint.jwt}
```
jmderby, line 166: by adding the key to the top level CrossmintWalletProvider it can ensure re-renders happen when they need to based on the jwt state

AlbertoElias (approved):

PR #937, opened 2025-02-13
UI design overhaul

jmderby:
## Description

This pr contains all of the changes from our design UI

Screenshot 2025-02-13 at 1 25 34 PM

Screenshot 2025-02-13 at 1 25 56 PM

Screenshot 2025-02-13 at 8 13 41 PM

### Features

- new feel 
- new prop exposed called `termsOfServiceText` that can pass a string or reactnode with links nested
- new prop exposed: `authModalTitle` that allows the title of the modal to be changed
- email input shows a "Continue" button that logs you in with google if you use a `gmail.com` domain
  - plus, if your already authenticated in the same browser, it'll auto sign in

## Test plan

- built and tested with yalc
- customized appearance object styles
- exhausted a lot of manual testing 

## Package updates

### DIFF
--- a/.changeset/bright-foxes-hope.md
+++ b/.changeset/bright-foxes-hope.md
@@ -1,0 +1,5 @@
+---
+"@crossmint/client-sdk-react-ui": minor
+---
+
+New UI look and feel
--- a/packages/client/ui/react-ui/src/components/auth/AuthForm.tsx
+++ b/packages/client/ui/react-ui/src/components/auth/AuthForm.tsx
@@ -19,7 +19,7 @@ const Web3AuthFlow = lazy(() =>
 );
 
 export function AuthForm({ className }: { className?: string }) {
-    const { step, appearance, loginMethods, baseUrl, error, termsOfServiceText } = useAuthForm();
+    const { step, appearance, loginMethods, baseUrl, error, termsOfServiceText, authModalTitle } = useAuthForm();
 
     return (
         <div
@@ -50,7 +50,7 @@ export function AuthForm({ className }: { className?: string }) {
                         className="text-2xl font-bold text-cm-text-primary"
                         style={{ color: appearance?.colors?.textPrimary }}
                     >
-                        Sign in to Crossmint
+                        {authModalTitle ?? "Sign in to Crossmint"}
                     </h1>
                     <p
                         className="text-base font-normal mb-3 text-cm-text-secondary"
@@ -77,18 +77,18 @@ export function AuthForm({ className }: { className?: string }) {
                 </div>
             ) : null}
 
-            {termsOfServiceText != null ? (
-                <p
+            {step === "initial" && termsOfServiceText != null ? (
+                <div
                     className="text-sm text-center text-cm-text-secondary mt-2"
                     style={{ color: appearance?.colors?.textSecondary }}
                 >
                     <style>{`
                         p a {
-                            color: ${appearance?.colors?.accent ?? "#04AA6D"};
+                            color: ${appearance?.colors?.textLink ?? "#1A74E9"};
                         }
                     `}</style>
                     {termsOfServiceText}
-                </p>
+                </div>
             ) : null}
 
             {step === "initial" || step === "otp" ? (
--- a/packages/client/ui/react-ui/src/components/auth/AuthForm.tsx
+++ b/packages/client/ui/react-ui/src/components/auth/AuthForm.tsx
@@ -19,12 +19,12 @@ const Web3AuthFlow = lazy(() =>
 );
 
 export function AuthForm({ className }: { className?: string }) {
-    const { step, appearance, loginMethods, baseUrl, error } = useAuthForm();
+    const { step, appearance, loginMethods, baseUrl, error, termsOfServiceText } = useAuthForm();
 
     return (
         <div
             className={classNames(
-                "relative p-6 pb-4 !min-[480px]:p-10 !min-[480px]:pb-8 flex flex-col gap-[10px] antialiased animate-none",
+                "relative pt-10 pb-[30px] px-6 !min-[480px]:px-10 flex flex-col gap-[10px] antialiased animate-none",
                 className
             )}
         >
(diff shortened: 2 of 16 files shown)


packages/client/ui/react-ui/src/components/auth/AuthForm.tsx
```diff
@@ -50,24 +50,17 @@ export function AuthForm({ className }: { className?: string }) {
                         className="text-2xl font-bold text-cm-text-primary"
                         style={{ color: appearance?.colors?.textPrimary }}
                     >
-                        Sign In
+                        Sign in to Crossmint
```
AlbertoElias, line 53: wow, we're putting this front and center. i hope clients are okay with that

AlbertoElias (approved):

packages/client/ui/react-ui/src/components/auth/AuthForm.tsx
jmderby: I'm slightly concerned about this decision too perhaps i'll add a heading prop or something. wdyt?

AlbertoElias: Sg

packages/client/ui/react-ui/src/components/auth/methods/email/EmailSignIn.tsx
```diff
@@ -93,19 +112,56 @@ export function EmailSignIn({ setOtpEmailData }: { setOtpEmailData: (data: OtpEm
                                     }}
                                 />
                             )}
-                            {!emailError && !isLoading && (
+                            {showGoogleContinueButton ? (
+                                <button
```
AlbertoElias, line 116: this feels prime for a separate component so the oauthlistener hook is kept out side of this flow

packages/client/ui/react-ui/src/components/auth/methods/email/EmailSignIn.tsx
```diff
@@ -93,19 +112,56 @@ export function EmailSignIn({ setOtpEmailData }: { setOtpEmailData: (data: OtpEm
...
+                                                height: "18px",
+                                            }}
+                                        />
+                                    ) : (
+                                        <span className="text-cm-accent" style={{ color: appearance?.colors?.accent }}>
+                                            Continue
+                                        </span>
+                                    )}
+                                </button>
+                            ) : !emailError && !isLoading ? (
```
AlbertoElias, line 149: all these conditional chains are getting hard to parse. can we break this down?

packages/client/ui/react-ui/src/components/auth/methods/google/GoogleSignIn.tsx
```diff
@@ -26,7 +26,7 @@ export function GoogleSignIn({ ...props }: ButtonHTMLAttributes<HTMLButtonElemen
                 borderRadius: appearance?.borderRadius,
                 backgroundColor: appearance?.colors?.buttonBackground,
             }}
-            onClick={isLoading ? undefined : createPopupAndSetupListeners}
+            onClick={isLoading ? undefined : () => createPopupAndSetupListeners()}
```
AlbertoElias, line 29: why do we need this?

packages/client/ui/react-ui/src/hooks/useOAuthWindowListener.ts
```diff
@@ -25,13 +25,33 @@ export const useOAuthWindowListener = (provider: OAuthProvider) => {
...
             throw new Error("Child window not initialized");
         }
         setIsLoading(true);
         setError(null);
-        const popup = await PopupWindow.init(oauthUrlMap[provider], {
+
+        const baseUrl = new URL(oauthUrlMap[provider]);
+
+        // If we have a providerLoginHint, we need to add it to the beginning of the url
+        if (providerLoginHint != null) {
```
AlbertoElias, line 38: can you explain what this is about?

packages/client/ui/react-ui/src/components/common/CountdownButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+    countdownText,
+    countdownCompleteText,
+    handleOnClick,
+}: CountdownButtonProps) {
+    const [seconds, setSeconds] = useState(0);
+    const [canResend, setCanResend] = useState(true);
+    const [retryCount, setRetryCount] = useState(0);
+
+    useEffect(() => {
+        let timer: NodeJS.Timeout;
```
AlbertoElias, line 25: as this is clientside, there should be another type for this

packages/client/ui/react-ui/src/components/common/CountdownButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+                        setCanResend(true);
+                        return 0;
+                    }
+                    return prev - 1;
+                });
+            }, 1000);
+        }
+
+        return () => {
+            if (timer) {
```
AlbertoElias, line 41: null check?

packages/client/ui/react-ui/src/components/auth/methods/google/GoogleSignIn.tsx
jmderby: The onClick handler passes an event but now because I've added a param to `createPopupAndSetupListeners` ts throws 
```
Type '((providerLoginHint?: string | undefined) => Promise<void>) | undefined' is not assignable to type 'MouseEventHandler<HTMLButtonElement> | undefined'.
  Type '(providerLoginHint?: string) => Promise<void>' is not assignable to type 'MouseEventHandler<HTMLButtonElement>'.
```

packages/client/ui/react-ui/src/hooks/useOAuthWindowListener.ts
jmderby: the login hint is now used in the "continue with google" button and will pass to Stytch and automatically login a user if they already have an existing session in the browser

jmderby: we're fitting that query param at the very first position of the url and then appending the remaining ones back on the url. URL is coming from the /start endpoint which happens before the user fills out the email input which is why I had to append it after the fact on the client side, rather than the /start endpoint api side

.changeset/bright-foxes-hope.md
jmderby, line 1: added as minor instead of patch because this is a fairly large ui change

packages/client/ui/react-ui/src/hooks/useOAuthWindowListener.ts
AlbertoElias: i see, and why do we need to prepend it in the url? probably worth linking to those docs in the code

PR #941, opened 2025-02-14
Auth: Prevent /start calls when logged in

jmderby:
## Description

Changes prevent /start endpoint to be called unless they are logged out

## Test plan

Tested and /start calls weren't made when user was logged in or logging in. 

## Package updates


mPaella (approved):
cool, lets give it a shot

PR #943, opened 2025-02-15
Auth: add onLoginSuccess callback

jmderby:
## Description

Adds a callback for successful logins called onLoginSuccess

## Test plan

works for both pre-wallet users and logged in users

## Package updates

### DIFF
--- a/packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
+++ b/packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
@@ -119,11 +119,12 @@ export function CrossmintAuthProvider({
         setDialogOpen(false);
 
         // FOR STANDALONE AUTH ONLY! EXCLUDING PASSKEY HELPERS.
-        // Skip if passkey helpers are enabled because they have to still create a wallet.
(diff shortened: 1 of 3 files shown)


packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -115,7 +117,13 @@ export function CrossmintAuthProvider({
         }
 
         setDialogOpen(false);
-    }, [crossmint.jwt]);
+
+        // FOR STANDALONE AUTH ONLY! EXCLUDING PASSKEY HELPERS.
```
alfonso-paella, line 121: could we make it fire after the passkey helpers in this scenario?

alfonso-paella: or maybe you are doing it below

alfonso-paella (approved):

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
jmderby: Yeah, i had to do it below in the wallet provider too

PR #956, opened 2025-02-18
Init Wallet SDK

Destiner:
## Description

Initiates a new Wallet SDK (`@crossmint/wallets-sdk`) and defines the high-level project structure.

Also introduces code generation based on the OpenAPI schema using [HeyAPI](https://heyapi.dev/openapi-ts/get-started) and a high-level API client.

Example (to be used internally):

```ts
const apiClient = new ApiClient(apiKey);

const walletCreationResponse = await apiClient.createWallet({
  type: "evm-smart-wallet",
  config: {
    adminSigner: {
      type: "evm-keypair",
      address: account.address,
    },
  },
});
const address = walletCreationResponse.address;
```

## Test plan

* Build passes
* Stub definitions correspond to the PRD


packages/wallets/src/sdk.ts
```diff
@@ -0,0 +1,75 @@
...
+    public async getOrCreateWallet(
+        type: "evm-mpc-wallet",
+        ...args: WalletTypeToArgs["evm-mpc-wallet"]
+    ): Promise<EVMMPCWallet>;
+    public async getOrCreateWallet(
+        type: "solana-smart-wallet",
+        ...args: WalletTypeToArgs["solana-smart-wallet"]
+    ): Promise<SolanaSmartWallet>;
+    public async getOrCreateWallet(
+        type: "solana-mpc-wallet",
```
Destiner, line 57: Looks awkward, and a bit harder to maintain, but this will give the consumers great DX:

```ts
const wallet = await sdk.getOrCreateWallet("evm-smart-wallet", adminSigner: …);
//      ^ this will be typed as `EVMSmartWallet`               ^ this will have full autocomplete
```

mPaella: i think u can just make another type map like `WalletTypeToWallet` to use as the return type on the actual implementation func, and it should do the same thing without needing to define all these

packages/wallets/src/solana/wallet.ts
```diff
@@ -0,0 +1,56 @@
...
+    sendTransaction: (parameters: {
+        transaction: Transaction | VersionedTransaction;
+        options?: {
+            skipPreflight?: boolean;
+            maxRetries?: number;
+        };
+    }) => Promise<string>; // Returns transaction signature
+}
+
+export class SolanaSmartWallet implements SolanaWallet {
```
mPaella, line 28: same about checking goat repo

packages/wallets/src/evm/chains.ts
```diff
@@ -0,0 +1,34 @@
...
+const SMART_WALLET_TESTNETS: readonly SmartWalletChain[] = objectValues(SmartWalletTestnet);
+
+const SmartWalletMainnet = {
+    BASE: Blockchain.BASE,
+    POLYGON: Blockchain.POLYGON,
+    OPTIMISM: Blockchain.OPTIMISM,
+    ARBITRUM: Blockchain.ARBITRUM,
+} as const;
+type SmartWalletMainnet = ObjectValues<typeof SmartWalletMainnet>;
+const SMART_WALLET_MAINNETS: readonly SmartWalletChain[] = objectValues(SmartWalletMainnet);
```
mPaella, line 19: i think i would just define this as like `SMART_WALLET_SUPPORTED_TESTNETS = [Blockchain.BASE_SEPOLIA, ...]`

rather than an object and having to re-define all the names of the blockchains

packages/wallets/src/sdk.ts
Destiner: addressed below

packages/wallets/src/sdk.ts
```diff
@@ -0,0 +1,75 @@
...
+        ...args: WalletTypeToArgs["solana-mpc-wallet"]
+    ): Promise<SolanaMPCWallet>;
+
+    // biome-ignore lint/suspicious/useAwait: <explanation>
+    public async getOrCreateWallet<WalletType extends keyof WalletTypeToArgs>(
+        _type: WalletType,
+        ..._args: WalletTypeToArgs[WalletType]
+    ): Promise<EVMSmartWallet | EVMMPCWallet> {
```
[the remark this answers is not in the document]
Destiner: nice, this simplifies things a lot! fixed in https://github.com/Crossmint/crossmint-sdk/pull/956/commits/9abda4d5eb276b9490d6bdde12405de5d7afde85

packages/wallets/src/evm/wallet.ts
```diff
@@ -0,0 +1,71 @@
...
+    ) => Promise<Hex>;
+
+    sendTransaction: (parameters: {
+        to: Address;
+        data?: Hex;
+        value?: bigint;
+    }) => Promise<Hex>;
+}
+
+export class EVMSmartWallet implements ViemWallet {
```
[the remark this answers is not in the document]
Destiner: Looked into it. Some things are similar, others are different (e.g. we don't need `getResolveName` here).

Probably best to keep separate to not have a tight dependency between two different projects.

packages/wallets/src/solana/wallet.ts
Destiner: addressed above

packages/wallets/openapi-ts.config.ts
Destiner, line 1: There are plenty of tooling to generate the. Went with HeyAPI:

- low-level API (easy to build a high-level SDK around it)
- can set base URL dynamically
- has a plugin for zod (not used currently)
- `fetch`-based (no runtime dependencies like `axios`)
- generates both runtime code and the typings

packages/wallets/src/api/client.ts
```diff
@@ -0,0 +1,302 @@
...
+
+type CreateSignatureParams = CreateSignatureRequestDto;
+type CreateSignatureResponse = WalletsV1Alpha2SignatureResponseDto;
+type ApproveSignatureParams = SubmitApprovalDto;
+type ApproveSignatureResponse = WalletsV1Alpha2SignatureResponseDto;
+type GetSignatureResponse = WalletsV1Alpha2SignatureResponseDto;
+
+type GetTransactionsResponse = WalletsV1Alpha2TransactionsResponseDto;
+type GetNftsResponse = Nftevm | Nftsol;
+type GetBalanceResponse = WalletBalanceResponseDto;
```
Destiner, line 51: Renaming the types before the export to keep them aligned with `ApiClient` methods

packages/wallets/src/api/client.ts
```diff
@@ -0,0 +1,302 @@
...
+        );
+    }
+
+    async createWallet(params: CreateWalletParams): Promise<CreateWalletResponse> {
+        const response = await walletsV1ControllerCreateWallet4({
+            client: this.client,
+            body: params,
+            headers: {
+                "X-API-KEY": this.apiKey,
+            },
```
alfonso-paella, line 82: there seems to be some redundancy here, as we will need to pass this on every call

Any way to abstract that out s e.g. the client takes care of adding this header?

packages/wallets/src/api/client.ts
```diff
@@ -0,0 +1,302 @@
...
+type GetTransactionsResponse = WalletsV1Alpha2TransactionsResponseDto;
+type GetNftsResponse = Nftevm | Nftsol;
+type GetBalanceResponse = WalletBalanceResponseDto;
+
+class ApiClient {
+    private readonly client: Client;
+
+    constructor(
+        private readonly apiKey: string,
+        private readonly jwt?: string
```
alfonso-paella, line 58: the jwt can change over time (e.g. when the user refreshes the token). im not sure this ap client supports jwt mutations?

packages/wallets/src/api/client.ts
```diff
@@ -0,0 +1,302 @@
...
+type CreateSignatureResponse = WalletsV1Alpha2SignatureResponseDto;
+type ApproveSignatureParams = SubmitApprovalDto;
+type ApproveSignatureResponse = WalletsV1Alpha2SignatureResponseDto;
+type GetSignatureResponse = WalletsV1Alpha2SignatureResponseDto;
+
+type GetTransactionsResponse = WalletsV1Alpha2TransactionsResponseDto;
+type GetNftsResponse = Nftevm | Nftsol;
+type GetBalanceResponse = WalletBalanceResponseDto;
+
+class ApiClient {
```
alfonso-paella, line 53: I thought we had already something like a api client class in base SDK. Did you check? Would it be possible to reuse parts of that?

packages/wallets/src/api/client.ts
```diff
@@ -0,0 +1,302 @@
...
+    constructor(
+        private readonly apiKey: string,
+        private readonly jwt?: string
+    ) {
+        const validationResult = validateAPIKey(this.apiKey);
+        if (!validationResult.isValid) {
+            throw new Error("Invalid API key");
+        }
+        if (validationResult.usageOrigin === APIKeyUsageOrigin.CLIENT && !jwt) {
+            throw new Error("JWT is required for client API key");
```
alfonso-paella, line 65: Im nt sure jwt is always needed - why did you reach that conclusion?

alfonso-paella: I think it depends on the specific API you;re calling

packages/wallets/src/evm/chains.ts
```diff
@@ -0,0 +1,22 @@
+import { BlockchainIncludingTestnet as Blockchain } from "@crossmint/common-sdk-base";
+
+const SMART_WALLET_TESTNET_CHAINS = [
+    Blockchain.BASE_SEPOLIA,
+    Blockchain.POLYGON_AMOY,
+    Blockchain.OPTIMISM_SEPOLIA,
+    Blockchain.ARBITRUM_SEPOLIA,
+];
+
+const SMART_WALLET_MAINNET_CHAINS = [Blockchain.BASE, Blockchain.POLYGON, Blockchain.OPTIMISM, Blockchain.ARBITRUM];
```
AlbertoElias, line 10: don't we support more of them?

packages/wallets/src/evm/wallet.ts
```diff
@@ -0,0 +1,71 @@
+import type { Address, Hex, SignableMessage, PublicClient, HttpTransport, TypedData, TypedDataDefinition } from "viem";
+
+import type { SmartWalletChain } from "./chains";
+
+export interface ViemWallet {
```
AlbertoElias, line 5: interesting that viem doesnt export smth like this haha

packages/wallets/src/api/client.ts
Destiner: this is the limitation of this tool (& multiple others) i'm afraid.

I've refactored it a bit in https://github.com/Crossmint/crossmint-sdk/pull/956/commits/a4adf98aefb1fb01ee16c7035cdf710e00faf9c6 by adding a `getHeaders` helper

packages/wallets/src/api/client.ts
Destiner: Didn't know it can change, thanks for context!

No it doesn't support that. But note that the SDK doesn't support that as well (I know there was a discussion on supporting that, but I haven't seen a specific API proposed in the PRD). So it doesn't limit us, at least for now.

packages/wallets/src/api/client.ts
Destiner: We check the presence of JWT when the API key is client-side. 

Are there any situations where we access the client-side APIs, but don't have the JWT?

packages/wallets/openapi-ts.config.ts
```diff
@@ -0,0 +1,13 @@
+import { defineConfig } from "@hey-api/openapi-ts";
+
+export default defineConfig({
+    input: "src/openapi.json",
```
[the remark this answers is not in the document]
Destiner: Not sure this tool can do it, but also how do we access it remotely given the repo is private?

packages/wallets/src/evm/chains.ts
Destiner: added Story in https://github.com/Crossmint/crossmint-sdk/pull/956/commits/7594825780a70acaa840a8913a1aaa03f18d5ca3

to add Mode we first need to update the chain definitions in `@crossmint/common-sdk-base`.

packages/wallets/src/api/client.ts
Destiner: ended up using `CrossmintApiClient` to make API calls in https://github.com/Crossmint/crossmint-sdk/pull/956/commits/9a34dd9f082adc0962ed5d5cc483adfaf8007f85

packages/wallets/src/api/client.ts
Destiner: we're using `CrossmintApiClient` now, so no longer an issue

packages/wallets/src/evm/chains.ts
AlbertoElias: Is it not possible to reuse the supported chains that we have in the openapi schema? It'd be nice to only have to update it in one place and not have to remember to do it here as well

packages/wallets/openapi-ts.config.ts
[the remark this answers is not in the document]
AlbertoElias: We do have them public. The general one is: `https://crossmint.com/api/json` and the wallets specific one is: `https://crossmint.com/api/openapi/wallets/json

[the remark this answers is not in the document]
AlbertoElias: This is definitely smth that would be very nice to have

packages/wallets/src/api/client.ts
AlbertoElias: Then we don't use the autogenerated low level calls and types..?

packages/wallets/src/api/client.ts
```diff
@@ -0,0 +1,223 @@
...
+type CreateSignatureResponse = WalletsV1Alpha2SignatureResponseDto;
+type ApproveSignatureParams = SubmitApprovalDto;
+type ApproveSignatureResponse = WalletsV1Alpha2SignatureResponseDto;
+type GetSignatureResponse = WalletsV1Alpha2SignatureResponseDto;
+
+type GetTransactionsResponse = WalletsV1Alpha2TransactionsResponseDto;
+type GetNftsResponse = Nftevm | Nftsol;
+type GetBalanceResponse = WalletBalanceResponseDto;
+
+class ApiClient {
```
mPaella, line 40: why not extend `CrossmintApiClient`?

Destiner: that makes total sense https://github.com/Crossmint/crossmint-sdk/pull/956/commits/3059e57e1a362d5dea5e0b57cd123cb2219f5587

packages/wallets/src/api/client.ts
devlyn37, line 1: Do we want `me` requests here too?

devlyn37 (commented):
Left a few comments, only major thing I'd consider blocking is making SDK instantiation consistent w/ `CrossmintAuth`

packages/wallets/src/openapi.json
alberto-crossmint, line 1: Hmm so what's the strategy we'll be following to keep this up to date? I during the time we're reviewing that we've misaligned this a couple of times :D

packages/wallets/src/api/client.ts
alberto-crossmint: Agree! But I also think we should do that on our main app so the spec is correctly generated, there's no benefit on keeping the Dto suffix everywhere

alberto-crossmint: But ofc we can do that later

packages/wallets/src/evm/wallet.ts
alberto-crossmint: It seems like we should be able to extract those types from a viems LocalAccount?

packages/wallets/src/sdk.test.ts
```diff
@@ -0,0 +1,30 @@
...
+});
+
+describe("WalletSDK", () => {
+    let sdk: WalletSDK;
+
+    beforeEach(() => {
+        // Reset mocks before each test
+        vi.clearAllMocks();
+        sdk = new WalletSDK(
+            "sk_staging_A4vDwAp4t5az6fVQMpQK6qapBnAqgpxrrD35TaFQnyKgxehNbd959uZeaHjNCadWDXrgLRAK1CxeasZjtYEq4TbFkKMBBvbQ9oinAxQf8LbHsSYW2DMzT8fBko3YGLq9t7ZiXZjmgkTioxGVUUjyLtWLeBKwNUDLgpshWjaoR7pKRnSE9SqhwjQbiK62VKiBTdA3KvHsyG9k8mLMcKrDyfXp"
```
alberto-crossmint, line 21: delete?

packages/wallets/src/sdk.ts
```diff
@@ -0,0 +1,65 @@
...
+
+type SolanaAdminSigner =
+    | {
+          type: "solana-keypair";
+          address: string;
+      }
+    | {
+          type: "solana-fireblocks-custodial";
+      };
+
```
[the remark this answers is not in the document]
alberto-crossmint: It's nested on the wallet type under config.adminSigner but yeah it should be possible!

packages/wallets/src/solana/wallet.ts
```diff
@@ -0,0 +1,56 @@
+import type { Connection, PublicKey, Transaction, VersionedTransaction } from "@solana/web3.js";
+
+export interface SolanaWallet {
+    // Get the wallet's public key (equivalent to EVM address)
+    getPublicKey: () => PublicKey;
+
+    // Optional nonce (similar concept but implementation differs)
+    getNonce?: ((parameters?: { seed?: string | undefined } | undefined) => Promise<bigint>) | undefined;
```
alberto-crossmint, line 8: Does not apply!

alberto-crossmint: We should delete

packages/wallets/src/solana/wallet.ts
```diff
@@ -0,0 +1,56 @@
+import type { Connection, PublicKey, Transaction, VersionedTransaction } from "@solana/web3.js";
+
+export interface SolanaWallet {
```
alberto-crossmint, line 3: Can we define only mandatory methods on this interface? Like, the signMessage or signTransaction methods will need to be defined on the `SolanaSmartWallet` but they are not supported so will need to throw. That's a bad ux imo because user would see these methods as available when developing without knowing that they are not supported

packages/wallets/src/solana/wallet.ts
```diff
@@ -0,0 +1,56 @@
...
+    public async transactions() {}
+    public async nfts() {}
+
+    public getPublicKey(): PublicKey {
+        throw new Error("Not implemented");
+    }
+    // biome-ignore lint/suspicious/useAwait: stub
+    public async getNonce(): Promise<bigint> {
+        throw new Error("Not implemented");
+    }
```
alberto-crossmint, line 41: we should delete

packages/wallets/src/solana/wallet.ts
Destiner: Those were taken from the PRD as is.

Removed, and feel free to make further changes!

packages/wallets/src/sdk.ts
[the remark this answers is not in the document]
Destiner: yup it's possible! https://github.com/Crossmint/crossmint-sdk/pull/956/commits/6ac63ecd761e6f006954598dae8f30e08ac77413

packages/wallets/src/sdk.ts
```diff
@@ -0,0 +1,65 @@
...
+type WalletTypeToArgs = {
+    "evm-smart-wallet": [adminSigner: EVMAdminSigner, linkedUser?: string];
+    "evm-mpc-wallet": [linkedUser?: string];
+    "solana-smart-wallet": [adminSigner: SolanaAdminSigner, linkedUser?: string];
+    "solana-mpc-wallet": [linkedUser?: string];
```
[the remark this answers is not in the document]
Destiner: https://github.com/Crossmint/crossmint-sdk/pull/956/commits/454c083ed702859d60088b7dbbb6a918e6d2c5cc

packages/wallets/src/solana/wallet.ts
Destiner: https://github.com/Crossmint/crossmint-sdk/pull/956/commits/34fbd3504c05d8d292e551f9f86a71fffb301d68

packages/wallets/src/solana/wallet.ts
Destiner: https://github.com/Crossmint/crossmint-sdk/pull/956/commits/34fbd3504c05d8d292e551f9f86a71fffb301d68

packages/wallets/src/sdk.test.ts
Destiner: makes sense https://github.com/Crossmint/crossmint-sdk/pull/956/commits/80e9afc50abf3659a02127cf58e72167a61142d4

packages/wallets/src/evm/chains.ts
```diff
@@ -0,0 +1,29 @@
+import { BlockchainIncludingTestnet as Blockchain } from "@crossmint/common-sdk-base";
+
+const SMART_WALLET_TESTNET_CHAINS = [
```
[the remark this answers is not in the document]
Destiner: makes sense https://github.com/Crossmint/crossmint-sdk/pull/956/commits/80e9afc50abf3659a02127cf58e72167a61142d4

packages/wallets/src/api/client.ts
alfonso-paella: Yes! Anything that doesn't hit /me I think

For example i think fetching wallet balance and others

In general - we dont have an invariant in the server that is "if its a client side api, it requires jwt". It is something we configure call by call

packages/wallets/src/api/client.ts
alfonso-paella: I'm not sure i follow!

The JWT can mutate without an API - it updates whenever we refresh the token, which is an internal operation of the auth SDK.

packages/wallets/src/evm/wallet.ts
Destiner: I've looked into this, and the best we can probably do is to import each method individually.

packages/wallets/src/api/client.ts
Destiner: yup added https://github.com/Crossmint/crossmint-sdk/pull/956/commits/f3687ea5223621bb9e8ff731eb97d9c13e90cbf1

packages/wallets/src/sdk.ts
```diff
@@ -0,0 +1,65 @@
...
+};
+
+type WalletTypeToWallet = {
+    "evm-smart-wallet": EVMSmartWallet;
+    "evm-mpc-wallet": EVMMPCWallet;
+    "solana-smart-wallet": SolanaSmartWallet;
+    "solana-mpc-wallet": SolanaMPCWallet;
+};
+
+class WalletSDK {
```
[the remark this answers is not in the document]
Destiner: changed the entry point interface to match `CrossmintAuth` https://github.com/Crossmint/crossmint-sdk/pull/956/commits/ec9098e891f63ebb2bc8cc34eb51c3173cf18de9

the functionality (e.g. JWT refresh) is not there yet

packages/wallets/src/sdk.test.ts
Destiner: good call https://github.com/Crossmint/crossmint-sdk/pull/956/commits/bdf54499abaed010439219a06a3302a162b976e2

Destiner: Actually this is needed for unit tests to work. Rolling back.

packages/wallets/src/api/client.ts
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    async getMeWallet(type: WalletType): Promise<GetWalletResponse> {
+        const response = await this.get(`${this.apiPrefix}/me:${type}`, {
+            headers: {
+                "Content-Type": "application/json",
+            },
+        });
+        return response.json();
+    }
```
alberto-crossmint, line 92: As discussed offline we may need some kind of wrapper for client and server side operations on top of this class, to avoid checking in consumer classes whether we're calling server or client endpoints. Approving this PR so we get started with the wallet impl! But I think we should iterate on that later

packages/wallets/src/api/client.ts
Destiner: Still using the types, just not the runtime code.

packages/wallets/src/api/client.ts
Destiner: The initial SDK spec didn't account for that, and this implementation doesn't either.

However, we're thinking of adopting most/all functionality from `CrossmintAuth`, which will include JWT refresh.

packages/wallets/openapi-ts.config.ts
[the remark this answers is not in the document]
Destiner: Oh didn't know, will look into this!

PR #968, opened 2025-03-03
React Smart Wallet App: fix build issue

panosinthezone:
## Description
The package inside `apps/wallets/smart-wallet/react` wasn't building.

**Root cause**
The abitype package (which is a dependency of viem) was looking for @babel/runtime@7.24.7, specifically the wrapRegExp.js file.
However, the project had @babel/runtime@7.26.0 and @babel/runtime@7.9.0 installed, but not version 7.24.7.
This version mismatch caused the build to fail because it couldn't find the specific file in the expected location.

**Solution**:
Installed specific version of @babel/runtime that was being referenced using:
`pnpm install @babel/runtime@7.24.7 --save-exact`

## Test plan
Can build with `cd apps/wallets/smart-wallet/react &&npm run build`
image

### DIFF
--- a/apps/wallets/smart-wallet/react/package.json
+++ b/apps/wallets/smart-wallet/react/package.json
@@ -10,18 +10,8 @@
         "test:e2e-ui": "pnpm exec playwright test --ui"
     },
     "browserslist": {
-        "production": [
-            "chrome >= 67",
-            "edge >= 79",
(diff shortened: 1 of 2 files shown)


panosinthezone:
There are other build issues with the skd but will not address right now

AlbertoElias (approved):

PR #970, opened 2025-03-05
Wallets SDK: EVM + Solana smart wallets

Destiner:
## Description

Implements both client-side and server-side EVM smart wallets.

Also, updates the `ApiClient` to route calls into either client-side (`/me`) or server-side endpoints based on the provided API key.

## Test plan

- CI pass
- Creating wallets, making transactions and signatures should work
  - When providing a server-side API key, it should use the server-side endpoints (using the wallet locator)
  - When providing a client-side API key and a JWT, it should use the client-side `/me` endpoints (using the wallet type)

## Package updates

* `@crossmint/wallets-sdk`: initial `0.1.0` release


packages/wallets/package.json
```diff
@@ -26,6 +26,7 @@
         "@crossmint/common-sdk-base": "workspace:*",
         "@hey-api/client-fetch": "0.8.1",
         "@solana/web3.js": "1.95.1",
+        "ox": "0.6.9",
```
Destiner, line 29: Needed for passkeys

packages/wallets/src/solana/wallet.ts
```diff
@@ -1,7 +1,13 @@
 import type { Connection, PublicKey, Transaction, VersionedTransaction } from "@solana/web3.js";
 
+import type { CreateWalletDto } from "@/api/gen/types.gen";
+
 import type { ApiClient } from "../api";
 
+export type SolanaSigner = NonNullable<
+    Extract<CreateWalletDto, { type: "solana-smart-wallet" }>["config"]
+>["adminSigner"];
+
```
Destiner, line 10: Moved the signer definition into the module itself, similar to EVM.

@alberto-crossmint I've tried to minimize Solana-related changes, but I think this one is necessary. You'll probably want to use this to instantiate the `SolanaWallet` anyway.

packages/wallets/src/sdk.test.ts
```diff
@@ -1,6 +1,6 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
-import WalletSDK from "./sdk";
+import { CrossmintWallet } from "./sdk";
```
Destiner, line 3: Aligning this with the PRD, should have done this in the previous PR!

packages/wallets/src/api/client.ts
```diff
@@ -1,4 +1,4 @@
-import { type Crossmint, CrossmintApiClient } from "@crossmint/common-sdk-base";
+import { type Crossmint, APIKeyUsageOrigin, CrossmintApiClient, validateAPIKey } from "@crossmint/common-sdk-base";
```
Destiner, line 1: Here, we're making the API client more high-level by:

1. Aligning client- and server-side endpoint methods to have a similar signature
2. Switching between those endpoints based on the API key

That way, the consumer of `ApiClient` don't need to worry about that at all!

packages/wallets/src/evm/wallet.ts
```diff
@@ -26,48 +68,318 @@ export interface ViemWallet {
...
-    public async nfts() {}
+    public async balances(tokens: Address[]) {
+        return await this.apiClient.getBalance(this.walletLocator, {
+            chains: [this.chain],
+            tokens,
+        });
+    }
+    public async transactions() {
+        const transactions = await this.apiClient.getTransactions(this.walletLocator);
+        return transactions.transactions.filter((transaction) => transaction.walletType === "evm-smart-wallet");
```
AlbertoElias, line 93: why do we need this filter?

packages/wallets/src/evm/wallet.ts
```diff
@@ -26,48 +68,318 @@ export interface ViemWallet {
...
+            },
+        });
+        const signatureId = signatureCreationResponse.id;
+
+        // Approve signature
+        const pendingApprovals = signatureCreationResponse.approvals?.pending || [];
+        if (pendingApprovals.length !== 1) {
+            throw new Error(`Expected 1 pending approval, got ${pendingApprovals.length}`);
+        }
+        const pendingApproval = pendingApprovals[0];
```
AlbertoElias, line 136: we should find pending approvals by locator, making sure the signer in the pendingApproval is the same as `this.signerLocator`. Same in transactions and typed data

packages/wallets/src/evm/wallet.ts
```diff
@@ -26,48 +68,318 @@ export interface ViemWallet {
...
+            throw new Error("Transaction hash not found");
+        }
+        return transactionHash as Hex;
+    }
+
+    private get walletLocator(): string {
+        if (this.apiClient.isServerSide) {
+            return this.address;
+        } else {
+            return "evm-smart-wallet";
```
AlbertoElias, line 271: in client side we're actually returning the walletType, the locator would be me:evm-smart-wallet. I would consider changing the response here and remove the `me` prefix from the apiClient

packages/wallets/src/evm/wallet.ts
```diff
@@ -26,48 +68,318 @@ export interface ViemWallet {
...
+        }
+    }
+
+    private get signerLocator(): string {
+        switch (this.adminSigner.type) {
+            case "evm-passkey":
+                return `evm-passkey:${this.adminSigner.id}`;
```
AlbertoElias, line 278: can't you just return `this.adminSigner.locator`?

packages/wallets/src/evm/wallet.ts
```diff
@@ -26,48 +68,318 @@ export interface ViemWallet {
...
+                    ],
+                });
+            }
+        }
+    }
+
+    private async approveSignature(signatureId: string, message: Hex) {
+        switch (this.adminSigner.type) {
+            case "evm-passkey": {
+                const { metadata, signature } = await WebAuthnP256.sign({
```
AlbertoElias, line 336: could we abstract all the signing logic as it's the same for both txs and signatures?

packages/wallets/src/openapi.json
```diff
@@ -1785,7 +1785,8 @@
                                             "viction",
                                             "mode",
                                             "space",
-                                            "soneium"
+                                            "soneium",
+                                            "story"
```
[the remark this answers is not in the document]
AlbertoElias: can we still not fetch via the url? and there are no autogen options that can fetch via url?

packages/wallets/src/sdk.ts
```diff
@@ -25,7 +20,7 @@ type WalletTypeToWallet = {
     "solana-mpc-wallet": SolanaMPCWallet;
 };
 
-class CrossmintWallet {
+export class CrossmintWallet {
```
AlbertoElias, line 23: does this not work server side for evm-mpc-wallet?

AlbertoElias (commented):
Overall looking great!! Left some comments and qs

packages/wallets/src/evm/wallet.ts
```diff
@@ -26,48 +68,318 @@ export interface ViemWallet {
...
     public async signMessage(parameters: {
         message: SignableMessage;
     }): Promise<Hex> {
-        throw new Error("Not implemented");
+        if (typeof parameters.message !== "string") {
+            throw new Error("Message must be a string");
+        }
+
+        // Create signature
+        const signatureCreationResponse = await this.apiClient.createSignature(this.walletLocator, {
```
AlbertoElias, line 121: We should consider abstracting these very long functions into more declarative utility functions with good naming so the whole flow is easier to follow without the need of comments

packages/wallets/src/utils/index.ts
```diff
@@ -0,0 +1,3 @@
+export async function sleep(ms: number) {
```
AlbertoElias, line 1: when calling sleep, we should consider the case where api calls might timeout or be rate limited

packages/wallets/src/evm/chains/index.ts
```diff
@@ -0,0 +1,68 @@
+import { BlockchainIncludingTestnet as Blockchain } from "@crossmint/common-sdk-base";
```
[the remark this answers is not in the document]
Destiner: yup there's `WAL-4166` for that.

Agree we should remove duplication as much as possible.

packages/wallets/src/evm/wallet.ts
Destiner: Mostly to simplify the output types.

packages/wallets/src/sdk.ts
Destiner: MPC wallets are not yet supported

packages/wallets/src/evm/wallet.ts
Destiner: good call https://github.com/Crossmint/crossmint-sdk/pull/970/commits/be3ae7f56923d1c2f6c399dab611f1ce858ce321

packages/wallets/src/evm/wallet.ts
Destiner: yes we can https://github.com/Crossmint/crossmint-sdk/pull/970/commits/4dc9c7bad11149114862754c9680a998f048c80b

packages/wallets/src/evm/wallet.ts
Destiner: yup https://github.com/Crossmint/crossmint-sdk/pull/970/commits/ff08c0be2000caee5452b6562d1c0e9e6a942c52

packages/wallets/src/evm/chains/definitions/storyTestnet.ts
```diff
@@ -0,0 +1,21 @@
+import { defineChain } from "viem";
```
alberto-crossmint, line 1: These chains are defined on latest viem. Do we need the current version or could we upgrade?

Destiner: I've tried to upgrade, but I think it requires upgrading the entire monorepo.

But in general I think we're safe to upgrade to the latest version.

packages/wallets/src/utils/index.ts
Destiner: 👍

We should improve the error handling in general (the way we handle errors and the way we expose them).

Adding to the backlog as it's non-blocking.

packages/wallets/src/evm/wallet.ts
AlbertoElias: but we're fetching the transactions for that specific walletLocator, so it won't return solana txs as it's not a solana wallet

packages/wallets/src/evm/wallet.ts
```diff
@@ -26,48 +71,288 @@ export interface ViemWallet {
...
+        } else {
+            return `me:evm-smart-wallet`;
+        }
+    }
+
+    private get signerLocator(): string {
+        return this.adminSigner.locator;
+    }
+
+    private async generateSignature(message: Hex): Promise<{ signature: Hex; metadata?: WebAuthnP256.SignMetadata }> {
```
alberto-crossmint, line 282: Maybe signWithAdminSigner? generateSignature looks way too similar to the function to create the signature. Also we should be clear about that is not the wallet who signs that message but the admin itself

alberto-crossmint (approved):

packages/wallets/src/evm/wallet.ts
Destiner: This is about typing. The `apiClient.getTransactions` doesn't discriminate the output type based on the wallet locator type. Could probably add the logic to the `ApiClient`

packages/wallets/src/evm/wallet.ts
```diff
@@ -26,48 +71,288 @@ export interface ViemWallet {
...
+        if (signature === undefined) {
+            throw new Error("Signature not available");
+        }
+
+        // Get signature status until success
+        let signatureResponse: GetSignatureResponse | null = null;
+        while (signatureResponse === null || signatureResponse.status === "pending") {
+            await sleep(STATUS_POLLING_INTERVAL_MS);
+            signatureResponse = await this.apiClient.getSignature(this.walletLocator, signatureId);
+        }
```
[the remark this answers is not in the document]
Destiner: https://github.com/Crossmint/crossmint-sdk/pull/970/commits/8da25c4d5ac7d8991449bc2e06b9c1435dde9084

packages/wallets/src/evm/wallet.ts
Destiner: i like it https://github.com/Crossmint/crossmint-sdk/pull/970/commits/fa63bd2d2cdc04351695393d21a3d5e0bf46b717

PR #971, opened 2025-03-06
Smart Wallet React SDK: fix wallet persistence & passkey prompts

Destiner:
## Description

Fixes `@crossmint/client-sdk-react-ui` by storing the wallet state in LocalStorage. Also uses newly added wallet action callbacks from `@crossmint/client-sdk-smart-wallet` to render passkey prompts.

Passkey callbacks and wallet persistence were removed from `@crossmint/client-sdk-smart-wallet` in #966 to reduce its scope. Removing that from the TS SDK was intentional, but this broke some of the functionality in React SDK.

This PR ports that functionality into the React SDK. This makes the passkey prompts being rendered again, and persists the wallet state across browser sessions.

## Test plan

* CI passes
* Lack of breaking changes
* Using React SDK should persist the wallet state and render the prompt
  * Can test by running the `apps/wallets/smart-wallet/next` playground app

## Package updates

```
"@crossmint/client-sdk-smart-wallet": minor
"@crossmint/client-sdk-react-ui": patch
```

### DIFF
--- a/packages/client/ui/react-ui/package.json
+++ b/packages/client/ui/react-ui/package.json
@@ -42,6 +42,7 @@
         "color": "4.2.3",
         "input-otp": "1.2.4",
         "lodash.isequal": "4.5.0",
+        "ox": "0.6.9",
         "react-jss": "10.10.0",
         "tailwind-merge": "2.4.0",
         "tailwindcss": "3.4.10",
@@ -49,12 +50,12 @@
         "zod": "3.22.4"
     },
     "devDependencies": {
-        "pino-pretty": "^7.6.0",
         "@testing-library/react": "^16.0.1",
         "@types/color": "4.2.0",
         "@types/lodash.isequal": "4.5.6",
         "@types/react": "^18.3.0",
         "@types/react-dom": "^18.3.0",
+        "pino-pretty": "^7.6.0",
         "react": "^18.3.0",
         "react-dom": "^18.3.0"
     },
--- a/packages/client/ui/react-ui/src/hooks/useWalletCache.ts
+++ b/packages/client/ui/react-ui/src/hooks/useWalletCache.ts
@@ -1,4 +1,4 @@
-import { useCallback } from "react";
+import { useCallback, useMemo } from "react";
 import type { Address } from "viem";
 import type { WebAuthnP256 } from "ox";
 import type { PasskeySigner } from "@crossmint/client-sdk-smart-wallet";
@@ -42,10 +42,6 @@ export function useWalletCache(userId: string | undefined) {
         [userId]
     );
 
-    const getWalletAddress = useCallback((): Address | undefined => {
-        return getCache()?.wallet?.address;
-    }, [getCache]);
-
     const setWalletAddress = useCallback(
         (address: Address) => {
             const currentCache = getCache() || {};
@@ -60,7 +56,7 @@ export function useWalletCache(userId: string | undefined) {
         [getCache, setCache]
     );
 
-    const isWalletInitialized = useCallback((): boolean => {
+    const isWalletInitialized = useMemo((): boolean => {
         const cache = getCache();
         if (cache == null) {
             return false;
@@ -85,7 +81,7 @@ export function useWalletCache(userId: string | undefined) {
         [getCache, setCache]
     );
 
-    const getPasskey = useCallback((): PasskeySigner | undefined => {
+    const passkey = useMemo(() => {
         const cache = getCache();
         if (cache === null || cache.passkey === undefined) {
             return undefined;
@@ -122,11 +118,10 @@ export function useWalletCache(userId: string | undefined) {
     );
 
     return {
-        getWalletAddress,
         setWalletAddress,
         isWalletInitialized,
         setWalletInitialized,
-        getPasskey,
+        passkey,
         setPasskey,
     };
 }
--- a/packages/client/ui/react-ui/src/hooks/useWalletCache.ts
+++ b/packages/client/ui/react-ui/src/hooks/useWalletCache.ts
@@ -1,0 +1,132 @@
+import { useCallback } from "react";
+import type { Address } from "viem";
+import type { WebAuthnP256 } from "ox";
+import type { PasskeySigner } from "@crossmint/client-sdk-smart-wallet";
+
+type WalletCache = {
+    wallet?: {
+        address: Address;
+        initialized?: boolean;
+    };
+    passkey?: {
+        authenticatorId: string;
+        domain: string;
+        passkeyName: string;
+        pubKeyPrefix: number;
+        pubKeyX: string;
+        pubKeyY: string;
+    };
+};
+
+export function useWalletCache(userId: string | undefined) {
+    const getCache = useCallback((): WalletCache | null => {
+        if (userId == undefined) {
+            return null;
+        }
+        const key = `smart-wallet-${userId}`;
+        const cache = localStorage.getItem(key);
+        if (cache == null) {
+            return null;
+        }
+        return JSON.parse(cache) as WalletCache;
+    }, [userId]);
+
+    const setCache = useCallback(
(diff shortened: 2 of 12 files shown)


packages/client/wallets/smart-wallet/src/smartWalletService.ts
```diff
@@ -56,7 +58,17 @@ export interface UserParams {
     jwt: string;
 }
 
+export interface Callbacks {
+    onWalletCreationStarted?: () => Promise<void>;
```
Destiner, line 62: See previously defined callbacks [here](https://github.com/Crossmint/crossmint-sdk/commit/98d819729a917b88d333c5574b8726455681fe12#diff-c5100b9aa596950b4758a14a50a50f71c64f8b742801b6d3ea98e4f0a24a7febL107-L110)

I've generilzed them a bit (e.g. made them available for all signer types except passkeys only) to enable more use cases!

packages/client/wallets/smart-wallet/src/smartWalletService.ts
```diff
@@ -69,8 +81,10 @@ export class SmartWalletService {
     public async getOrCreate(
         user: UserParams,
         chain: SmartWalletChain,
-        walletParams: WalletParams
+        walletParams: WalletParams,
+        callbacks?: Callbacks
```
Destiner, line 85: Adding an optional `callbacks` param to prevent breaking changes

packages/client/ui/react-ui/package.json
```diff
@@ -42,19 +42,20 @@
         "color": "4.2.3",
         "input-otp": "1.2.4",
         "lodash.isequal": "4.5.0",
+        "ox": "0.6.9",
```
Destiner, line 45: Need this for passkeys

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -76,13 +95,15 @@ export function CrossmintWalletProvider({
...
-        const signer = await smartWalletSDK.createPasskeySigner("Crossmint Wallet");
+    const createPasskeySigner = async (name: string) => {
+        await createPasskeyPrompt("create-wallet")();
+        const signer = await smartWalletSDK.createPasskeySigner(name);
         setPasskeySigner(signer);
         return signer;
     };
 
     const getOrCreateWallet = async (config?: WalletConfig) => {
+        console.log("getOrCreateWallet 1", config);
```
AlbertoElias, line 106: leftover?

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -116,10 +156,116 @@ export function CrossmintWalletProvider({
...
+                    pubKeyX: passkeySigner.credential.publicKey.x.toString(),
+                    pubKeyY: passkeySigner.credential.publicKey.y.toString(),
+                },
+            });
+            return passkeySigner;
+        }
+        return {
+            type: "PASSKEY",
+            credential: {
+                id: cache.passkey.authenticatorId,
```
AlbertoElias, line 204: if you just created a passkey, wouldn't `cache.passkey.authenticatorId` be null as `cache` for fetched before calling `updateCache`?

AlbertoElias (commented):
were we using localstorage before? what's the benefit of adding it?

Destiner:
@AlbertoElias yes, LocalStorage [was used](https://github.com/Crossmint/crossmint-sdk/blob/cefc9e87e8d07f577857fa3c8c753bac7bb32ce8/packages/client/wallets/smart-wallet/src/blockchain/wallets/account/cache.ts) in the previous version of TS SDK.

The idea is to cache the signer/wallet configuration for the user to prevent the initial API call on startup.

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
Destiner: Assuming you mean the `if (cache === null || cache.passkey === undefined) {` condition just above that, there's an early return that outputs the newly created passkey.

In other words, if we've just created a passkey, this code won't be reached.

AlbertoElias: ah missed that!

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -44,10 +46,25 @@ type WalletContext = {
     getOrCreateWallet: (
         config?: Pick<WalletConfig, "signer" | "type">
     ) => Promise<{ startedCreation: boolean; reason?: string }>;
-    createPasskeySigner: () => Promise<PasskeySigner | null>;
+    createPasskeySigner: (name: string) => Promise<PasskeySigner | null>;
     clearWallet: () => void;
 };
 
+type WalletCache = {
```
AlbertoElias, line 53: We should extract all new cache and passkey functionality as this provider is already getting too big and that way we can make the provider itself more declarative and easier to read

Destiner: @AlbertoElias agree the file became too large.

Moved all the persistance stuff into a new `useWalletCache` hook.

packages/client/ui/react-ui/src/hooks/useWalletCache.ts
```diff
@@ -0,0 +1,132 @@
...
+                wallet: {
+                    ...currentCache.wallet,
+                    initialized,
+                },
+            });
+        },
+        [getCache, setCache]
+    );
+
+    const getPasskey = useCallback((): PasskeySigner | undefined => {
```
AlbertoElias, line 88: why does this need to be a function instead of a state?

AlbertoElias: Just a minor comment btw, I was thinking the sets could also save the state to both localstorage and an in memory state, but could be a silly optimization. It would make the code slightly cleaner as you could just fetch the values as is without getters. But leave it up to you

AlbertoElias (approved):

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -99,6 +83,7 @@ export function CrossmintWalletProvider({
         await createPasskeyPrompt("create-wallet")();
         const signer = await smartWalletSDK.createPasskeySigner(name);
         setPasskeySigner(signer);
```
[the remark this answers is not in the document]
Destiner: nice catch, fixed https://github.com/Crossmint/crossmint-sdk/pull/971/commits/0b57aa7c542e8b5f1920d31a1253f6676b55824c

packages/client/ui/react-ui/src/hooks/useWalletCache.ts
Destiner: you're right about function vs variable, rewrote with `useMemo` in https://github.com/Crossmint/crossmint-sdk/pull/971/commits/a1cd119ef6235485d7463dfe164784810816aab5

AlbertoElias: love this!

Destiner: Not sure about merging localstorage and memory state. There's already `useWallet`, so changing `useWalletCache` to `useWalletState` or something like that would be confusing IMO. It's possible to do a bigger rewrite, but that will take more time and can introduce more bugs.

Probably doesn't make sense since this package will be deprecated/rewritten anyway.

AlbertoElias: yeah, useMemo is more than enough!

Issue #18, opened 2022-01-24
[feature] Send client SDK name and version with all requests


alfonso-paella:
This will help troubleshoot and debug

I think typically the standard way to do this is to set the HTTP headers:
* X-Client-Version: sdk version
* X-Client-Name: client name

But we should do some research to confirm

pgarciaegido:
Hey @alfonso-paella ,

I have a few questions about this one:

1. When we mean all requests we mean the first GET request we make to the `/api/crossmint/onboardingRequests/${clientId}/status` endpoint, but also to the following requests we do in the popup window, correct? Im not sure we can directly pass he headers using the client API `window.open`. We probably need to pass them via query params at first and find a way to do it in next(https://nextjs.org/docs/api-reference/next.config.js/headers).
2. How would I access the client name? If im not mistaken, the only source of external information to the button is the `/status` endpoint (above mentioned).

alfonso-paella:
Re: 1

Yeah we can pass headers for the /status endpoint, and then pass the same info encoded in custom url params for the window.open. Maybe we can stop there for now, and down the line we can decide if we also need to transitively pass headers from the checkout ui to the crossbit-main backend

----------

re 2:
X-client-name should just be a hardcoded name we give the client SDK which is calling.

We can use the same names we use in npm, e.g. "client-sdk-react-ui" (and if we later have for vue, angular, android, iOS, ... we will use diff names for those)

alfonso-paella:
I think this is done @pgarciaegido

PR #976, opened 2025-03-11
api client: sdk meta headers

mPaella:
## Description

include sdk meta headers in all api client fetches

## Test plan
pretty straight forward

### DIFF
--- a/packages/common/base/src/apiClient/CrossmintApiClient.ts
+++ b/packages/common/base/src/apiClient/CrossmintApiClient.ts
@@ -42,8 +42,10 @@ export class CrossmintApiClient extends ApiClient {
         return ApiClient.normalizePath(baseUrl);
     }
 
-    get authHeaders() {
+    get commonHeaders() {
         return {
+            "x-sdk-name": this.internalConfig.sdkMetadata.name,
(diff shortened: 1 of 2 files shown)


packages/common/base/src/apiClient/CrossmintApiClient.ts
```diff
@@ -42,8 +42,10 @@ export class CrossmintApiClient extends ApiClient {
         return ApiClient.normalizePath(baseUrl);
     }
 
-    get authHeaders() {
+    get commonHeaders() {
         return {
+            "x-sdk-name": this.internalConfig.sdkMetadata.name,
+            "x-sdk-version": this.internalConfig.sdkMetadata.version,
```
alfonso-paella, line 48: i think we called these x-client-name, x-client-version in other places no?

alfonso-paella: cant find it in the code but see it referenced here

https://github.com/Crossmint/crossmint-sdk/issues/18

mPaella, line 48:
```suggestion
            "x-client-name": this.internalConfig.sdkMetadata.name,
            "x-client-version": this.internalConfig.sdkMetadata.version,
```

PR #985, opened 2025-03-18
Wallets SDK improvements

Destiner:
## Description

Essential improvements to the Wallets SDK to unlock React and RN work:

* Add `getWallet` function to get existing (server) wallets
* Add (optional) `options` param to define wallet action callbacks
* Add (optional) `experimental_sign` to the passkey signer config to override the message signing

## Test plan

* CI pass
* It should be possible to provide a custom signing function when using passkeys (useful for RN)
* It should be possible to define callbacks based on wallet/tx creation (useful for React SDK)

## Package updates

```
"@crossmint/wallets-sdk": patch
```

### DIFF
--- a/packages/wallets/src/evm/wallet.ts
+++ b/packages/wallets/src/evm/wallet.ts
@@ -51,6 +51,7 @@ export type EVMSignerInput =
               x: string;
               y: string;
           };
+          experimental_sign?: (message: string) => Promise<{ signature: Hex; metadata: WebAuthnP256.SignMetadata }>;
       };
 export type EVMSigner = EVMSignerInput & {
     locator: string;
@@ -290,6 +291,10 @@ export class EVMSmartWallet implements ViemWallet {
     private async signWithAdminSigner(message: Hex): Promise<{ signature: Hex; metadata?: WebAuthnP256.SignMetadata }> {
         switch (this.adminSigner.type) {
             case "evm-passkey": {
+                // Custom signing function
+                if (this.adminSigner.experimental_sign) {
+                    return this.adminSigner.experimental_sign(message);
+                }
                 const { metadata, signature } = await WebAuthnP256.sign({
                     credentialId: this.adminSigner.id,
                     challenge: message,
--- a/packages/wallets/src/evm/wallet.ts
+++ b/packages/wallets/src/evm/wallet.ts
@@ -29,6 +29,11 @@ export interface TransactionInput {
     value?: bigint;
 }
 
+export type PasskeySigningCallback = (
+    message: string
+) => Promise<{ signature: Hex; metadata: WebAuthnP256.SignMetadata }>;
+export type PasskeyCreationCallback = (name: string) => Promise<{ id: string; publicKey: { x: string; y: string } }>;
+
 export type EVMSignerInput =
     | {
           type: "evm-keypair";
@@ -46,8 +51,8 @@ export type EVMSignerInput =
     | {
           type: "evm-passkey";
           name?: string;
-          signingCallback?: (message: string) => Promise<{ signature: Hex; metadata: WebAuthnP256.SignMetadata }>;
-          creationCallback?: (name: string) => Promise<{ id: string; publicKey: { x: string; y: string } }>;
+          signingCallback?: PasskeySigningCallback;
+          creationCallback?: PasskeyCreationCallback;
       };
 export type EVMSigner = EVMSignerInput & {
     locator: string;
--- a/packages/wallets/src/evm/wallet.ts
+++ b/packages/wallets/src/evm/wallet.ts
@@ -47,6 +47,7 @@ export type EVMSignerInput =
           type: "evm-passkey";
           name?: string;
           signingCallback?: (message: string) => Promise<{ signature: Hex; metadata: WebAuthnP256.SignMetadata }>;
+          creationCallback?: (name: string) => Promise<{ id: string; publicKey: { x: string; y: string } }>;
       };
 export type EVMSigner = EVMSignerInput & {
     locator: string;
--- a/packages/wallets/src/evm/wallet.ts
+++ b/packages/wallets/src/evm/wallet.ts
@@ -45,17 +45,20 @@ export type EVMSignerInput =
       }
     | {
           type: "evm-passkey";
-          id: string;
-          name: string;
-          publicKey: {
-              x: string;
-              y: string;
-          };
-          experimental_sign?: (message: string) => Promise<{ signature: Hex; metadata: WebAuthnP256.SignMetadata }>;
+          name?: string;
+          signingCallback?: (message: string) => Promise<{ signature: Hex; metadata: WebAuthnP256.SignMetadata }>;
       };
 export type EVMSigner = EVMSignerInput & {
     locator: string;
-};
+} & (
+        | {
+              type: "evm-passkey";
+              id: string;
+          }
+        | {
+              type: "evm-keypair";
+          }
+    );
 
 export interface ViemWallet {
     getAddress: () => Address;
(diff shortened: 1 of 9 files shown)


packages/wallets/src/evm/wallet.ts
```diff
@@ -44,6 +51,7 @@ export type EVMSignerInput =
               x: string;
               y: string;
           };
+          experimental_sign?: (message: string) => Promise<{ signature: Hex; metadata: WebAuthnP256.SignMetadata }>;
```
Destiner, line 54: Prefixing as `experimental` to signal that the feature is early and unstable

packages/wallets/src/services/wallet-factory.ts
```diff
@@ -1,38 +1,61 @@
+import { PublicKey } from "@solana/web3.js";
```
Destiner, line 1: Did a refactor here to unify the code for `getOrCreateWallet` and `getWallet`

packages/wallets/src/solana/wallet.ts
```diff
@@ -85,11 +89,12 @@ export class SolanaSmartWallet extends SolanaWallet {
         apiClient: ApiClient,
         publicKey: PublicKey,
         adminSignerInput: SolanaSignerInput,
+        callback: Callbacks = {},
```
AlbertoElias, line 92: should it be called `callbacks` in plural?

packages/wallets/src/solana/wallet.ts
```diff
@@ -85,11 +89,12 @@ export class SolanaSmartWallet extends SolanaWallet {
...
         client: { public: Connection } = {
             public: getConnectionFromEnvironment(apiClient.environment),
         }
     ) {
-        super(apiClient, publicKey, client);
+        super(apiClient, publicKey, callback, client);
```
AlbertoElias, line 97: I feel like callbacks should be the last param

packages/wallets/src/utils/options.ts
```diff
@@ -0,0 +1,18 @@
...
+    onWalletCreationStart?: () => Promise<void>;
+    onWalletCreationComplete?: (wallet: Wallet) => Promise<void>;
+    onWalletCreationFail?: (error: Error) => Promise<void>;
+    onTransactionStart?: (transaction: Transaction) => Promise<void>;
+    onTransactionComplete?: (transaction: Transaction) => Promise<void>;
+    onTransactionFail?: (error: Error) => Promise<void>;
+}
+
+export interface WalletOptions {
+    experimental_callbacks?: Callbacks;
```
AlbertoElias, line 17: i dont think these should be experimental as we already had them

packages/wallets/src/services/wallet-factory.ts
```diff
@@ -1,38 +1,61 @@
...
     public async getOrCreateWallet<WalletType extends keyof WalletTypeToArgs>(
         type: WalletType,
-        args: WalletTypeToArgs[WalletType]
+        args: WalletTypeToArgs[WalletType],
+        options?: WalletOptions
+    ): Promise<WalletTypeToWallet[WalletType]> {
+        try {
+            const walletResponse = await this.createWallet(type, args, options);
+            const wallet = this.createWalletInstance(type, walletResponse, args, options);
+            await options?.experimental_callbacks?.onWalletCreationComplete?.(wallet);
```
AlbertoElias, line 23: feels weird to me to have `onWalletCreationStart` in one method, and the rest here

packages/wallets/src/services/wallet-factory.ts
```diff
@@ -53,25 +76,56 @@ export class WalletFactory {
...
+    private createWalletInstance<WalletType extends keyof WalletTypeToArgs>(
+        type: WalletType,
+        walletResponse: CreateWalletResponse,
+        args: WalletTypeToArgs[WalletType],
+        options?: WalletOptions
+    ): WalletTypeToWallet[WalletType] {
+        this.assertCorrectWalletType(walletResponse, type);
+
+        if (type === "evm-smart-wallet") {
+            const { chain, adminSigner } = args as WalletTypeToArgs["evm-smart-wallet"];
```
AlbertoElias, line 99: There's a bit of repeated code here and also this method ends up being way too long already, and will be even more once we support the other 5 wallet types we have. Can we abstract it?

AlbertoElias (commented):
Left a few comments, but looking good. I'm ambivalent about the `experimental` thing. As we're doing <1.0.0 releases, I'm happy to iterate on the API as we go and break things

packages/wallets/src/utils/options.ts
Destiner: I don't think callbacks in their current form have a valid use case outside of React SDK. I also think we could modify the React SDK to not rely on callbacks at all while keeping all the functionality. Finally, if we decide to keep them, we might remove some, or change the parameters we pass.

Destiner: It's not required as we're doing `0.x` release and can do breaking changes, but IMO it's a good DX to communicate we might remove/change that feature specifically!

packages/wallets/src/solana/wallet.ts
Destiner: nice catch https://github.com/Crossmint/crossmint-sdk/pull/985/commits/fb2d5bd15ba95fdb21757a7245dec828c474c2e8

packages/wallets/src/solana/wallet.ts
Destiner: makes sense https://github.com/Crossmint/crossmint-sdk/pull/985/commits/856a9eb2d24641276899382f9d0c392e88f28fcc

packages/wallets/src/services/wallet-factory.ts
Destiner: yeah I agree https://github.com/Crossmint/crossmint-sdk/pull/985/commits/e72d66c8275c0340e6f6c96f47c9ca02b8f8e2f5

packages/wallets/src/services/wallet-factory.ts
Destiner: Made a light refactor to handle much more VMs in the future easily!

https://github.com/Crossmint/crossmint-sdk/pull/985/commits/74f7f614690aca8bccef2531d3ea43e9e53ec00b

packages/wallets/src/sdk.ts
```diff
@@ -1,25 +1,56 @@
...
+    public async findWalletData<T extends WalletType>(
+        type: T,
+        address?: string
+    ): Promise<WalletTypeToWalletData[T] | null> {
+        const locator = this.apiClient.isServerSide ? address : `me:${type}`;
+        if (!locator) {
+            throw new Error("Address is required for server wallets");
+        }
+        const response = await this.apiClient.getWallet(locator);
+        // @ts-ignore
```
Destiner, line 47: Need to update `ApiClient` output types to accomodate for HTTP error responses

packages/wallets/src/utils/options.ts
AlbertoElias: If you have any suggestions on how to handle this without callbacks in the React SDK, would love to hear them!

AlbertoElias (approved):

packages/wallets/src/utils/options.ts
AlbertoElias: and that wrapper would live in the react sdk?

Destiner: Yup

AlbertoElias: That could work. I still feel callbacks are cleaner and easier to integrate if anybody else ends up needing them

PR #988, opened 2025-03-19
Wallets SDK: Small Fixes

jmderby:
## Description

PR fixes the following: 
- passing the adminSigner.name or fallback to existing
- nfts method - the pagination params order was incorrect
- fixes nfts method for solana
- renamed `createWallet` to `getOrCreateWalletInternal`
- only call onWalletCreationStart callback when actually creating a wallet

## Test plan

wallets-sdk works as expected without errors regarding signing, wallet.nfts(...), or wallet creation passkey prompts

## Package updates

@crossmint/wallets-sdk

### DIFF
--- a/packages/wallets/src/evm/utils.ts
+++ b/packages/wallets/src/evm/utils.ts
@@ -23,7 +23,7 @@ export function getEvmAdminSigner(
             return {
                 type: "evm-passkey",
                 id: responseSigner.id,
-                name: input.name,
+                name: input?.name ?? responseSigner.name,
                 locator: responseSigner.locator,
             };
     }
--- a/packages/wallets/src/services/wallet-factory.ts
+++ b/packages/wallets/src/services/wallet-factory.ts
@@ -20,8 +20,7 @@ export class WalletFactory {
         options?: WalletOptions
     ): Promise<WalletTypeToWallet[WalletType]> {
         try {
-            await options?.experimental_callbacks?.onWalletCreationStart?.();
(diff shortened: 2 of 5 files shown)


packages/wallets/src/evm/utils.ts
```diff
@@ -23,7 +23,7 @@ export function getEvmAdminSigner(
             return {
                 type: "evm-passkey",
                 id: responseSigner.id,
-                name: input.name,
+                name: input?.name ?? responseSigner.name,
```
jmderby, line 26: input.id is the only required param therefore, fallback to existing wallet signer name

Destiner (approved):

packages/wallets/src/services/wallet-factory.ts
```diff
@@ -53,6 +53,7 @@ export class WalletFactory {
             if (existingWallet && !existingWallet.error) {
                 return existingWallet;
             }
+            await options?.experimental_callbacks?.onWalletCreationStart?.();
```
AlbertoElias, line 56: can we move the other callbacks to this function as well?

Destiner: yeah and let's wrap the `await this.apiClient.createWallet` call so that we have those callbacks for every wallet type!

jmderby: going to follow this up later as this is a blocker for react sdk!

PR #990, opened 2025-03-20
chore: More tests for common/base/ package

wsdt:
# Changelog
[X] No dependencies added
[X] Tiny change to`common/base/src/types/utils.ts` to freeze the ReadOnly array at runtime to match the expected behaviour

## Unit tests
- Added  more unit tests to `packages/common/base/**/*`
- No changes to existing tests except of adding an additional boundary to `Should disallow when prefix is invalid`

Would love some feedback on this!

Personally have used CrossMint on couple of projects and love what you do. 

Hope you find this helpful! 

--

Subtle sale (sorry for the shameless promotion): 
**We help with your QA while you can focus on what you do best - pushing forward your roadmap.** 

- What we do: [QA Services](https://wavect.io/services/software-quality-assurance/)
- [Simple subscription at 750€ / week](https://app-eu1.hubspot.com/payments/XwYYTHbMgzRq?referrer=PAYMENT_LINK), cancel anytime and get last week refunded if not happy

test

### DIFF
--- a/packages/common/base/src/apiClient/ApiClient.test.ts
+++ b/packages/common/base/src/apiClient/ApiClient.test.ts
@@ -1,0 +1,135 @@
+import { describe, expect, test, vi, beforeEach } from "vitest";
+import { ApiClient } from "./ApiClient";
+
+// Mock fetch globally
+const mockFetch = vi.fn();
+global.fetch = mockFetch;
+
+class TestApiClient extends ApiClient {
+    get authHeaders(): HeadersInit {
+        return { "x-api-key": "test-key" };
+    }
+
+    get baseUrl(): string {
+        return "https://api.test.com";
+    }
+}
+
+describe("ApiClient", () => {
+    let client: TestApiClient;
+
+    beforeEach(() => {
+        client = new TestApiClient();
+        mockFetch.mockClear();
+    });
+
+    describe("buildUrl", () => {
+        test("should correctly build URL with normalized paths", () => {
+            const url = client.buildUrl("/test/path");
+            expect(url).toBe("https://api.test.com/test/path");
+        });
+
+        test("should handle paths with leading and trailing slashes", () => {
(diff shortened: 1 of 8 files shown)


packages/common/base/src/types/utils.ts
```diff
@@ -1,4 +1,4 @@
 export type ObjectValues<T extends object> = T[keyof T];
 export function objectValues<T extends object>(obj: T): ReadonlyArray<T[keyof T]> {
-    return Object.values(obj);
+    return Object.freeze(Object.values(obj));
```
wsdt, line 3: Freeze object over just relying on "type safety" to match the expected behaviour

packages/common/base/vitest.config.ts
```diff
@@ -0,0 +1,14 @@
+import { defineConfig } from "vitest/config";
+import path from "path";
+
+export default defineConfig({
```
wsdt, line 4: Necessary to import  `@/apiKey` from within the test environment.

packages/common/base/src/apiClient/ApiClient.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    });
+
+    describe("buildUrl", () => {
+        test("should correctly build URL with normalized paths", () => {
+            const url = client.buildUrl("/test/path");
+            expect(url).toBe("https://api.test.com/test/path");
+        });
+
+        test("should handle paths with leading and trailing slashes", () => {
+            const url = client.buildUrl("/test/path/");
```
wsdt, line 33: Note: using `//test/path/` for example results in `https://api.test.com//test/path` which likely is expected behaviour. But I thought I would mention it here.

wsdt:
@AlbertoElias @jmderby @devlyn37 @afeight @pgarciaegido @mPaella

PR #994, opened 2025-03-20
Wallets SDK: fix code gen script

jmderby:
## Description

Fix our build script to run the code generation stuff rather than postinstall (which gets triggered in ANY subsequent projects using the sdk)

Error from attempting to run `pnpm i` in our smarter wallets demo repo
Screenshot 2025-03-20 at 2 12 08 PM

## Test plan

ensure original intent of generating codegen still works and that this doesn't cause side effects

## Package updates

@crossmint/wallets-sdk

### DIFF
--- a/packages/wallets/package.json
+++ b/packages/wallets/package.json
@@ -15,7 +15,7 @@
     "types": "./dist/index.d.ts",
     "files": ["dist", "LICENSE"],
     "scripts": {
-        "build": "tsup && pnpm generate",
+        "build": "pnpm generate && tsup",
         "dev": "tsup --watch",
         "test:vitest": "vitest run",
         "openapi-ts": "openapi-ts",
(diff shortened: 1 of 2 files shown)


packages/wallets/package.json
```diff
@@ -15,11 +15,11 @@
...
     "files": ["dist", "LICENSE"],
     "scripts": {
-        "build": "tsup",
+        "build": "pnpm generate && tsup",
         "dev": "tsup --watch",
         "test:vitest": "vitest run",
-        "generate": "pnpm dlx @hey-api/openapi-ts",
-        "postinstall": "pnpm generate"
+        "openapi-ts": "openapi-ts",
+        "generate": "pnpm run openapi-ts"
```
AlbertoElias, line 22: i think we are supposed to be using hey though

jmderby: still uses hey-api, this is what the docs say how to do it  https://www.npmjs.com/package/@hey-api/openapi-ts#:~:text=%40hey%2Dapi/openapi%2Dts%20via%20CLI

AlbertoElias (approved):

PR #1009, opened 2025-03-26
Wallets SDK: improve error handling

Destiner:
## Description

- Makes errors type-safe.
- Handles API errors gracefully.
- Removes the `removeDelegatedSigner` method
- Adds missing EVM chains (Mode, Mode Testnet, BSC)
- Validates provided `adminSigner` config (e.g. signer address / type) against the API response when getting an existing wallet

## Test plan

- Type-safe errors are thrown
- API errors are exposed to the developer
- Able to use with Mode and BSC
- Throws on admin signer mismatch

## Package updates

```
"@crossmint/wallets-sdk": minor
"@crossmint/client-sdk-base": patch
"@crossmint/common-sdk-base": patch
```


packages/wallets/src/solana/services/delegated-signers-service.ts
```diff
@@ -26,9 +26,4 @@ export class SolanaDelegatedSignerService {
         const response = await this.apiClient.getSigner(this.walletLocator, signer);
         return response;
     }
-
-    // biome-ignore lint/suspicious/useAwait: <explanation>
-    public async removeDelegatedSigner(_address: string): Promise<void> {
-        throw new Error("Not implemented");
-    }
```
Destiner, line 33: Removing as it's not implemented and wont be anytime soon.

Technically a breaking change

packages/wallets/src/evm/wallet.ts
Destiner, line 1: Big refactoring here, I think this module is much cleaner now.

packages/wallets/package.json
```diff
@@ -23,6 +23,7 @@
         "generate:docs": "typedoc"
     },
     "dependencies": {
+        "@crossmint/client-sdk-base": "workspace:*",
```
Destiner, line 26: Needed for wallet error codes

packages/wallets/src/services/wallet-factory.ts
```diff
@@ -173,12 +189,75 @@ export class WalletFactory {
         }
     }
 
+    /**
```
Destiner, line 192: Here, we validate against an existing wallet config:
1) Making sure the signer type is the same 
2) For keypair signers, making sure the signer address is the same

packages/wallets/src/api/client.ts
```diff
@@ -5,44 +5,61 @@ import { SDK_NAME, SDK_VERSION } from "../utils/constants";
...
     DelegatedSignerDto,
+    WalletsV1ControllerCreateDelegatedSigner4Error,
+    WalletsV1ControllerGetDelegatedSigner4Error,
 } from "./gen/types.gen";
+import { InvalidApiKeyError } from "../utils/errors";
 
 type CreateWalletParams = CreateWalletDto;
-type CreateWalletResponse = WalletV1Alpha2ResponseDto;
-type GetWalletResponse = WalletV1Alpha2ResponseDto;
+type GetWalletSuccessResponse = WalletV1Alpha2ResponseDto;
```
AlbertoElias, line 37: i think this is getting long enough where it would make more sense in a separate types file

packages/wallets/src/solana/services/delegated-signers-service.ts
AlbertoElias: we are actually adding this soon

packages/wallets/src/services/wallet-factory.ts
```diff
@@ -173,12 +189,75 @@ export class WalletFactory {
...
+     * @param existingWallet Onchain wallet state
+     * @param args Wallet config passed during wallet initialization
+     */
+    private checkWalletConfig<WalletType extends keyof WalletTypeToArgs>(
+        existingWallet: GetWalletSuccessResponse,
+        args: WalletTypeToArgs[WalletType]
+    ) {
+        switch (existingWallet.type) {
+            case "evm-smart-wallet": {
+                const { adminSigner: adminSignerInput } = args as WalletTypeToArgs["evm-smart-wallet"];
```
AlbertoElias, line 204: the code is the same for keypairs so we can unify for both evm and solana smart wallets

AlbertoElias (approved):
Left some small comments, but looking way better now!

packages/wallets/src/api/client.ts
Destiner: agree https://github.com/Crossmint/crossmint-sdk/pull/1009/commits/c78f549f4fc8dd77791b05509721e0e1c479ee07

packages/wallets/src/services/wallet-factory.ts
Destiner: https://github.com/Crossmint/crossmint-sdk/pull/1009/commits/18f1956a88869cb63e2d2d82d173db4a1be12230

packages/wallets/package.json
Destiner: hmm that would require moving the `CrossmintSDKError` from `client-base` to `common-base` as well

Destiner: looks good https://github.com/Crossmint/crossmint-sdk/pull/1009/commits/b2cf7ea221287e05e7914ed05270ab8ca185d990

PR #1023, opened 2025-04-01
Web3 Auth: Support EOA wallet adminSigner

jmderby:
### Overview

Adds ability to sign in using a eoa wallet (solana/evm) and use that wallet as a adminSigner for a non-custodial crossmint smart wallet.

uses our existing wallet connector via Dynamic 

### Smarter wallets demo
demo has been updated to reflect all changes

## Test plan

Test web3 login, signing, ensured all features are added in a backwards compatible way

## Package updates

@crossmint/client-sdk-auth


packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -94,6 +94,8 @@ export function CrossmintWalletProvider({
     const [passkeyPromptState, setPasskeyPromptState] = useState<PasskeyPromptState>({ open: false });
 
     const getOrCreateWallet = async (props: GetOrCreateWalletProps) => {
+        console.log("getOrCreateWallet", props);
```
jmderby, line 97: this log helps us debug (if necessary) tremendously

packages/client/auth/src/CrossmintAuthClient.ts
```diff
@@ -221,16 +221,12 @@ export class CrossmintAuthClient extends CrossmintAuth {
         }
     }
 
-    public async signInWithSmartWallet(address: string) {
+    public async signInWithSmartWallet(address: string, type: "ethereum" | "solana") {
         try {
-            const queryParams = new URLSearchParams({ signinAuthenticationMethod: "evm" });
```
jmderby, line 226: this query param is not currently being used in server, safe to remove...

apps/wallets/smart-wallet/next/src/app/_lib/providers.tsx
```diff
@@ -44,7 +50,13 @@ function CrossmintProviders({ children }: { children: ReactNode }) {
                     },
                 }}
                 authModalTitle="Sign in to Wallet Demo"
-                loginMethods={["google", "email", "farcaster", "twitter"]}
+                loginMethods={[
+                    "google",
+                    "email",
+                    "farcaster",
+                    "twitter",
+                    web3LoginMethod as "web3" | "web3:evm-only" | "web3:solana-only",
```
AlbertoElias, line 58: we should type this and set web3LoginMethod to that type

packages/client/auth/src/CrossmintAuthClient.ts
```diff
@@ -244,10 +240,10 @@ export class CrossmintAuthClient extends CrossmintAuth {
         }
     }
 
-    public async authenticateSmartWallet(address: string, signature: string) {
+    public async authenticateSmartWallet(address: string, type: "ethereum" | "solana", signature: string) {
         try {
             const queryParams = new URLSearchParams({
-                signinAuthenticationMethod: "evm",
+                signinAuthenticationMethod: type === "ethereum" ? "evm" : "solana",
```
AlbertoElias, line 246: why not just make the type be `evm` instead of `ethereum`?

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -117,67 +123,78 @@ export function CrossmintAuthProvider({
...
             console.log("User not logged in");
             return;
         }
 
         const user = await crossmintAuth.getUser();
         setUser(user);
-    };
+    }, [crossmint.jwt, crossmintAuth]);
+
+    const authContextValue = useMemo(
```
AlbertoElias, line 182: do we need this useMemo?

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -216,41 +242,68 @@ export function CrossmintAuthProvider({
...
+    const canGetOrCreateWallet =
+        createOnLogin === "all-users" && walletStatus === "not-loaded" && accessToken != null && sdkHasLoaded;
 
-    const handleWalletCreation = useCallback(() => {
+    const handleWalletCreation = useCallback(async () => {
         if (!canGetOrCreateWallet) {
             return;
         }
+
+        let adminSigner: any = defaultAdminSigner;
```
AlbertoElias, line 269: can we avoid any?

packages/client/ui/react-ui/src/utils/mapSignerToWalletType.ts
```diff
@@ -0,0 +1,15 @@
+export function mapSignerToWalletType(type?: string): "evm-smart-wallet" | "solana-smart-wallet" | undefined {
```
AlbertoElias, line 1: dont we already have this in the wallets sdk? if so, maybe we can export there and re-use

packages/client/ui/react-ui/src/providers/auth/web3/DynamicWeb3WalletConnect.tsx
```diff
@@ -37,31 +50,40 @@ export function DynamicWeb3WalletConnect({
...
-                        console.log({ chain });
-
+                        const chain = await dynamicChainToCrossmintChain(args);
+                        const type = chain === "solana" ? "solana" : "ethereum";
                         try {
-                            const res = await crossmintAuth?.signInWithSmartWallet(address);
-                            const signature = (await wallet.signMessage?.(res.challenge)) as `0x${string}`;
-                            const authResponse = (await crossmintAuth?.authenticateSmartWallet(address, signature)) as {
+                            const res = await crossmintAuth?.signInWithSmartWallet(address, type);
+                            const rawSignature = (await args.connector?.proveOwnership(
```
AlbertoElias, line 64: this is a dynamic thing?

AlbertoElias (approved):
left some questions but lgtm overall

apps/wallets/smart-wallet/next/src/app/_lib/providers.tsx
jmderby: for sure, i'll just need to expose the type

packages/client/ui/react-ui/src/providers/CrossmintAuthProvider.tsx
jmderby: i'm generally against useMemo, but in this case I think it's appropriate to use it only because it's a fairly heavy context value obj and imo poses a large cascade of re-rendering throughout the comp tree

packages/client/auth/src/CrossmintAuthClient.ts
jmderby: i actually forgot to remove this completely 😅

jmderby: actually my bad, I update our client code to use "evm" but then updated it here. Reason being is stytch takes "ethereum" in our light /start endpoint wrapper

packages/client/ui/react-ui/src/providers/auth/web3/DynamicWeb3WalletConnect.tsx
jmderby: yes you bet dm me if you want more context on this, i'll share convo with you

packages/client/ui/react-ui/src/utils/mapSignerToWalletType.ts
jmderby: we've got WalletTypeToWallet but nothing light enough that just returns wallet type str

PR #1028, opened 2025-04-02
Solana sendTransaction: Fix signing for keypair

jmderby:
## Description

Fixed our signing mechanism for signing solana txns.

To provide some context here, according to Claude this is whats happening

> This is a security protection - Phantom (and other Solana wallets) deliberately prevent you from using the signMessage method to sign transaction data. This prevents potential security exploits where users think they're signing a message but are actually signing a transaction.

Therefore, from my understanding the following was necessary:

1. the server sends a serialized transaction message from the transaction
2. then, we need to convert that serialized transaction message back into a transaction and use the signTransaction function from our wallet provider (dynamic in our case i'm not 100 percent confident this will work with other providers but it should based on my research!)

## Test plan

Tested signing with wallet.sendTransaction and works

## Package updates

@crossmint/wallets-sdk

### DIFF
--- a/packages/wallets/src/solana/services/approvals-service.ts
+++ b/packages/wallets/src/solana/services/approvals-service.ts
@@ -1,4 +1,5 @@
 import bs58 from "bs58";
+import { VersionedMessage, VersionedTransaction } from "@solana/web3.js";
 import type { ApiClient, CreateTransactionSuccessResponse, SolanaWalletLocator } from "@/api";
 import type { SolanaNonCustodialSigner } from "../types/signers";
 import { PendingApprovalsError, InvalidSignerError, TransactionFailedError } from "../../utils/errors";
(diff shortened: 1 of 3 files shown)


packages/wallets/src/solana/services/approvals-service.ts
```diff
@@ -41,4 +50,21 @@ export class SolanaApprovalsService {
...
         return transaction;
     }
+
+    private retrieveValidSignature(signedTxn: VersionedTransaction) {
+        let signatureObj = null;
+        for (let i = 0; i < signedTxn.signatures.length; i++) {
+            const sig = signedTxn.signatures[i];
+            // Check if this signature is non-zero
+            if (Object.values(sig).some((byte) => byte !== 0)) {
+                signatureObj = sig;
```
AlbertoElias, line 60: so you're onlly keeping the last signature from `signedTxn.signatures`?

packages/wallets/src/solana/services/approvals-service.ts
```diff
@@ -24,8 +25,16 @@ export class SolanaApprovalsService {
                         `Signer ${approval.signer} is required for the transaction but was not found in the signer list`
                     );
                 }
+                // Convert the decoded transaction message from server back into a VersionedTransaction
```
jmderby, line 28: To provide some context here, according to Claude this is whats happening

> This is a security protection - Phantom (and other Solana wallets) deliberately prevent you from using the signMessage method to sign transaction data. This prevents potential security exploits where users think they're signing a message but are actually signing a transaction.

Therefore, from my understanding the following was necessary:
1. the server sends a serialized transaction message from the transaction 
2. then, we need to convert that serialized transaction message back into a transaction and use the signTransaction function from our wallet provider (dynamic in our case i'm not 100 percent confident this will work with other providers but it should based on my research!)

packages/wallets/src/solana/services/approvals-service.ts
```diff
@@ -41,4 +50,21 @@ export class SolanaApprovalsService {
         }
         return transaction;
     }
+
+    private retrieveValidSignature(signedTxn: VersionedTransaction) {
+        let signatureObj = null;
+        for (let i = 0; i < signedTxn.signatures.length; i++) {
```
AlbertoElias, line 56: you can also find the signature that corresponds to that signer like this: https://github.com/Paella-Labs/crossbit-main/pull/18179/files#diff-783c98152b8acbc0153ec1f0049113d0881919bbd1815cf3b0de98439f4e71ebR201

AlbertoElias (approved):

PR #1031, opened 2025-04-08
Wallets SDK: Standardize wallet return type

jmderby:
## Description

• Standardizes return value from getOrCreateWallet method, this ensures the return type matches the actual returned wallet instance
• Adds `get` prefix to all getter methods for consistency
• Adds getDelegatedSigners method to solana-smart-wallet

additional context: 

what we currently **have**: (similar applies to evm)
Screenshot 2025-04-07 at 12 20 20 PM

what we **want**:
Screenshot 2025-04-07 at 2 05 50 PM

## Test plan

Added additional automated test cases for evm and existing sol, manually tested wallets object, and type remain the same but return value is now correct.

## Package updates

@crossmint/wallets-sdk

### DIFF
--- a/packages/wallets/src/services/wallet-factory.ts
+++ b/packages/wallets/src/services/wallet-factory.ts
@@ -4,8 +4,8 @@ import type { Address } from "viem";
 import type { EvmWalletType, SolanaWalletType, WalletTypeToArgs, WalletTypeToWallet } from "./types";
 import type { ApiClient, CreateWalletResponse, GetWalletSuccessResponse } from "../api";
 import { createPasskeySigner, getEvmAdminSigner } from "../evm/utils";
-import { ISolanaMPCWallet, ISolanaSmartWallet, type SolanaMPCWallet, type SolanaSmartWallet } from "../solana";
-import { IEVMSmartWallet, type EVMSmartWallet } from "../evm";
+import { SolanaMPCWalletImpl, SolanaSmartWalletImpl, type SolanaMPCWallet, type SolanaSmartWallet } from "../solana";
+import { EVMSmartWalletImpl, type EVMSmartWallet } from "../evm";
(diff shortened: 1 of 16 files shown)


packages/wallets/src/solana/wallet.ts
```diff
@@ -106,7 +102,7 @@ abstract class SolanaWallet {
     }
 }
 
-export class SolanaSmartWallet extends SolanaWallet {
+export class ISolanaSmartWallet extends SolanaWallet implements SolanaSmartWallet {
```
AlbertoElias, line 105: `I` is used for interfaces, so a bit weird to use it for a class

packages/wallets/src/services/wallet-factory.ts
```diff
@@ -165,23 +176,42 @@ export class WalletFactory {
...
+                } as SolanaSmartWallet;
             }
             case "solana-mpc-wallet": {
-                return new SolanaMPCWallet(
+                const wallet = new ISolanaMPCWallet(
                     this.apiClient,
                     new PublicKey(solanaResponse.address),
                     getConnectionFromEnvironment(this.apiClient.environment),
                     options?.experimental_callbacks ?? {}
-                ) as WalletTypeToWallet[WalletType];
```
AlbertoElias, line 184: it didnt work by just doing `as SolanaSmartWallet`? All this binding adds a lot of boilerplate. Have you also tried `satisfies SolanaSmartWallet`?

AlbertoElias (approved):
I'll approve because i don't know a better, but I wish we could avoid so much binding. Left a few questions

packages/wallets/src/services/wallet-factory.ts
jmderby: update all to use satisfies

PR #1032, opened 2025-04-08
Auth: Adds support for custom storage solutions in auth sdk

AlbertoElias:
## Description

We need non-cookie options for other environments such as react native

## Test plan

Tested with expo app and with smarter wallet

## Package updates

@client-auth: minor

### DIFF
--- a/packages/client/auth/src/utils/storage.ts
+++ b/packages/client/auth/src/utils/storage.ts
@@ -1,3 +1,5 @@
+import { deleteCookie, getCookie, setCookie } from "./cookies";
+
 export interface StorageProvider {
     get(key: string): Promise<string | undefined>;
     set(key: string, value: string, expiresAt?: string): Promise<void>;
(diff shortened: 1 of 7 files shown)


packages/client/auth/src/utils/storage.ts
```diff
@@ -0,0 +1,36 @@
...
+        if (typeof document === "undefined") {
+            return;
+        }
+        document.cookie = `${key}=; expires=Thu, 01 Jan 1970 00:00:00 UTC; path=/;`;
+    }
+}
+
+export function getDefaultStorageProvider(): StorageProvider {
+    return new CookieStorage();
+}
```
afeight, line 36: I see the need for this now in order to not introduce a breaking change, but would ideally mark this as something to remove in the future. Might throw a comment saying that

AlbertoElias: why would this be temporary?

PR #1040, opened 2025-04-12
Delegate signers & balances: add types / errors

jmderby:
## Description

PR improves the typing around get delegate signers and get balances. Exposes two types that also get surfaced through the react-ui: `DelegatedSigner` and `WalletBalance`. 

## Test plan

Ensure tests pass, type names make sense.

## Package updates

@crossmint/wallets-sdk

### DIFF
--- a/packages/wallets/src/openapi.json
+++ b/packages/wallets/src/openapi.json
@@ -9009,6 +9009,80 @@
                                                 }
                                             }
                                         ]
+                                    },
+                                    "delegatedSigners": {
+                                        "type": "array",
+                                        "items": {
+                                            "type": "object",
+                                            "properties": {
+                                                "type": {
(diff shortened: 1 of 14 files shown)


packages/wallets/src/openapi.json
```diff
@@ -9009,6 +9009,80 @@
                                                 }
```
jmderby, line 9009: having to manually update this schema anytime main api changes is pita. a nice to have would be hit the api and retrieve the schema for all /wallets endpoints would be nice but might not be trivial to implement.

AlbertoElias: yeah, not a fan of our autogen tool because we can't attach it to a url. maybe a pre-commit script?

AlbertoElias (approved):

PR #1047, opened 2025-04-15
Use Public RPCs for `client-sdk-smart-wallet`

devlyn37:
## Description

We have some RPCs hardcoded we're treating as public that are remnants of an old architecture. Now that everything is routed through our APIs, the remaining places we do use public PRCs:
- To implement a `viem` interface and support `getNonce`, which calls a smart contract.
- Exposed as `wallet.client.public` from our `EVMSmartWallet` object
Expect little enough usage for default public RPCs from viem.

This is also what we do in our latest package `@crossmint/wallets-sdk`

Screenshot 2025-04-14 at 8 25 52 PM

## Test plan

Didn't test because this PR only:
- deletes unused code
- changes the inputs to `createPublicClient` from our own RPCs to default public RPCs from `viem`. I copied the code populate `createPublicClient` with default RPCs from `@crossmint/wallets-sdk`.

## Package updates

patch `client-sdk-smart-wallet`

### DIFF
--- a/packages/client/wallets/smart-wallet/src/evm/wallet.ts
+++ b/packages/client/wallets/smart-wallet/src/evm/wallet.ts
@@ -27,11 +27,11 @@ export class EVMSmartWallet {
             /**
              * An interface to interact with the smart wallet, execute transactions, sign messages, etc.
              */
-            public: PublicClient<HttpTransport>;
+            wallet: SmartWalletClient;
             /**
              * An interface to read onchain data, fetch transactions, retrieve account balances, etc. Corresponds to public [JSON-RPC API](https://ethereum.org/en/developers/docs/apis/json-rpc/) methods.
              */
-            wallet: SmartWalletClient;
+            public: PublicClient<HttpTransport>;
         },
         public readonly chain: SmartWalletChain,
         private readonly apiService: CrossmintApiService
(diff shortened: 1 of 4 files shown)


packages/client/wallets/smart-wallet/src/evm/wallet.ts
```diff
@@ -27,11 +27,11 @@ export class EVMSmartWallet {
             /**
              * An interface to interact with the smart wallet, execute transactions, sign messages, etc.
              */
-            public: PublicClient<HttpTransport>;
+            wallet: SmartWalletClient;
             /**
              * An interface to read onchain data, fetch transactions, retrieve account balances, etc. Corresponds to public [JSON-RPC API](https://ethereum.org/en/developers/docs/apis/json-rpc/) methods.
              */
-            wallet: SmartWalletClient;
+            public: PublicClient<HttpTransport>;
```
devlyn37, line 34: Descriptions were swapped here

packages/client/wallets/smart-wallet/src/smartWalletService.ts
```diff
@@ -85,16 +84,16 @@ export class SmartWalletService {
...
         const { signer } = walletParams;
         const walletResponse = await this.createWallet(user, signer);
         const address = walletResponse.address;
         return new EVMSmartWallet(
             {
                 wallet: this.smartAccountClient(user, signer, chain, address),
-                public: publicClient,
+                public: createPublicClient({
+                    chain: viemNetworks[chain],
+                    transport: http(),
```
devlyn37, line 95: `http()` uses the default chain rpc

PR #1055, opened 2025-04-16
Auth: Properly handle auth status in react native

AlbertoElias:
## Description

We weren't fetching jwt properly which made the status show as logged-out for a moment until everything was refreshed. And added an in-progress

## Test plan

Tested with expo app

## Package updates

react-native: patch

### DIFF
--- a/packages/client/ui/react-native/src/providers/CrossmintAuthProvider.tsx
+++ b/packages/client/ui/react-native/src/providers/CrossmintAuthProvider.tsx
@@ -1,7 +1,12 @@
 import { type ReactNode, createContext, useEffect, useMemo, useRef, useState, useCallback } from "react";
 import * as WebBrowser from "expo-web-browser";
 import { CrossmintAuth, type StorageProvider } from "@crossmint/client-sdk-auth";
-import type { AuthMaterialWithUser, SDKExternalUser, OAuthProvider } from "@crossmint/common-sdk-auth";
+import {
+    type AuthMaterialWithUser,
+    type SDKExternalUser,
+    type OAuthProvider,
+    SESSION_PREFIX,
+} from "@crossmint/common-sdk-auth";
 
 import { useCrossmint } from "../hooks";
 import { SecureStorage } from "../utils/SecureStorage";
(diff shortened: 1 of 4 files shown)


packages/client/ui/react-native/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -169,6 +183,7 @@ export function CrossmintAuthProvider({
 
     const loginWithOAuth = async (provider: OAuthProvider) => {
         try {
```
jcurbelo, line 185: we could add a `finally` block to ensure `setInProgress(false)` is called regardless of whether the try or catch block executes

```ts
const loginWithOAuth = async (provider: OAuthProvider) => {
        try {
            setInProgress(true);
            const oauthUrl =
                oauthUrlMap[provider] ?? (await crossmintAuth.getOAuthUrl(provider, { appSchema: resolvedAppSchema }));
            if (Platform.OS === "android") {
                await WebBrowser.openBrowserAsync(baseUrl.toString());
            } else {
                const result = await WebBrowser.openAuthSessionAsync(baseUrl.toString());
                if (result.type === "success") {
                    await createAuthSession(result.url);
                }
            }
            await WebBrowser.coolDownAsync();
        } catch (error) {
            console.error("[CrossmintAuthProvider] Error during OAuth login:", error);
        } finally { // Suggestion: Add `finally` block to ensure `setInProgress(false)` is called in all cases to prevent UI from being stuck in loading state
            setInProgress(false);
        }
    };
 ```

File: packages/client/ui/react-native/src/providers/CrossmintAuthProvider.tsx [hunk 2]
```diff
@@ -200,7 +215,10 @@ export function CrossmintAuthProvider({
                 : urlOrOneTimeSecret;
 
             if (oneTimeSecret != null) {
-                return await crossmintAuth.handleRefreshAuthMaterial(oneTimeSecret);
+                setInProgress(true);
```
jcurbelo, line 218: see https://github.com/Crossmint/crossmint-sdk/pull/1055/files#r2047670155 if we use `finally` block we might not need to handle `inProgress` state here

jcurbelo (approved):

packages/client/ui/react-native/src/providers/CrossmintAuthProvider.tsx
AlbertoElias: thing is, that this function finishes running doesnt mean it's no longer in progress. auth can be a 2-step process, and loginWithOauth doesnt finish until createAuthSession is called which may happen externally on android

jcurbelo: oh gotcha! makes sense!

PR #1069, opened 2025-04-23
NCS: Refactors window sdk to support webviews

AlbertoElias:
## Description

We are now working on mobile and interact with iframes in places such as a non-custodial signer and we want to make sure we can communicate back and forth properly there. I've refactored `client-window-sdk` to work with multiple transports and added the necessary pieces for a RN webview.

RN => Webview:

1. inject javascript that will call a handler function
2. the handler function dispatches a postmessage
3. the post message is picked up and replied

WebView => RN:

1. Calls RN custom `postMessage` function
2. We add an `onMessage` to the WebView that will pick up and reply.

## Test plan

Added a demo to the example expo app that can be tested in branch `albertoe/expo-ncs`

## Package updates

window-sdk: major


packages/client/window/src/EventEmitter.ts
```diff
@@ -51,49 +50,43 @@ export type OnActionArgs<
...
         const result = this.outgoingEvents[event].safeParse(data);
         if (result.success) {
-            if (Array.isArray(this.targetOrigin)) {
-                this.targetOrigin.forEach((origin) => {
-                    this.otherWindow?.postMessage({ event, data }, origin);
-                });
-            } else {
-                this.otherWindow?.postMessage({ event, data }, this.targetOrigin);
-            }
+            console.log("sending", event, data);
```
alberto-crossmint, line 66: Intentional?

packages/client/window/src/EventEmitter.ts
```diff
@@ -51,49 +50,43 @@ export type OnActionArgs<
...
+        const listener = (message: SimpleMessageEvent) => {
+            try {
+                if (message.data.event === event) {
+                    const data = this.incomingEvents[event].safeParse(message.data.data);
+                    if (data.success) {
+                        callback(data.data);
+                    } else {
+                        console.error("Invalid data for event", event, data.error);
+                    }
                 }
```
alberto-crossmint, line 83: I find this kind of dangerous, this component is integrated currently in the hosted checkout implementation and relies on events for errors, so any behavior change could break things. Do you think we can keep the existing behavior here?

packages/client/window/package.json
```diff
@@ -19,7 +19,18 @@
         "dev": "tsup --watch"
     },
     "dependencies": {
```
alberto-crossmint, line 21: I think that this package should be kept react-independent if possible. Adding react turns it into a way-bigger package, which may not make sense when used in other lightweight environments. It could complicate the integration of hosted checkout on some places because of dependencies conflicts. Do you think we could either
1. remove the react dependency while keeping the same functionality, or
2. Make the transport injectable when we build the window object so we can just pass this object from the parent application.

alberto-crossmint: Also, we're aiming to keep the iframe code as small as possible. The reason is that it's a critical part of our infrastructure and we're aiming at zero dependencies so we can build trust and maybe auditing the code and the hash, so adding react could complicate things. Also we try to prevent have any dynamic loading content there

alberto-crossmint: If that's hard to do I'd say we can
- create a experimental version of this package in npm.js that we use today for the demo, without merging it into the main repo
- Load that version in the iframe code 
- Investigate if this can be react independent the following days

alberto-crossmint: Do you think that makes sense?

alberto-crossmint (approved):

PR #1070, opened 2025-04-23
Common sdk auth: Updated twitter metadata type

jmderby:
## Description

Since we now expose the twitter username in oauth login method, we needed to update the types.

## Test plan

na

## Package updates

@crossmint/common-sdk-auth

### DIFF
--- a/packages/common/auth/src/types/user.ts
+++ b/packages/common/auth/src/types/user.ts
@@ -8,10 +8,14 @@ export type FarcasterMetadata = {
     verifications: string[];
 };
 
+export type TwitterMetadata = {
+    username: string;
+};
+
 export type SDKExternalUser = {
     id: string;
     email?: string;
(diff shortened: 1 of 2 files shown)


AlbertoElias (approved):

packages/common/auth/src/types/user.ts
```diff
@@ -8,10 +8,14 @@ export type FarcasterMetadata = {
     verifications: string[];
 };
 
+export type TwitterMetadata = {
+    username: string;
```
alfonso-paella, line 12: could we add the numeric id too in the future?

alfonso-paella (approved):

packages/common/auth/src/types/user.ts
jmderby: the twitter id i.e. 1867634432334577664 ?

PR #1085, opened 2025-04-28
RN: fixes creating multiple same handlers

AlbertoElias:
## Description

We were creating a global event handler for every window handler, but we only need a single global handler.

## Test plan

will test with recovery key iframe once deployed

## Package updates

rn-window: patch


devlyn37 (approved):
Take my approval w/ a grain of salt since I'm not super familiar, but looks good to me and claude

PR #1087, opened 2025-04-28
RN: fixes creating multiple same handlers #1085

AlbertoElias:
## Description

We were creating a global event handler for every window handler, but we only need a single global handler.

## Test plan

will test with recovery key iframe once deployed

## Package updates

rn-window: patch


devlyn37 (approved):
lgtm

PR #1091, opened 2025-04-29
NCS: Small fixes to support quickstart

AlbertoElias:
## Description

Some fixes that were missing for the quickstart:

- Better status to handle the full flow
- Fix a bad import
- Release rn-window

## Test plan

Tested with the expo app

## Package updates

react-native: patch
rn-window: patch


jmderby (approved):
nice! naming is way better now

PR #1093, opened 2025-04-30
NCS: use raw Ed25519 signing for NCSs

alberto-crossmint:
## Description

These events are being deprecated, as they need the Solana specific logic on the iframe side which increases the static bundle size and increases the attack surface. We're moving to a minimal signing logic implementation on the iframe and moving all the serialization specific logic to the SDK instead!

## Test plan

Tested in the live application https://simple-ncs-demo.vercel.app/

### DIFF
--- a/.changeset/chilled-weeks-occur.md
+++ b/.changeset/chilled-weeks-occur.md
@@ -1,0 +1,7 @@
+---
+"@crossmint/client-sdk-react-native-ui": minor
+"@crossmint/client-sdk-react-ui": minor
+"@crossmint/client-signers": patch
+---
+
+Unifying Solana NCS signatures
--- a/.changeset/chilled-weeks-occur.md
+++ b/.changeset/chilled-weeks-occur.md
@@ -1,6 +1,6 @@
 ---
-"@crossmint/client-sdk-react-native-ui": minor
-"@crossmint/client-sdk-react-ui": minor
+"@crossmint/client-sdk-react-native-ui": patch
+"@crossmint/client-sdk-react-ui": patch
 "@crossmint/client-signers": patch
 ---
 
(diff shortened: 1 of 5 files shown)


.changeset/chilled-weeks-occur.md
```diff
@@ -0,0 +1,7 @@
+---
+"@crossmint/client-sdk-react-native-ui": minor
+"@crossmint/client-sdk-react-ui": minor
```
AlbertoElias, line 3: We're doing patch versions in react libraries for all the ncs updates. We also do a separate changeset file for each library to manage the changelogs differently. But if it's the same update for all, no worries

packages/client/ui/react-native/src/providers/CrossmintRecoveryKeyProvider.tsx
```diff
@@ -175,14 +179,15 @@ export function CrossmintRecoveryKeyProvider({
...
                             const response = await parent.sendAction({
-                                event: "request:sign-transaction",
-                                responseEvent: "response:sign-transaction",
+                                event: "request:sign",
+                                responseEvent: "response:sign",
                                 data: {
                                     authData: { jwt, apiKey },
                                     data: {
-                                        transaction: bs58.encode(transaction.serialize()),
-                                        chainLayer: "solana",
```
AlbertoElias, line 185: Why do we remove chain layer? Can't we only have the chain layer and then the frame decides which type of signature it needs?

AlbertoElias (approved):

packages/client/ui/react-native/src/providers/CrossmintRecoveryKeyProvider.tsx
alberto-crossmint: Hmm that's a fair point! But I think sticking to the signature type may be safer because there are chains in which multiple signer types are allowed (i.e. Sui). Also that way the signers gets simplified by just knowing about raw cryptographic payloads and we don't have the temptation to have chain-specific behavior there

AlbertoElias: I think this is okay as long as these apis never get exposed to 3rd party developers, as that would add an unnecessary layer of complication. But as wallets or react sdk will wrap this, we're okay

PR #1094, opened 2025-04-30
NCS: make iframe URL configurable

alberto-crossmint:
## Description

We need to change to the render deployment, the AWS one keeps failing randomly

### DIFF
--- a/packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
+++ b/packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
@@ -30,7 +30,11 @@ type PasskeyPromptState =
 type ValidWalletState =
     | { status: "not-loaded" | "in-progress" }
     | { status: "loaded"; wallet: EVMSmartWallet; type: "evm-smart-wallet" }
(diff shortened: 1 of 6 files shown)


packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -84,11 +88,13 @@ export function CrossmintWalletProvider({
     showPasskeyHelpers = true,
     appearance,
     experimental_enableRecoveryKeys = false,
+    experimental_signersURL = undefined,
```
jmderby, line 91: don't forget to also add this to the CrossmintAuthProvider so we can pass that there too

jmderby (approved):
just one comment, but otherwise looks good!

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
AlbertoElias: shouldnt it be in the wallet provider? i think the auth provider should be completely independent of this flow

PR #1100, opened 2025-05-02
React sdk POC: refactor WalletProvider and AuthProvider hierarchy

jmderby:
## Description

I've put together a [google doc](https://docs.google.com/document/d/1FqpIhqZqdRkVu128oSrZFWYQzpCO4klCxOLUKAU4j8s/edit?tab=t.0#heading=h.z2nkiuhq73s3) for for more context but it might be less confusing to just review this PR 😅 

This pr is a POC to demonstrate the possibilities around supporting our new decoupled standard for CrossmintAuthProvider and CrossmintWalletProvider.

This pr contains the necessary changes to support our desired design while still maintaining backwards compatibility. This would of course be a major version bump because developers would need to use the wallet provider for any app and then opt-into auth via CrossmintAuthProvider. 

## Test plan

na

## Package updates


packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -85,12 +109,14 @@ export const WalletContext = createContext<WalletContext>({
 
 export function CrossmintWalletProvider({
     children,
+    createOnLogin = undefined,
```
jmderby, line 112: createOnLogin is TBD as we're still specifying the design for this. So this is subject to change.

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -126,20 +153,25 @@ export function CrossmintWalletProvider({
...
+    createOnLogin?: CreateOnLogin;
     showPasskeyHelpers?: boolean;
     appearance?: UIConfig;
     walletState: ValidWalletState;
     setWalletState: Dispatch<SetStateAction<ValidWalletState>>;
     experimental_enableRecoveryKeys?: boolean;
 }) {
     const { crossmint } = useCrossmint("CrossmintWalletProvider must be used within CrossmintProvider");
+    const crossmintAuthContext = useContext(AuthContext);
+    const isUsingCrossmintAuthProvider = crossmintAuthContext?.isWeb3Enabled ?? false;
```
jmderby, line 173: tldr here; we check if the auth provider is being used and if so we can still support our EOA wallets for signing in and creating on login!

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -257,12 +289,119 @@ function WalletProvider({
...
     );
 
+    // skip wallet creation if using CrossmintAuthProvider
+    const canAutomaticallyGetOrCreateWallet =
+        createOnLogin?.walletType != null &&
+        walletState.status === "not-loaded" &&
+        crossmint.jwt != null &&
+        !isUsingCrossmintAuthProvider;
+
+    const handleAutomaticWalletCreation = useCallback(async () => {
```
jmderby, line 299: adds support for allowing createOnLogin for our wallet provider. This is a new feature!

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -257,12 +289,119 @@ function WalletProvider({
...
+    }, [walletState.status, crossmint.jwt, clearWallet]);
+
+    useEffect(() => {
+        handleAutomaticWalletCreation();
+        handleWalletCleanup();
+    }, [handleAutomaticWalletCreation, handleWalletCleanup]);
+
     return (
         <WalletContext.Provider value={contextValue}>
+            {isUsingCrossmintAuthProvider && createOnLogin != null ? (
```
jmderby, line 325: and of course we conditionally render the provider for obvious reasons

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -76,6 +93,13 @@ type WalletContext =
     | (LoadedWalletState<"evm-smart-wallet"> & WalletContextFunctions)
     | (LoadedWalletState<"solana-smart-wallet"> & WalletContextFunctions);
 
+type CreateOnLogin = {
```
jmderby, line 96: my thoughts around createOnLogin is instead of it being a string like "all-users" or "off" we can just pass an object in with the necessary createOnLogin options. again tho as i mentioned this is TBD.

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -257,12 +289,119 @@ function WalletProvider({
         [walletState, experimental_getOrCreateWalletWithRecoveryKey]
     );
 
+    // skip wallet creation if using CrossmintAuthProvider
```
AlbertoElias, line 292: what do you mean? dont we also want to auto create wallet when using crossmint auth?

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.test.tsx
jmderby, line 1: I've moved the tests to their own __tests__ folder to cleanup!

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
jmderby: still tbd on this as were finalizing the sdk tomorrow!

packages/client/ui/react-ui/src/types/wallet.ts
```diff
@@ -1,4 +1,12 @@
...
 
 export type { GetOrCreateWalletProps, CrossmintAuthEmbeddedWallets } from "@crossmint/client-sdk-react-base";
 export type GetOrCreateWalletAdminSigner = EVMSignerInput | SolanaSignerInput | undefined;
+
+export type CreateOnLogin = {
+    walletType: CrossmintAuthEmbeddedWallets["type"];
+    chain?: string;
+    signer?: GetOrCreateWalletAdminSigner;
+    owner?: unknown;
+};
```
jmderby, line 12: still tbd on this type as mentioned above we'll be finalizing the sdk this week

packages/client/ui/react-ui/package.json
```diff
@@ -15,8 +15,8 @@
     "types": "./dist/index.d.ts",
     "files": ["dist", "LICENSE"],
     "scripts": {
-        "build": "cross-env NODE_OPTIONS='--max-old-space-size=8192' tsup",
-        "dev": "cross-env NODE_OPTIONS='--max-old-space-size=8192' tsup --watch",
+        "build": "cross-env NODE_OPTIONS='--max-old-space-size=16384' tsup",
```
jmderby, line 18: found a thread online talking about this. Seems to be an issue with tsup when bundling so many packages in parallel. Increasing the memory seems to be the best "fix" for now 🤷‍♂️ 

_link to thread: https://github.com/egoist/tsup/issues/920_

.changeset/slow-bats-perform.md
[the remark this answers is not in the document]
AlbertoElias: are we expecting any other breaking changes like the names of `createOnLogin`? If so, we should wait to merge this until we finalize that.

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -257,8 +277,53 @@ function WalletProvider({
...
+     * This is only used for Bring Your Own Auth (BYOA).
+     * Defer to <CrossmintAuthWalletManager/> for automatic wallet creation if using CrossmintAuthProvider.
+     */
+    const canAutomaticallyGetOrCreateWallet =
+        createOnLogin?.walletType != null &&
+        walletState.status === "not-loaded" &&
+        crossmint.jwt != null &&
+        !isUsingCrossmintAuthProvider;
+
+    console.log({ canAutomaticallyGetOrCreateWallet });
```
AlbertoElias, line 290: leftover?

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -257,8 +277,53 @@ function WalletProvider({
         [walletState, experimental_getOrCreateWalletWithRecoveryKey]
     );
 
+    /**
+     * This is only used for Bring Your Own Auth (BYOA).
+     * Defer to <CrossmintAuthWalletManager/> for automatic wallet creation if using CrossmintAuthProvider.
+     */
+    const canAutomaticallyGetOrCreateWallet =
```
AlbertoElias, line 284: cant we do it so that `CrossmintAuthWalletManager` sets `createOnLogin?.signer` and  `createOnLogin?.owner` instead of calling `getOrCreateWallet`, and we only call `getOrCreateWallet` here in the wallet provider?

packages/client/ui/react-ui/src/providers/__tests__/CrossmintAuthProvider.test.tsx
```diff
@@ -7,12 +7,14 @@ import { beforeEach, describe, expect, vi, it, type MockInstance } from "vitest"
 import { fireEvent, render, waitFor } from "@testing-library/react";
 import { mock } from "vitest-mock-extended";
 
-import { useAuth, useWallet } from "../hooks";
-import { CrossmintProvider, useCrossmint } from "../hooks/useCrossmint";
-import { MOCK_API_KEY } from "../testUtils";
-import { CrossmintAuthProvider } from "./CrossmintAuthProvider";
-import type { CrossmintAuthProviderEmbeddedWallets, LoginMethod } from "@/types/auth";
+import { useAuth, useWallet } from "@/hooks";
```
AlbertoElias, line 10: i think the trend is to move away from `@` as it causes some issues like loading references in the IDE

AlbertoElias (commented):
Left some comments, we should def wait until we reach confirmation on the new API to merge this if it's going to be a major release. I'd also move the tests out of the `__tests__` folder

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
jmderby: Yes, this change will require some refactoring. I'm wondering what the best move is. I'd like to remove the need for CrossmintAuthWalletManager and just set a web3 EOA admin signer state somewhere and use that. I think this change would be big though looking at it locally on my end...

packages/client/ui/react-ui/src/providers/__tests__/CrossmintAuthProvider.test.tsx
jmderby: you think so? tbh I thought it was better to use the @ alias. We have it configured that way in our `tsconfig.json` in `compilerOptions.paths`

AlbertoElias: yeah, but we started moving away from it in our other repo. i prefer the `@` as well, so much cleaner, but it also adds extra overhead for the ts server

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
AlbertoElias: if what i suggested is less hard as we would keep `CrossmintAuthWalletManager`, I would stick to that. I do think we should aim for a solution where we only call getOrCreate once. Ideally it's what you're suggesting though

jmderby: The problem is i don't know how we can only make a single call to `getOrCreateWallet` while also handling the eoa signer scenario

jmderby:
Closing. Going with a different approach.

PR #1102, opened 2025-05-05
hosted: add appearance rules

pgarciaegido:
## Description

Adding rules to Hosted v3 types. After adding this, the rules were being passed to the crossmint page successfully

## Test plan

Tested locally, combined with https://github.com/Paella-Labs/crossbit-main/pull/18631
Screenshot 2025-05-05 at 16 30 02

### DIFF
--- a/packages/client/base/src/types/hosted/v3/CrossmintHostedCheckoutV3Props.ts
+++ b/packages/client/base/src/types/hosted/v3/CrossmintHostedCheckoutV3Props.ts
@@ -39,7 +39,8 @@ export interface CrossmintHostedCheckoutV3Appearance {
     rules?: HostedCheckoutV3AppearanceRules;
 }
 
-// TODO: Should we just allow all the other embedded rules?
+// This type is different that the one in Embedded.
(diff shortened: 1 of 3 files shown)


packages/client/base/src/types/hosted/v3/CrossmintHostedCheckoutV3Props.ts
```diff
@@ -36,8 +36,19 @@ export interface CrossmintHostedCheckoutV3Appearance {
     variables?: CrossmintHostedCheckoutV3AppearanceVariables;
     overlay?: CrossmintHostedCheckoutV3OverlayOptions;
     display?: "popup" | "new-tab" | "same-tab";
+    rules?: HostedCheckoutV3AppearanceRules;
 }
 
+// TODO: Should we just allow all the other embedded rules?
```
pgarciaegido, line 42: I wonder if we should just reuse the existing interface from embedded. Any reason not to do it?

mPaella (approved):

PR #1108, opened 2025-05-08
React Native: Place ncs functions under onAuthRequired callback

AlbertoElias:
## Description

Eventually, the wallets sdk will expose an onAuthRequired callback with access to the custom ncs functions to set up the signer. This refactor imitates that flow.

## Test plan

Tested with expo app and will implement in quickstart

## Package updates

react-base: patch
react-native: patch


packages/client/ui/react-native/src/providers/CrossmintRecoveryKeyProvider.tsx
```diff
@@ -294,12 +432,14 @@ export function CrossmintRecoveryKeyProvider({
 
             try {
                 const baseUrl = validateApiKeyAndGetCrossmintBaseUrl(apiKey);
-                const response = await fetch(`${baseUrl}/api/unstable/wallets/ncs/irrelevant/public-key`, {
+
+                const response = await fetch(`${baseUrl}api/unstable/wallets/ncs/irrelevant/public-key`, {
```
jmderby, line 436: im assuming baseUrl has a tailing slash

jmderby (approved):
Nice!

PR #1118, opened 2025-05-16
NCS: Get Status Event

devlyn37:
## Description

This PR introduces a new event as part of ongoing work to harden the `signers.crossmint.com` iframe.

## Test plan

n/a just adding some types and schemas to use in the iframe code base.

## Package updates

PATCH @crossmint/client-signers


alberto-crossmint:
lgtm!

PR #1121, opened 2025-05-17
NCS: Handle Device Key Share Mismatch (WAL-4558)

devlyn37:
## Description

If there's a device keyshare mismatch, route the user through the OTP flow again.

Master PR with full context [here](https://github.com/Crossmint/signer-frames/pull/20)

## Test plan

Refer to testing section in master PR listed above :)

## Package updates

"@crossmint/client-sdk-react-ui": patch

### DIFF
--- a/packages/client/ui/react-ui/src/providers/signers/CrossmintSignerProvider.tsx
+++ b/packages/client/ui/react-ui/src/providers/signers/CrossmintSignerProvider.tsx
@@ -340,10 +340,6 @@ export function CrossmintSignerProvider({
         try {
             return await signFn();
         } catch (error: any) {
-            console.log("Here's the error");
-            console.log(error);
-            console.log(JSON.stringify(error));
-            console.log(error.code);
             if (error.code === "invalid-device-share") {
                 await handleGetRecoverySigner();
                 return await signFn();
--- a/packages/client/ui/react-ui/src/providers/signers/CrossmintSignerProvider.tsx
+++ b/packages/client/ui/react-ui/src/providers/signers/CrossmintSignerProvider.tsx
@@ -135,25 +135,29 @@ export function CrossmintSignerProvider({
         }
         return new Promise<void>(async (resolve, reject) => {
             try {
-                // Attempt to get the signer's public key
                 const signerResponse = await iframeWindow.current?.sendAction({
-                    event: "request:get-public-key",
-                    responseEvent: "response:get-public-key",
+                    event: "request:get-status",
+                    responseEvent: "response:get-status",
                     data: {
                         authData: {
                             jwt,
                             apiKey,
                         },
-                        data: {
-                            chainLayer: "solana",
-                        },
                     },
                 });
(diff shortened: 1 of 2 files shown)


packages/client/ui/react-ui/src/providers/signers/CrossmintSignerProvider.tsx
```diff
@@ -258,63 +259,71 @@ export function CrossmintSignerProvider({
...
+                                }
+                                if (res?.signature == null) {
+                                    throw new Error("Failed to sign transaction");
+                                }
+                                transaction.addSignature(
+                                    new PublicKey(publicSignerAddress),
+                                    base58.decode(res.signature)
+                                );
+                                return transaction;
                             });
```
devlyn37, line 326: Only changes here are:

wrapping each method with `signWithRecovery`, and adding the following error handling for signWithRecovery to catch and handle.
```TS
if (res?.status === "error") {
    const err = new Error(res.error);
    (err as any).code = res.code;
    throw err;
}
```
I consider this approach to be gross, but didn't want to do any big refactors with the context that this section will need to be re-written soon to be unified with the React native SDK.

packages/client/ui/react-ui/src/providers/signers/CrossmintSignerProvider.tsx
```diff
@@ -326,6 +335,19 @@ export function CrossmintSignerProvider({
...
+        try {
+            return await signFn();
+        } catch (error: any) {
+            if (error.code === "invalid-device-share") {
+                await handleGetRecoverySigner();
+                return await signFn();
+            }
+            throw error;
+        }
+    }
```
devlyn37, line 349: see https://github.com/Crossmint/crossmint-sdk/pull/1121/files#r2094588331

packages/client/ui/react-ui/src/providers/signers/CrossmintSignerProvider.tsx
```diff
@@ -258,63 +259,71 @@ export function CrossmintSignerProvider({
...
-                                responseEvent: "response:sign",
-                                data: {
-                                    authData: {
-                                        jwt,
-                                        apiKey,
-                                    },
+                        signMessage: (message: Uint8Array) => {
+                            return signWithRecovery(async () => {
+                                if (iframeWindow.current == null) {
+                                    throw new Error("IFrame window not initialized");
```
alberto-crossmint, line 265: If this is a customer exposed error we maybe want to make it less obscure. Maybe "signer is not initialized" or something similar?

alberto-crossmint (approved):

packages/client/ui/react-ui/src/providers/signers/CrossmintSignerProvider.tsx
devlyn37: Agree, but didn't change this code (just looks like it in the diff since indentation changed), so out of scope for this PR!

PR #1123, opened 2025-05-20
NCS: Event adaptation

alberto-crossmint:
## Description

SDK part of the modification we agreed on for managing raw public keys instead of addresses
The change is basically
- `publicKey` type changes from string to `{bytes: string, encoding: Encoding, keyType: KeyType}` (consistent also with signatures)
- `signature` type changes from string to `{bytes: string, encoding: Encoding }` (maybe we should add here keyType too?)
- `address` changes to `publicKey` everywhere

## Test plan

When this integrates into the iframe, an end to end test will be performed

### DIFF
--- a/packages/client/signers/src/communications/schemas.ts
+++ b/packages/client/signers/src/communications/schemas.ts
@@ -1,19 +1,43 @@
 import { z } from "zod";
 
-const KeyType = z.enum(["secp256k1", "ed25519"]);
+const KeyTypeSchema = z
+    .enum(["secp256k1", "ed25519"])
+    .describe("Type of cryptographic key");
+const EncodingSchema = z
+    .enum(["base58", "base64"])
+    .describe("Encoding format for the key or data");
+
+const PublicKeySchema = z.object({
+    bytes: z.string().describe("The encoded public key value"),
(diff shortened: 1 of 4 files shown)


packages/client/signers/src/communications/schemas.ts
```diff
@@ -1,20 +1,33 @@
...
 
-const SupportedChainLayer = z.enum(["solana", "evm"]);
-const KeyType = z.enum(["secp256k1", "ed25519"]);
+const KeyTypeSchema = z.enum(["secp256k1", "ed25519"]).describe("Type of cryptographic key");
+const EncodingSchema = z.enum(["base58", "base64"]).describe("Encoding format for the key or data");
+
+const UserPublicKeySchema = z.object({
+    bytes: z.string().describe("The encoded public key value"),
+    encoding: EncodingSchema.describe("The encoding format of the public key payload"),
+    keyType: KeyTypeSchema.describe("The cryptographic algorithm of the public key"),
```
devlyn37, line 9: Whatever we name this, we should be consistent internally (TEE, Crossmint Proxy) we rename the different pieces

devlyn37 (approved):
Nice

PR #1132, opened 2025-05-21
React Native: Fixes callback dependencies so they dont update constantly

AlbertoElias:
## Description

There were some messy dependencies triggering constant refreshes 

## Test plan

Tested on example expo app, but will need to test on quickstart

## Package updates

react-base: patch
react-native: patch

### DIFF
--- a/packages/client/ui/react-native/src/providers/CrossmintRecoveryKeyProvider.tsx
+++ b/packages/client/ui/react-native/src/providers/CrossmintRecoveryKeyProvider.tsx
@@ -419,9 +419,9 @@ export function CrossmintRecoveryKeyProvider({
 
     const experimental_createRecoveryKeySigner = useCallback(
         async (emailInput: string): Promise<RecoverySigner | null> => {
-            if (jwt == null || apiKey == null) {
+            if (jwt == null || apiKey == null || appId == null) {
                 console.warn(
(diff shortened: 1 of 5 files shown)


packages/client/ui/react-native/src/providers/CrossmintRecoveryKeyProvider.tsx
```diff
@@ -503,7 +502,7 @@ export function CrossmintRecoveryKeyProvider({
                 return null;
             }
         },
-        [jwt, apiKey, buildRecoverySigner, getOrCreateWallet]
+        [jwt, apiKey, appId, buildRecoverySigner, getOrCreateWallet]
```
jmderby, line 505: once small comment - i know sometimes adding `getOrCreateWallet` in the dependency array can create infin re-renders just something to consider here

jmderby (approved):
lgtm nice

jmderby:
ci is failing due to memory. we should make it 16 and land that asap

packages/client/ui/react-native/src/providers/CrossmintRecoveryKeyProvider.tsx
AlbertoElias: the point of this pr is to avoid that with by wrapping `getOrCreateWallet` in a useCallback and with the changes I did of moving the wallet status to a ref

PR #1141, opened 2025-05-26
NCS: Add public key mapping to event response [WAL-4590]

alberto-crossmint:
## Description

Adapting the interface of some iframe evenets

## Test plan

Will be tested end to end on the iframe update

### DIFF
--- a/packages/client/signers/src/communications/schemas.ts
+++ b/packages/client/signers/src/communications/schemas.ts
@@ -16,6 +16,8 @@ const SignatureSchema = z.object({
     encoding: EncodingSchema.describe("The encoding format of the signature payload"),
 });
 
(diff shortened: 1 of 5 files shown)


packages/client/signers/src/communications/schemas.ts
```diff
@@ -107,9 +110,15 @@ export const GetPublicKeyPayloadSchema = {
...
-        })
+        z.union([
+            z.object({
+                signerStatus: z.enum(["ready"]).describe("Current status of the signer"),
+                publicKeys: PublicKeyMappingSchema.describe("The public keys of the created signer"),
+            }),
+            z.object({
+                signerStatus: z.enum(["new-device"]).describe("Current status of the signer"),
+            }),
+        ])
```
devlyn37, line 121: Should we have a single source of truth for these schema since they're shared across multiple responses?

devlyn37 (approved):
We should also remove `get-public-key` while you're here!

PR #1152, opened 2025-05-29
SSW: tweak poll-for-confirmation retry time

alberto-crossmint:
## Description

Adding backoff and reducing initial retry time


packages/wallets/src/solana/services/transactions-service.ts
```diff
@@ -37,6 +37,7 @@ export class SolanaTransactionsService {
             ...(params.signer != null ? [params.signer] : []),
             ...(params.additionalSigners || []),
         ]);
+        await sleep(1_000); // Rule of thumb: tx won't be confirmed in less than 1 second
```
AlbertoElias, line 40: so this is a way to give it a chance to be ready from the get go after this 1s?

PR #1153, opened 2025-05-30
React UI: Refactors to separate Auth and Wallet providers

AlbertoElias:
## Description

We wanted to separate concerns and have the wallet provider manage all the logic for wallet stuff and the auth provider only focus on Crossmint Auth. 

## Test plan

Tested all flows with the quickstart app

## Package updates

NA


jmderby (approved):
Amazing refactor!

PR #1154, opened 2025-05-30
React Native: Uses email signer and new wallets sdk in react native

AlbertoElias:
## Description

Refactors react native to use new wallets sdk and email signer

## Test plan

Tested with example expo app

## Package updates

NA


jmderby (approved):
nice job! 
1/3 of the original size too!

PR #1161, opened 2025-06-02
Refactor: Make user object available in crossmint instance

AlbertoElias:
## Description

With our new email signer, we want to make it very easy for the wallet and react sdk to be able to create the email signer. But the email is store in the auth provider (either Crossmint Auth or Bring Your Own Auth). With this change, auth providers can call `setUser` or `crossmint.user` and store the email in the crossmint instance which is shared by all sdks. And then the wallet sdk can fetch the email from there and create a wallet automatically.

## Test plan

Tested with example apps


.changeset/itchy-pigs-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@crossmint/client-sdk-smart-wallet": patch
```
jmderby, line 2: this package is no longer maintained and basically deprecated. do we need to bother with updated it?

apps/wallets/quickstart-devkit/app/home.tsx
```diff
@@ -54,26 +54,6 @@ export function HomeContent() {
                     <CrossmintAuthLoginButton />
                     {/* <PrivyLoginButton /> */}
                     {/* <DynamicLabsLoginButton /> */}
-
```
jmderby, line 57: nice!

packages/client/react-base/src/hooks/useCrossmint.tsx
```diff
@@ -1,9 +1,11 @@
 import { type ReactNode, createContext, useCallback, useContext, useMemo, useRef, useState } from "react";
-import { type Crossmint, createCrossmint } from "@crossmint/common-sdk-base";
+import { type Crossmint, type CrossmintConfig, type User, createCrossmint } from "@crossmint/common-sdk-base";
+import isEqual from "lodash.isequal";
```
jmderby, line 3: nit this pkg is deprecated but i don't know what else to use instead

packages/client/wallets/smart-wallet/src/sdk.ts
```diff
@@ -1,4 +1,4 @@
-import { type APIKeyEnvironmentPrefix, validateAPIKey } from "@crossmint/common-sdk-base";
+import { type APIKeyEnvironmentPrefix, createCrossmint, validateAPIKey } from "@crossmint/common-sdk-base";
```
jmderby, line 1: nit package is no longer maintained

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -290,7 +288,7 @@ export function CrossmintWalletProvider({
                     ? createPortal(
                           <EmailSignersDialog
                               rejectRef={rejectRef}
-                              email={email}
+                              email={email ?? ""}
```
jmderby, line 291: should do `email={crossmint?.user?.email}` then delete the email var above. Or we should only render this comp if email exists.

packages/client/ui/react-native/src/providers/CrossmintAuthProvider.tsx
```diff
@@ -85,12 +90,15 @@ export function CrossmintAuthProvider({
...
                     onLogout: () => {
-                        setJwt(undefined);
                         setUser(undefined);
+                        setCrossmintUser(undefined);
                     },
                     onTokenRefresh: (authMaterial: AuthMaterialWithUser) => {
-                        setJwt(authMaterial.jwt);
                         setUser(authMaterial.user);
+                        setCrossmintUser({
+                            email: authMaterial.user.email,
```
jmderby, line 99: email might not exist, we should optional chain this `authMaterial.user?.email`

jmderby:
One other consideration is, will it be super obvious when to call `setUser` vs `setJwt`? I wonder if we should just expose one ever for BYOA as Crossmint auth wont require manually calling. 

`setAuth({ jwt: "abc..."}, { user: {email: "test@gmail.com" }}); // jwt or email `

What're your thoughts?

.changeset/itchy-pigs-behave.md
jmderby: i guess separate pr but maybe we should just delete it or remove it from the build script 😅

PR #1171, opened 2025-06-04
Wallets sdk: Fix passkey signer validation

jmderby:
## Description

Fix issue with `validateWalletConfig` not working for passkey signer types

## Test plan

passkey signer validation works.

## Package updates

@crossmint/wallets-sdk

### DIFF
--- a/packages/wallets/src/utils/signer-validation.ts
+++ b/packages/wallets/src/utils/signer-validation.ts
@@ -5,13 +5,13 @@ export function deepCompare(obj1: Record<string, unknown>, obj2: Record<string, 
         return;
     }
     if (obj1 == null || obj2 == null) {
-        throw new WalletCreationError(`Wallet signer ${path} does not match existing wallet's signer ${path}`);
+        throw new Error(`Cannot compare null or undefined objects at path: ${path}`);
     }
     for (const key of Object.keys(obj1)) {
         const newPath = path ? `${path}.${key}` : key;
         if (!(key in obj2)) {
             throw new WalletCreationError(
-                `Wallet signer ${newPath} does not match existing wallet's signer ${newPath}`
+                `Wallet signer configuration mismatch at "${newPath}" - expected "${obj2[key]}" from existing wallet but found "${obj1[key]}"`
             );
         }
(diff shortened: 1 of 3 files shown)


packages/wallets/src/wallets/wallet-factory.ts
jmderby, line 1: Primary issue was revolved around calling `this.configureSigner` for when the signer type is passkeys. That method was then trigger the passkey prompts due to trying to create a new credential.

We also need to do deep recursive checks now that passkey object will be checked considering there's nested props.

packages/wallets/src/wallets/wallet-factory.ts
```diff
@@ -239,14 +250,26 @@ export class WalletFactory {
                 );
             }
 
-            const signerProps = Object.keys(configuredArgsSigner) as Array<keyof typeof configuredArgsSigner>;
-            for (const prop of signerProps) {
-                if (prop in existingWalletSigner && configuredArgsSigner[prop] !== existingWalletSigner[prop]) {
+            const deepCompare = (obj1: any, obj2: any, path = ""): void => {
```
AlbertoElias, line 253: we should move this to a util file

packages/wallets/src/wallets/wallet-factory.ts
```diff
@@ -239,14 +250,26 @@ export class WalletFactory {
...
+                if (obj1 === obj2) {
+                    return;
+                }
+                if (typeof obj1 !== "object" || obj1 === null || obj2 === null) {
```
AlbertoElias, line 257: probs want to do == null

packages/wallets/src/utils/signer-validation.ts
```diff
@@ -0,0 +1,19 @@
+import { WalletCreationError } from "./errors";
+
+export function deepCompare(obj1: Record<string, unknown>, obj2: Record<string, unknown>, path = "") {
+    if (obj1 === obj2) {
+        return;
+    }
+    if (obj1 == null || obj2 == null) {
+        throw new WalletCreationError(`Wallet signer ${path} does not match existing wallet's signer ${path}`);
```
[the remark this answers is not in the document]
jmderby: how about something like this? 

> throw new Error(`Cannot compare null or undefined objects at path: ${path}`);

PR #1180, opened 2025-06-09
React: Makes getOrCreateWallet response from react include wallet instance

AlbertoElias:
## Description

Clients prefer this and it feels more natural than depending exclusively on hooks

## Test plan

All tests pass

## Package updates

react-base: patch
react-native: patch
react-ui: patch

### DIFF
--- a/packages/client/react-base/src/hooks/useWalletState.ts
+++ b/packages/client/react-base/src/hooks/useWalletState.ts
@@ -78,13 +78,14 @@ export function useWalletState({
 
                 const wallet = await crossmintWallets.getOrCreateWallet(props);
                 setState({ status: "loaded", wallet });
+                return { startedCreation: true, wallet };
             } catch (error: unknown) {
                 console.error("There was an error creating a wallet ", error);
                 setState(deriveErrorState(error));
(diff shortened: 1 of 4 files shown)


packages/client/react-base/src/hooks/useWalletState.ts
```diff
@@ -78,13 +78,14 @@ export function useWalletState({
 
                 const wallet = await crossmintWallets.getOrCreateWallet(props);
                 setState({ status: "loaded", wallet });
+                return { startedCreation: true, wallet };
             } catch (error: unknown) {
                 console.error("There was an error creating a wallet ", error);
                 setState(deriveErrorState(error));
+                return { startedCreation: true, reason: `There was an error creating a wallet ${error}` };
```
jmderby, line 85: more of a nit: wonder if instead of "creating" we use "fetching" to encapsulate both creating and/or getting language

jmderby (approved):
lgtm

packages/client/react-base/src/hooks/useWalletState.ts
AlbertoElias: not sure fetching is great for that. We need a word in english for `getOrCreate` hahahaha

PR #1207, opened 2025-06-18
Wallets sdk: Update balance api

jmderby:
## Description

Update the balance api to return a better format.

closes #[WAL-4879](https://linear.app/crossmint/issue/WAL-4879/update-balance-method)

## Test plan

Tested in quickstart sample

## Package updates

@crossmint/wallets-sdk

### DIFF
--- a/packages/wallets/src/wallets/wallet.ts
+++ b/packages/wallets/src/wallets/wallet.ts
@@ -1,6 +1,6 @@
 import { isValidAddress } from "@crossmint/common-sdk-base";
-import type { Activity, ApiClient, Balances, GetSignatureResponse } from "../api";
-import type { PendingApproval, DelegatedSigner, WalletOptions } from "./types";
+import type { Activity, ApiClient, GetBalanceSuccessResponse, GetSignatureResponse } from "../api";
+import type { PendingApproval, DelegatedSigner, WalletOptions, Balances, TokenBalance } from "./types";
 import {
     InvalidSignerError,
     SignatureNotAvailableError,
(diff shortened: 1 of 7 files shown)


packages/wallets/src/wallets/wallet.ts
```diff
@@ -61,21 +75,85 @@ export class Wallet<C extends Chain> {
...
+        const transformTokenBalance = (tokenData: GetBalanceSuccessResponse[number]): TokenBalance => {
+            let contractAddress: string | undefined;
+            const chainData = tokenData.forChain?.[this.chain];
+            if (chainData && "contractAddress" in chainData) {
+                contractAddress = chainData.contractAddress;
+            }
+
+            return {
+                symbol: tokenData.token,
+                name: tokenData.token, // API doesn't provide name, using symbol as fallback
```
jmderby, line 118: iirc, we'll update the api to include the actual name.

packages/wallets/src/wallets/wallet.ts
```diff
@@ -61,21 +75,85 @@ export class Wallet<C extends Chain> {
...
      */
-    public async balances(tokens: string[], chains?: Chain[]): Promise<Balances> {
+    public async balances(tokens?: string[], chains?: Chain[]): Promise<Balances> {
+        const nativeToken = this.chain === "solana" ? "SOL" : "ETH";
+        const allTokens = [nativeToken, "USDC", ...(tokens ?? [])];
+
         const response = await this.#apiClient.getBalance(this.address, {
             chains: chains ?? [this.chain],
-            tokens,
+            tokens: allTokens.map((token) => token.toLowerCase()),
```
jmderby, line 90: api should also be able to lower case input regardless of uppercase and not throw a 500 if passed

PR #1228, opened 2025-06-24
⚠️ Fix issue with email signers and jwt not being set

jmderby:
## Description

A rather critical bug has been found and fixed. 

**issue 1:** 
jwt value wasn't using a single source of truth and incorrect value was being used even though jwt was being set on crossmint instance.
**issue 2:** 
email signer error being thrown about email not being set even though it was being set in wallet instance. 

## Test plan

Tested setting jwt in `createCrossmint` instance and setting email signer in `getOrCreateWallet` on wallet instance.

## Package updates

@crossmint/common-sdk-base
@crossmint/wallets-sdk

### DIFF
--- a/packages/common/base/src/apiClient/CrossmintApiClient.ts
+++ b/packages/common/base/src/apiClient/CrossmintApiClient.ts
@@ -30,7 +30,6 @@ export class CrossmintApiClient extends ApiClient {
             throw new Error(apiKeyValidationResult.message);
         }
         super();
-        this.crossmint.jwt = this.crossmint.jwt || this.crossmint.experimental_customAuth?.jwt;
         this.parsedAPIKey = apiKeyValidationResult;
         this.internalConfig = internalConfig;
     }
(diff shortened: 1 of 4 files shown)


packages/common/base/src/apiClient/CrossmintApiClient.ts
```diff
@@ -30,6 +30,7 @@ export class CrossmintApiClient extends ApiClient {
             throw new Error(apiKeyValidationResult.message);
         }
         super();
+        this.crossmint.jwt = this.crossmint.jwt || this.crossmint.experimental_customAuth?.jwt;
```
jmderby, line 33: my reasoning around global mutation is we should create single source of truth for our jwt from wherever the crossmint.jwt value is retrieved from so this doesn't happen in the future again.

AlbertoElias: i wouldnt force set to `this.crossmint.jwt`. maybe set to a property that's only for CrossmintApiClient. And also consider the case where this.crossmint is mutated via `setJwt` or `setCustomAuth` and an api request is made. at that point we'll fetch the headers from this.crossmint.jwt and if it's in customAuth, it won't be picked up.

So... we need to make sure crossmint.jwt is synced with customAuth.jwt and remember that we are deprecating crossmint.jwt. and getheaders should still user customAuth.jwt

packages/wallets/src/signers/email/email.ts
```diff
@@ -208,14 +208,13 @@ export class EmailSigner implements Signer {
     }
 
     static async pregenerateSigner(email: string, crossmint: Crossmint): Promise<string> {
-        if (email == null || crossmint.experimental_customAuth?.email == null) {
+        const emailToUse = email ?? crossmint.experimental_customAuth?.email;
```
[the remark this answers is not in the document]
jmderby: before we were doing 
`if(email == null || crossmint.experimental_customAuth?.email == null)`

if one or the other was null it would throw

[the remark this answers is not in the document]
AlbertoElias: ah, of course, it should've been &&

packages/common/base/src/apiClient/CrossmintApiClient.ts
jmderby: Right, we should soon remove the `experimental_` flag from the var 😅 

I've added a private method for getting the jwt and removed the mutation 👍

AlbertoElias (approved):

PR #1231, opened 2025-06-25
WAL-4805: Add EVM email signers

maxisch:
## Description

Added EVM email signers. Created an abstract class for shared functionality with the Solana email signer.

## Test plan

Created a wallet and sent some usdc: https://scope.sh/11155111/op/0x756f435f8d3214d5cad10d4abe51632891e6952f8e3b844582f4c7fc91e63bb5

## Package updates
@crossmint/wallets-sdk

### DIFF
--- a/packages/wallets/src/signers/email/email-signer-api-client.ts
+++ b/packages/wallets/src/signers/email/email-signer-api-client.ts
@@ -13,9 +13,9 @@ export class EmailSignerApiClient extends CrossmintApiClient {
         });
     }
 
-    async pregenerateSigner(email: string) {
+    async pregenerateSigner(email: string, keyType: "ed25519" | "secp256k1") {
         const response = await this.post(`${this.apiPrefix}/derive-public-key`, {
-            body: JSON.stringify({ authId: `email:${email}`, keyType: "ed25519" }),
+            body: JSON.stringify({ authId: `email:${email}`, keyType }),
             headers: this.headers,
         });
         if (!response.ok) {
--- a/packages/wallets/src/signers/email/evm-email-signer.ts
+++ b/packages/wallets/src/signers/email/evm-email-signer.ts
@@ -1,0 +1,87 @@
+import type { EmailInternalSignerConfig } from "../types";
+import { EmailSignerApiClient } from "./email-signer-api-client";
+import { EmailSigner, DEFAULT_EVENT_OPTIONS } from "./email-signer";
+import type { Crossmint } from "@crossmint/common-sdk-base";
+import { Address, PublicKey } from "ox";
+
+export class EvmEmailSigner extends EmailSigner {
+    constructor(config: EmailInternalSignerConfig) {
+        super(config);
+    }
+
+    locator() {
+        return `evm-keypair:${this.config.signerAddress}`;
+    }
+
+    async signMessage() {
+        return await Promise.reject(new Error("signMessage method not implemented for email signer"));
+    }
+
+    async signTransaction(transaction: string): Promise<{ signature: string }> {
+        await this.handleAuthRequired();
+
+        const hexString = transaction.startsWith("0x") ? transaction.slice(2) : transaction;
+
+        const res = await this.config._handshakeParent?.sendAction({
+            event: "request:sign",
+            responseEvent: "response:sign",
+            data: {
+                authData: {
+                    jwt: this.config.crossmint.experimental_customAuth?.jwt ?? "",
+                    apiKey: this.config.crossmint.apiKey,
+                },
+                data: {
+                    keyType: "secp256k1",
+                    bytes: hexString,
+                    encoding: "hex",
+                },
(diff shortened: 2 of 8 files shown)


packages/wallets/src/signers/email/evm-email-signer.ts
```diff
@@ -0,0 +1,85 @@
...
+            return this.publicKeyToEvmAddress(publicKey.bytes);
+        } catch (error) {
+            console.error("[EvmEmailSigner] Failed to pregenerate signer:", error);
+            throw error;
+        }
+    }
+
+    private static publicKeyToEvmAddress(publicKeyHex: string): string {
+        const publicKey = PublicKey.from(`0x${publicKeyHex}`);
+        return Address.fromPublicKey(publicKey, { checksum: true });
```
maxisch, line 83: do we want to return the checksum address? I think we do, but double checking

packages/wallets/src/signers/email/evm-email-signer.ts
```diff
@@ -0,0 +1,85 @@
...
+        await this.handleAuthRequired();
+
+        const hexString = transaction.startsWith("0x") ? transaction.slice(2) : transaction;
+
+        const res = await this.config._handshakeParent?.sendAction({
+            event: "request:sign",
+            responseEvent: "response:sign",
+            data: {
+                authData: {
+                    jwt: this.config.crossmint.experimental_customAuth?.jwt ?? "",
```
alberto-crossmint, line 30: If the jwt field is not defined we should raise an error! Doing validation here will always be easier than propagating errors in the iframe handler

packages/wallets/src/signers/email/evm-email-signer.ts
```diff
@@ -0,0 +1,85 @@
...
+        if (res?.status === "error") {
+            throw new Error(res.error);
+        }
+
+        if (res?.signature == null) {
+            throw new Error("Failed to sign transaction");
+        }
+
+        return { signature: res.signature.bytes };
+    }
```
alberto-crossmint, line 51: Let's check that the encoding and the key type we get is correct! Just as we do with Solana email signer

packages/wallets/src/signers/email/email-signer-api-client.ts
```diff
@@ -13,9 +13,9 @@ export class EmailSignerApiClient extends CrossmintApiClient {
         });
     }
 
-    async pregenerateSigner(email: string) {
+    async pregenerateSigner(email: string, keyType: "ed25519" | "secp256k1") {
```
alberto-crossmint, line 16: We have the keyType type in the signers package!

packages/wallets/src/signers/email/evm-email-signer.ts
```diff
@@ -0,0 +1,85 @@
...
+        super(config);
+    }
+
+    locator() {
+        return `evm-keypair:${this.config.signerAddress}`;
+    }
+
+    async signMessage() {
+        return await Promise.reject(new Error("signMessage method not implemented for email signer"));
+    }
```
alberto-crossmint, line 18: Why is this not implemented? I think we know exactly how to implement :) and that may be required to some features as doing signature handling. Could we add it either in this PR or in a follow up?

packages/wallets/src/signers/email/evm-email-signer.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    async signTransaction(transaction: string): Promise<{ signature: string }> {
+        await this.handleAuthRequired();
+
+        const hexString = transaction.startsWith("0x") ? transaction.slice(2) : transaction;
```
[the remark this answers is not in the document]
alberto-crossmint: Our decoding logic in the iframe not allowing for 0x is a pita though, we should change that!

packages/wallets/src/signers/email/evm-email-signer.ts
alberto-crossmint, line 83: you gotta love ox 😍

alberto-crossmint: I can't think of a reason for not doing it!

packages/wallets/src/wallets/wallet-factory.ts
```diff
@@ -228,15 +233,20 @@ export class WalletFactory {
...
-                type: "solana-keypair",
-                address: emailSignerAddress,
-            };
+            if (chain === "solana") {
+                const emailSignerAddress = await SolanaEmailSigner.pregenerateSigner(email, this.apiClient.crossmint);
+                return {
+                    type: "solana-keypair",
+                    address: emailSignerAddress,
+                };
+            } else {
```
alberto-crossmint, line 243: else-if! Let's not have catch-all branches. I think we want to fail and throw an error if we reach this point with a wallet that is not solana or evm

PR #1233, opened 2025-06-25
NCS: extract some API communication schemas

alberto-crossmint:
## Description

Extracting communications schemas to this common library so we can handled them in a centralized way

## Test plan

It's just copied from the iframes repo, so nothing else. Will substitute this lib in both the iframe, the relayer and the tee and test e2e.


maxisch (approved):

devlyn37 (commented):
We may actually want to go in the other direction. Open Signer is meant to be independent from Crossmint, and shouldn't depend on Crossmint specific packages. I'd propose we instead begin moving any Open Signer related pieces we need centralized (API & Iframe Event Schema, etc) to the Open Signer repository.

Details need refinement, but was thinking the end state should look something like:
- We publish an `open-signer/schema` package from the Open Signer repository.
- Crossmint's SDKs use this package to interface with the iframe.
- Crossmint's relay uses this package and implements the relay endpoints according to the minimal schema defined in the package.

Not sure we'll prioritize this right away, but if we're aligned we should avoid moving in the opposite direction!

PR #1234, opened 2025-06-25
WAL-5151: Add EVM sign message

maxisch:
## Description

Add capability 

## Test plan

Tested with this react component: https://github.com/Crossmint/crossmint-sdk/pull/1235

### DIFF
--- a/packages/wallets/src/signers/email/evm-email-signer.ts
+++ b/packages/wallets/src/signers/email/evm-email-signer.ts
@@ -2,7 +2,7 @@ import type { EmailInternalSignerConfig } from "../types";
 import { EmailSignerApiClient } from "./email-signer-api-client";
 import { EmailSigner, DEFAULT_EVENT_OPTIONS } from "./email-signer";
 import type { Crossmint } from "@crossmint/common-sdk-base";
-import { Address, PublicKey, PersonalMessage, Hex } from "ox";
+import { Address, PublicKey, PersonalMessage } from "ox";
 
 export class EvmEmailSigner extends EmailSigner {
     constructor(config: EmailInternalSignerConfig) {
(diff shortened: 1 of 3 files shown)


packages/wallets/src/signers/email/evm-email-signer.ts
```diff
@@ -13,8 +13,9 @@ export class EvmEmailSigner extends EmailSigner {
         return `evm-keypair:${this.config.signerAddress}`;
     }
 
-    async signMessage() {
-        return await Promise.reject(new Error("signMessage method not implemented for email signer"));
+    async signMessage(message: string) {
```
alberto-crossmint, line 16: Is this the correct interface? [Ethers signMessage](https://github.com/ethers-io/ethers.js/blob/4eada383ab9833f9b4847ea9bdf39910c4eb508e/src.ts/wallet/base-wallet.ts#L100) receives a `string | Uint8Array` input. Assuming that the input it receives will always be a hex string diverges from the de facto standard.

I guess we want to leave it like this because of consistency with other signers in the SDK! But I'm bringing this up in case you have other opinion

packages/wallets/src/signers/email/evm-email-signer.ts
```diff
@@ -13,8 +13,9 @@ export class EvmEmailSigner extends EmailSigner {
...
+        const messageToSign = PersonalMessage.getSignPayload(message as `0x${string}`);
```
alberto-crossmint, line 17: Should we do 
`const messageRaw = isHex(message) ? message as Hex : new toHex(message)` 
and then 
`const messageToSign = PersonalMessage.getSignPayload(messageRaw)`? that way it would work for both hex raw payloads but also for signing utf8 messages

alberto-crossmint: That way we maintain compatibility with all other signers and also with ethers interface for utf msgs

alberto-crossmint (approved):
Approved!! Left a minor comment

PR #1238, opened 2025-06-26
Signer types vary based on chain (TypeScript)

jmderby:
## Description

Typescript types weren't working for when unsupported signer types were passed based on chain passed. For instance, passing `createOnLogin={{ chain: "solana", signer: { type: "passkey" } }}` wasn't throwing a ts error and as you probably know, we don't support passkeys on solana yet. 

closes https://linear.app/crossmint/issue/WAL-4999/signer-types-for-createonlogin-dont-fail-based-on-the-chain

## Test plan

tested that type error throws on react, react native, and just wallets sdk 

## Package updates

@crossmint/common-sdk-base
@crossmint/wallets-sdk

### DIFF
--- a/packages/wallets/src/signers/types.ts
+++ b/packages/wallets/src/signers/types.ts
@@ -93,7 +93,7 @@ export type PasskeySignResult = {
 
 export type SignerConfigForChain<C extends Chain> = C extends SolanaChain
     ? EmailSignerConfig | BaseSignerConfig<C>
-    : PasskeySignerConfig | BaseSignerConfig<C>;
+    : EmailSignerConfig | PasskeySignerConfig | BaseSignerConfig<C>;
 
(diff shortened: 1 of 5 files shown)


packages/wallets/src/signers/types.ts
```diff
@@ -93,7 +93,7 @@ export type PasskeySignResult = {
 
 export type SignerConfigForChain<C extends Chain> = C extends SolanaChain
     ? EmailSignerConfig | BaseSignerConfig<C>
-    : PasskeySignerConfig | BaseSignerConfig<C>;
+    : EmailSignerConfig | PasskeySignerConfig | BaseSignerConfig<C>;
```
jmderby, line 96: Now that evm supports email signers, I added the corresponding email type here!

PR #1240, opened 2025-06-26
Move _handshakeParent from signer to options

jmderby:
## Description

Move _handshakeParent property from the signer to the options config. Also in CrossmintWalletBaseProvider, rename `_getEmailSignerIframe` to `_getSignerConnection`.

closes https://linear.app/crossmint/issue/WAL-4794/handshakeparent-should-not-be-on-email-signer

## Test plan

ensured that the `_handshakeParent` is no longer on signer and instead in options.

## Package updates

@crossmint/wallets-sdk

### DIFF
--- a/packages/client/react-base/src/providers/CrossmintWalletBaseProvider.tsx
+++ b/packages/client/react-base/src/providers/CrossmintWalletBaseProvider.tsx
@@ -18,7 +18,7 @@ export type CrossmintWalletBaseContext = {
     status: "not-loaded" | "in-progress" | "loaded" | "error";
     getOrCreateWallet: <C extends Chain>(props: WalletArgsFor<C>) => Promise<Wallet<Chain> | undefined>;
     onAuthRequired?: EmailSignerConfig["onAuthRequired"];
-    _getEmailSignerIframe?: () => HandshakeParent<typeof signerOutboundEvents, typeof signerInboundEvents>;
+    _getSignerConnection?: () => HandshakeParent<typeof signerOutboundEvents, typeof signerInboundEvents>;
 };
 
 export const CrossmintWalletBaseContext = createContext<CrossmintWalletBaseContext>({
@@ -26,7 +26,7 @@ export const CrossmintWalletBaseContext = createContext<CrossmintWalletBaseConte
     status: "not-loaded",
     getOrCreateWallet: () => Promise.resolve(undefined),
     onAuthRequired: undefined,
-    _getEmailSignerIframe: undefined,
+    _getSignerConnection: undefined,
 });
 
 export interface CrossmintWalletBaseProviderProps {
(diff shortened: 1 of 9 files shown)


packages/client/react-base/src/providers/CrossmintWalletBaseProvider.tsx
```diff
@@ -18,15 +18,15 @@ export type CrossmintWalletBaseContext = {
     status: "not-loaded" | "in-progress" | "loaded" | "error";
     getOrCreateWallet: <C extends Chain>(props: WalletArgsFor<C>) => Promise<Wallet<Chain> | undefined>;
     onAuthRequired?: EmailSignerConfig["onAuthRequired"];
-    _getEmailSignerIframe?: () => HandshakeParent<typeof signerOutboundEvents, typeof signerInboundEvents>;
+    _getSignerConnection?: () => HandshakeParent<typeof signerOutboundEvents, typeof signerInboundEvents>;
```
aigustin, line 21: ok, two things: 
1. _getSignerConnection is too generic. I would call it something like: `getTEEConnector` which is what essentially this is doing. I'd remove the underscore too, its not a private var, anyone can override this, it's fine. 
2. I'd also rename the `HandshakeParent` to `TEEConnection`

packages/wallets/src/wallets/types.ts
```diff
@@ -32,6 +34,7 @@ export type Callbacks = {
 
 export type WalletOptions = {
     experimental_callbacks?: Callbacks;
+    _handshakeParent?: HandshakeParent<typeof signerOutboundEvents, typeof signerInboundEvents>;
```
aigustin, line 37: This name should also be consistent, we shouldn't use two different names

packages/client/react-base/src/providers/CrossmintWalletBaseProvider.tsx
AlbertoElias: How about clientTEEConnection? To separate the frame tee client (client) from the server one

packages/client/react-base/src/providers/CrossmintWalletBaseProvider.tsx
```diff
@@ -18,15 +18,15 @@ export type CrossmintWalletBaseContext = {
     status: "not-loaded" | "in-progress" | "loaded" | "error";
     getOrCreateWallet: <C extends Chain>(props: WalletArgsFor<C>) => Promise<Wallet<Chain> | undefined>;
     onAuthRequired?: EmailSignerConfig["onAuthRequired"];
-    _getEmailSignerIframe?: () => HandshakeParent<typeof signerOutboundEvents, typeof signerInboundEvents>;
+    clientTEEConnection?: () => HandshakeParent<typeof signerOutboundEvents, typeof signerInboundEvents>;
```
aigustin, line 21: HandshakeParent should also be renamed. Is it the function is getClientTEEConnection and the object is ClientTEEConnection?

aigustin (approved):

PR #1242, opened 2025-06-26
Housekeeping: cleanup peer dep warnings upon installation

jmderby:
## Description

When installing @crossmint/client-sdk-react-ui, there was an abundance of peer dep warnings on the screen. This PR attempts to remediate those number of warnings. I've also reached out to the dynamic team in requesting a fix from their side as well. convo here https://crossmint.slack.com/archives/C04TYMXHT7C/p1750961861054149

Our node versions also align with the main server. 

Screenshot 2025-07-28 at 6 57 59 PM

To improve even more we could address the react-native warning - just need to figure out whats wrong with that.

## Test plan

build passes and no major versions bumped

## Package updates

@crossmint/client-sdk-walletconnect
@crossmint/client-sdk-base
@crossmint/common-sdk-base
@crossmint/wallets-sdk


apps/payments/create-react-app/package.json
jmderby, line 1: This is a super old demo app, instead of deprecating it right now i just marked it to skip the build in CI. makes ci take longer too. 

Reason for doing so is because when ci is true, the warnings are treated as errors and there were a bunch of warnings for some strange reason.

PR #1246, opened 2025-06-28
Wallets sdk: Expose new param options.experimental_prepareOnly on EVM/SOL wallet

jmderby:
## Description

Add new param `options` with `experimental_prepareOnly` that will return an unsigned transaction id if used. 

added to wallet `send` method and evm and solana specific `sendTransaction` methods. 

Closes: https://linear.app/crossmint/issue/WAL-5150/add-preparetransaction-to-solanawallet-and-evmwallet-server-side
Closes: https://linear.app/crossmint/issue/WAL-5179/update-sendtransaction-method-on-evm-with-missing-inputs

## Test plan

new params works on both evm and sol wallet instances. return types also infer based on the options passed for the txn return type.

## Package updates

@crossmint/wallets-sdk

### DIFF
--- a/packages/wallets/src/wallets/evm.ts
+++ b/packages/wallets/src/wallets/evm.ts
@@ -1,11 +1,19 @@
-import { createPublicClient, type TypedDataDomain, http } from "viem";
-import type { TypedData, TypedDataDefinition, HttpTransport } from "viem";
-import type { EVMTransactionInput, Transaction } from "./types";
+import {
+    createPublicClient,
+    encodeFunctionData,
+    http,
+    type TypedDataDomain,
+    type TypedData,
+    type TypedDataDefinition,
+    type HttpTransport,
+} from "viem";
+import { isValidEvmAddress } from "@crossmint/common-sdk-base";
+import type { EVMTransactionInput, FormattedEVMTransaction, PreparedTransaction, Transaction } from "./types";
 import { type EVMSmartWalletChain, toViemChain } from "../chains/chains";
 import { Wallet } from "./wallet";
 import type { Chain, EVMChain } from "../chains/chains";
 import { InvalidTypedDataError, SignatureNotCreatedError, TransactionNotCreatedError } from "../utils/errors";
-import { isValidEvmAddress } from "@crossmint/common-sdk-base";
+import type { CreateTransactionSuccessResponse } from "@/api";
 
 export class EVMWallet extends Wallet<EVMChain> {
     constructor(wallet: Wallet<EVMChain>) {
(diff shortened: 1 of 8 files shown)


packages/wallets/src/wallets/wallet.ts
```diff
@@ -446,31 +446,31 @@ export class Wallet<C extends Chain> {
...
+                queryParams.append("cluster", "devnet");
+            }
+            explorerLink = `https://explorer.solana.com/tx/${transactionResponse.onChain.txId}?${queryParams.toString()}`;
         }
-        explorerLink = `https://explorer.solana.com/tx/${transactionResponse.onChain.txId}?${queryParams.toString()}`;
-    }
-    if (explorerLink == null || explorerLink === "") {
-        explorerLink = "Explorer link not available";
+        if (explorerLink == null || explorerLink === "") {
+            explorerLink = "Explorer link not available";
```
AlbertoElias, line 470: Isn't it easier for the developer to parse a null value?

packages/wallets/src/wallets/wallet.ts
```diff
@@ -446,31 +446,31 @@ export class Wallet<C extends Chain> {
...
+        let explorerLink: string | undefined;
+        if (transactionResponse.walletType === "evm-smart-wallet") {
+            explorerLink = transactionResponse.onChain.explorerLink;
+        } else if (transactionResponse.walletType === "solana-smart-wallet") {
+            const queryParams = new URLSearchParams();
+            const env = getEnvironmentForKey(this.apiClient.crossmint.apiKey);
+            if (env !== "production") {
+                queryParams.append("cluster", "devnet");
+            }
+            explorerLink = `https://explorer.solana.com/tx/${transactionResponse.onChain.txId}?${queryParams.toString()}`;
```
AlbertoElias, line 467: The api should now return the explorer link for Solana as well

packages/wallets/src/wallets/evm.ts
```diff
@@ -123,4 +123,47 @@ export class EVMWallet extends Wallet<EVMChain> {
             chain: toViemChain(this.chain),
         });
     }
+
+    private async createTransaction(transaction: FormattedEVMTransaction): Promise<CreateTransactionSuccessResponse> {
```
AlbertoElias, line 127: As the api supports sending multiple transactions, shouldn't we support that too in the sdk?

packages/wallets/src/wallets/evm.ts
```diff
@@ -123,4 +123,47 @@ export class EVMWallet extends Wallet<EVMChain> {
...
+
+        return transactionCreationResponse;
+    }
+
+    private buildTransaction(params: EVMTransactionInput): FormattedEVMTransaction {
+        if ("transaction" in params) {
+            return { transaction: params.transaction };
+        }
+
+        if (params.abi == null) {
```
AlbertoElias, line 147: If abi is null, shouldn't we support that it should either have the value or they are passing in the data directly?

AlbertoElias (commented):
Left some comments around supporting the flexibility that the api has to offer

packages/wallets/src/wallets/evm.ts
jmderby: Confirmed, we're going to add this in a follow up pr only because we currently don't support multiple from the sdk!

packages/wallets/src/wallets/evm.ts
```diff
@@ -37,17 +37,21 @@ export class EVMWallet extends Wallet<EVMChain> {
...
-        return {
-            txId: createdTransaction.id,
-        };
+
+        if (params.options?.experimental_prepareOnly) {
+            return {
+                hash: undefined,
+                explorerLink: undefined,
+                transactionId: createdTransaction.id,
+            } as Transaction<T["options"] extends { experimental_prepareOnly: true } ? true : false>;
```
AlbertoElias, line 51: can this be cleaned up and maybe abstracted away to make the type more declarative?

AlbertoElias: and in here, cant we just pass in true to the type?

AlbertoElias: got something like this from chatgpt:

```
type TransactionReturn<T> = Transaction<IsPrepareOnly<T>>;

public async sendTransaction<T extends EVMTransactionInput>(
  params: T
): Promise<TransactionReturn<T>> {
  const builtTransaction = this.buildTransaction(params);
  const createdTransaction = await this.createTransaction(builtTransaction);

  if (params.options?.experimental_prepareOnly) {
    return {
      hash: undefined,
      explorerLink: undefined,
      transactionId: createdTransaction.id,
    } as Transaction<true>;
  }

  return await this.approveAndWait(createdTransaction.id);
}
```

AlbertoElias (approved):

packages/wallets/src/wallets/evm.ts
jmderby: I wish; i tried that. I battled with this for a while before coming to this as the "cleanest" way to make the return type work. The alternative is to use function overloading but that's far from "clean"

PR #1255, opened 2025-07-03
Wallets sdk: Migrate rest api 2.5

jmderby:
## Description

Migrate sdk to support the new 2.5 rest api.

## Test plan

test all signers types and honestly everything.

## Package updates

@crossmint/wallets-sdk

### DIFF
--- a/.changeset/red-kiwis-behave.md
+++ b/.changeset/red-kiwis-behave.md
@@ -2,4 +2,4 @@
 "@crossmint/client-sdk-react-base": patch
 ---
 
-Simplified signer logic inside getOrCreateWallet
+Fix bug in getOrCreateWallet


packages/wallets/src/signers/email/evm-email-signer.ts
```diff
@@ -11,7 +11,7 @@ export class EvmEmailSigner extends EmailSigner {
     }
 
     locator() {
-        return `evm-keypair:${this.config.signerAddress}`;
+        return `external-wallet:${this.config.signerAddress}`;
```
AlbertoElias, line 14: this should be email. or better even, this.config.locator

packages/wallets/src/wallets/wallet-factory.ts
```diff
@@ -142,16 +122,14 @@ export class WalletFactory {
...
+                        (walletResponse.chainType === "solana" || walletResponse.chainType === "evm") &&
+                        walletResponse.config?.adminSigner.type === "email"
                     )
                 ) {
-                    throw new WalletCreationError("Wallet signer 'email' has no address");
+                    throw new WalletCreationError("Wallet signer 'email' has no email address");
                 }
 
-                const address = walletResponse.config.adminSigner.address;
+                const address = walletResponse.config.adminSigner.email;
```
AlbertoElias, line 132: shouldnt this variable be called email?

packages/wallets/src/wallets/wallet-factory.ts
```diff
@@ -168,53 +146,58 @@ export class WalletFactory {
...
+            };
         }
 
-        return await this.toExternalSignerConfig(chain, signer);
+        return this.toExternalSignerConfig(signer);
     }
 
-    private async toExternalSignerConfig<C extends Chain>(
-        chain: C,
+    private toExternalSignerConfig<C extends Chain>(
```
AlbertoElias, line 175: cant we remove this function as we no longer need to convert to v2 types?

packages/wallets/src/wallets/wallet.ts
```diff
@@ -384,11 +394,7 @@ export class Wallet<C extends Chain> {
 
         do {
             await new Promise((resolve) => setTimeout(resolve, STATUS_POLLING_INTERVAL_MS));
-            signatureResponse = await this.#apiClient.getSignature(
-                // @ts-ignore id type is wrong
-                this.walletLocator,
-                signatureId
-            );
+            signatureResponse = await this.#apiClient.getSignature(this.walletLocator as EvmWalletLocator, signatureId);
```
AlbertoElias, line 397: now we need this type assertion?

biome.json
```diff
@@ -29,6 +29,9 @@
     "organizeImports": {
         "enabled": false
     },
+    "files": {
+        "maxSize": 2097152
+    },
```
jmderby, line 34: increased from default of 1MB to 2MB due to our openapi.json file becoming ~2x the size. 
Required or biome will throw an error.

packages/wallets/src/utils/signer-validation.ts
jmderby, line 1: Instead of checking a complete entire 1:1 object equality with every field matching exactly; we need to ONLY check the existing wallet signers fields match the signer arg passed in. 

for example: 
an existing email signer response might look something like 
```json
{
    "type": "email",
    "email": "jonathan.temp@paella.dev",
    "locator": "email:jonathan.temp@paella.dev"
}
```

but the email signer passed into the sdk might look: 
```json
{
    "type": "email",
    "onAuthRequired": ...,
    "email": "jonathan.temp@paella.dev"
}
```

so we just want to check that it's "type" is "email" and the "email" matches the response.  So this updated compare method just checks the existing signer object against the `newSignerConfig` which is just the signer that gets passed into via args.

packages/wallets/src/wallets/wallet-factory.ts
```diff
@@ -77,12 +64,14 @@ export class WalletFactory {
         walletResponse: GetWalletSuccessResponse,
         args: WalletArgsFor<C>
     ): Wallet<C> {
+        this.validateExistingWalletConfig(walletResponse, args);
```
jmderby, line 67: anytime we create a new wallet instance and use the signer args, we call this to ensure the correct signer validation.

packages/client/ui/react-ui/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -105,7 +105,7 @@ export function CrossmintWalletProvider({
         let onWalletCreationStart = callbacks?.onWalletCreationStart;
         let onTransactionStart = callbacks?.onTransactionStart;
 
-        if (createOnLogin?.signer.type === "passkey" && showPasskeyHelpers) {
+        if (createOnLogin?.signer?.type === "passkey" && showPasskeyHelpers) {
```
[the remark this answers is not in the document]
jmderby: reverted, that change was when i was testing

packages/client/react-base/src/providers/CrossmintWalletBaseProvider.tsx
jmderby, line 1: We were running into issues with calling getOrCreateWallet before the custom auth was set and required a janky "fix". But the true fix was to move the logic inside the wallet sdk. Since the wallet sdk has access to the latest `experimental_customAuth` value, we can use it there instead. 

I've left a comment where this was moved to here: https://github.com/Crossmint/crossmint-sdk/pull/1255/files#r2186080719

packages/wallets/src/wallets/wallet-factory.ts
```diff
@@ -168,108 +147,64 @@ export class WalletFactory {
...
-                type: "evm-passkey",
-                id: existingWallet.config.adminSigner.id,
-                name: existingWallet.config.adminSigner.name,
-                publicKey: {
-                    x: existingWallet.config.adminSigner.publicKey.x,
-                    y: existingWallet.config.adminSigner.publicKey.y,
-                },
-            };
+        if (args.signer.type === "external-wallet" && experimental_customAuth?.externalWalletSigner != null) {
+            args.signer = experimental_customAuth.externalWalletSigner as SignerConfigForChain<C>;
```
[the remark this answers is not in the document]
jmderby: This fixes any render issues in react when calling getOrCreateWallet before the `experimental_customAuth` object contains the email or external-wallet signers!

packages/client/react-base/src/providers/CrossmintWalletBaseProvider.tsx
```diff
@@ -72,10 +72,11 @@ export function CrossmintWalletBaseProvider({
                 if (args?.signer?.type === "email") {
                     const email = args.signer.email ?? experimental_customAuth?.email;
                     const _onAuthRequired = args.signer.onAuthRequired ?? onAuthRequired;
+
                     if (email == null) {
-                        throw new Error(
-                            "Email not found in customAuth or signer. Please set email in customAuth or signer."
-                        );
+                        // Hacky way to re-render when email gets set from crossmint auth (if applicable)
```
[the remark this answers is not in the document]
jmderby: i've came up with a true simple solution, comments around my changes below!

biome.json
aigustin: why did it become 2x the size? can't we remove the old endpoints for this?

apps/wallets/quickstart-devkit/app/home.tsx
jmderby, line 1: I also added a firebase component so we can test firebase BYOA too!

PR #1266, opened 2025-07-10
Wallets sdk: Add getTransaction method

jmderby:
## Description

Expose getTransaction method 

## Test plan

tested method manually 

## Package updates

@crossmint/wallets-sdk

### DIFF
--- a/packages/wallets/src/wallets/wallet.ts
+++ b/packages/wallets/src/wallets/wallet.ts
@@ -174,7 +174,7 @@ export class Wallet<C extends Chain> {
      * Get the wallet transactions
      * @returns The transactions
      */
-    public async transactions() {
+    public async experimental_transactions() {
         return await this.#apiClient.getTransactions(this.walletLocator);
     }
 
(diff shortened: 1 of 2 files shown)


packages/wallets/src/wallets/wallet.ts
```diff
@@ -174,10 +174,18 @@ export class Wallet<C extends Chain> {
      * Get the wallet transactions
      * @returns The transactions
      */
-    public async experimental_transactions() {
+    public async transactions() {
```
AlbertoElias, line 177: this should remain experimental. our transactions endpoint only exposes the outward txs triggered via crossmint which could be confusing

AlbertoElias (approved):

packages/wallets/src/wallets/wallet.ts
jmderby: cc: @0xaguspunk

aigustin: lets do the same with transaction then

AlbertoElias: transaction is fine as non experimental because it fetches with a crossmint id, so it can only fetch a crossmint transaction issued by this wallet

PR #1267, opened 2025-07-10
wallets sdk: fix activity api version

jmderby:
## Description

Rollback to older activity api version. 

## Test plan

tested

## Package updates

@crossmint/wallets-sdk


packages/wallets/src/api/client.ts
```diff
@@ -141,9 +142,13 @@ class ApiClient extends CrossmintApiClient {
     }
 
     async experimental_activity(walletLocator: WalletLocator, params: { chain: Chain }): Promise<GetActivityResponse> {
+        let legacyLocator = walletLocator;
+        if (!this.isServerSide) {
```
AlbertoElias, line 146: why? we also need to convert for server side

AlbertoElias (approved):

PR #1272, opened 2025-07-10
WAL-5336: safe url parsing

maxisch:
## Description

Adding safe url parsing


devlyn37 (approved):
Fix LGTM, if we're not using this anywhere we could also just delete the code

maxisch:
> Fix LGTM, if we're not using this anywhere we could also just delete the code

Sorry, meant that we are not using it in our devkit flow that tests e2e, but it's probably used somewhere else in crossmint's product suite

PR #1282, opened 2025-07-15
WAL-5344: Harden transport for signers flow

maxisch:
## Description

We add a specific window transport for our ncs flow, with a tighter security control, as recommended by the auditors.

For context if the current changes were applied to the WindowTransport class, then our social auth flows would break.

## Test plan

Tested e2e ncs flows, and social login flows.


packages/client/window/src/transport/SignersWindowTransport.ts
maxisch, line 1: open to discuss if this should live here, or in the signers directory.
If the later, we would have to export WindowTransport as it's currently only visible within the current package

devlyn37 (approved):

PR #1287, opened 2025-07-16
React UI: exposes experimental_loginWithOauth

AlbertoElias:
## Description

Right now setting up headless oauth login is a pain because we need all our client-window listeners. This pr exposes that login method

## Test plan

EVM: Crossmint auth and BYOA (privy)

## Package updates

react-ui: patch


packages/client/ui/react-ui/src/providers/CrossmintAuthProviderInternal.tsx
```diff
@@ -39,7 +50,6 @@ export function CrossmintAuthProviderInternal({
     children,
     appearance,
     termsOfServiceText,
-    prefetchOAuthUrls = true,
```
jmderby, line 42: whats the main reason for removing this?

AlbertoElias: we just moved it to a different place

AlbertoElias:
@jmderby I tested evm: crossmint auth, crossmint auth with this new function and privy auth. I havent tested solana because it's broken anyway. But these all worked fine

PR #1288, opened 2025-07-16
Back to 6 digits otp

maxisch:
## Description

In case we need to release our SDK before the latest TEE code lands in staging, we need to merge this.
Once the TEE starts sending 9 digit OTPs, we have to revert this and release another version

## Test plan

Manually verified the modal expect 6 digits again


jmderby:
Closing for now

PR #1296, opened 2025-07-17
Wallets sdk: Improve error when missing onAuthRequired

jmderby:
## Description

Clients are running into issues when setting up email signers, this pr attempts to improve the messaging around setting up the onAuthRequired method when not using our react/native package.

## Test plan

Error message shows up as expected when encounter issue.

## Package updates

@crossmint/wallets-sdk

### DIFF
--- a/packages/wallets/src/signers/non-custodial/ncs-signer.ts
+++ b/packages/wallets/src/signers/non-custodial/ncs-signer.ts
@@ -36,6 +36,14 @@ export abstract class NonCustodialSigner implements Signer {
 
     protected async handleAuthRequired() {
         if (this.config.clientTEEConnection == null) {
+            if (this.config.onAuthRequired == null) {
+                throw new Error(
+                    `${this.type} signer requires the onAuthRequired callback to handle OTP verification. ` +
+                        `This callback manages the authentication flow (sending OTP and verifying user input). ` +
+                        `If using our React/React Native SDK, this is handled automatically by the provider. ` +
+                        `For other environments, implement: onAuthRequired: (needsAuth, sendOtp, verifyOtp, reject) => { /* your UI logic */ }`
+                );
+            }
             throw new Error("Handshake parent not initialized");
         }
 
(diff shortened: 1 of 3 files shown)


packages/wallets/src/signers/non-custodial/ncs-signer.ts
```diff
@@ -36,6 +36,14 @@ export abstract class NonCustodialSigner implements Signer {
...
     protected async handleAuthRequired() {
         if (this.config.clientTEEConnection == null) {
+            if (this.config.onAuthRequired == null) {
+                throw new Error(
+                    `${this.type} signer requires the onAuthRequired callback to handle OTP verification. ` +
+                        `This callback manages the authentication flow (sending OTP and verifying user input). ` +
+                        `If using our React/React Native SDK, this is handled automatically by the provider. ` +
+                        `For other environments, implement: onAuthRequired: (needsAuth, sendOtp, verifyOtp, reject) => { /* your UI logic */ }`
+                );
+            }
```
jmderby, line 46: ideally we link to some documentation in our error message but right now we don't have any helpful docs around this implementation.

AlbertoElias:
that's pretty useful, this is what i shared with clients;

```
const wallet = await crossmintClient.getOrCreateWallet({
        chain: 'solana',
        signer: {
          type: 'email',
          email: user.email, // you don't need this if you're setting the email in setCustomAuth
        },
        onAuthRequired: async (needsAuth, sendEmailWithOtp, verifyOtp, reject) => {
            if (needsAuth) {
                await sendEmailWithOtp(user.email);
                // Wait for user to input the otp, and once you have it, call =>
                await verifyOtp(otp);
            }
        },
});
onAuthRequired is a callback that gets called when you're going to sign a transaction. We give you certain values. needsAuth tells you if the email signer has been set up on this device. sendEmailWithOtp is a helper function that will send the email with the otp to the user. verifyOtp accepts that otp, validates it, and will set up the email signer on the device and then the transaction gets signed and sent
```

in case any of those explanatinos help

PR #1299, opened 2025-07-18
Fix tailwind override issue

jmderby:
## Description

**Problem**: Client applications using Crossmint's React UI components were experiencing CSS conflicts where their custom Tailwind configurations would override Crossmint component styles. For example, a client's custom `text-sm: 18px` configuration would override Crossmint's default `text-sm: 14px`, breaking the component UI in both their app and our component.

**Solution**: Implemented Twind's "library mode" to create completely isolated CSS for Crossmint components that cannot be overridden by client application styles. This includes:

1. **Isolated Stylesheet**: Using `@twind/core`'s `cssom()` to create a separate stylesheet with `data-crossmint-sdk` attribute that exists independently from client styles
2. **CSS Cascade Layers**: Implemented `@layer` declarations to ensure Crossmint styles take precedence over client styles in the CSS cascade
3. **Class Scoping**: Added `cm-` prefixing to all generated classes to prevent naming conflicts
4. **SSR Compatibility**: Added proper server-side rendering checks to prevent build-time errors

**Goal**: Ensure Crossmint components have truly isolated styling that renders consistently regardless of the client application's Tailwind configuration or custom CSS rules.

## Technical Implementation

- Modified `twind-instance.ts` to use library mode with isolated DOM stylesheet creation
- Implemented CSS cascade layers with `@layer base, client, crossmint;` declaration  
- Added class name hashing to prefix all utilities with `cm-` (e.g., `text-sm` becomes `cm-text-sm`)
- Ensured compatibility with both client-side and server-side rendering

## Test Plan

**Testing Environments**:
- Monorepo devkit application  
- External Next.js application via yalc package linking

**Test Scenarios**:
1. **Custom Tailwind Configuration**: Modified client `tailwind.config.js` to override utilities like `text-sm` and `bg-black` - verified client styles apply locally but don't affect Crossmint components
2. **CSS Specificity Conflicts**: Added conflicting CSS rules in client `globals.css` - verified Crossmint components maintain correct styling through cascade layer precedence.
3. **Cross-Environment Testing**: Confirmed functionality in both monorepo and external package built scenarios

**Verification**: 
- `style[data-crossmint-sdk]` element appears in DevTools with isolated Crossmint styles
- Client applications can safely customize Tailwind utilities without breaking Crossmint UI
- Crossmint components display correctly regardless of client Tailwind configuration

## Package Updates

**@crossmint/client-sdk-react-ui**

This is a significant technical improvement that solves a critical integration issue for client applications using Crossmint components.

### DIFF
--- a/.changeset/chilled-queens-double.md
+++ b/.changeset/chilled-queens-double.md
@@ -1,5 +1,0 @@
----
-"@crossmint/wallets-sdk": patch
----
-
-Improve error when missing onAuthRequired callback
--- a/apps/auth/nextjs-ssr/CHANGELOG.md
+++ b/apps/auth/nextjs-ssr/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @crossmint/server-sdk-next-starter
 
+## 0.4.95
+
+### Patch Changes
+
+- Updated dependencies [7a12798]
+  - @crossmint/client-sdk-react-ui@2.2.17
+  - @crossmint/server-sdk@1.2.35
+
 ## 0.4.94
 
 ### Patch Changes
--- a/apps/auth/nextjs-ssr/package.json
+++ b/apps/auth/nextjs-ssr/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@crossmint/auth-ssr-nextjs-demo",
-    "version": "0.4.94",
+    "version": "0.4.95",
     "private": true,
     "scripts": {
         "dev": "next dev",
--- a/apps/auth/node-and-express-server/CHANGELOG.md
+++ b/apps/auth/node-and-express-server/CHANGELOG.md
@@ -1,5 +1,11 @@
 # crossmint-auth-node
 
+## 1.1.46
+
+### Patch Changes
+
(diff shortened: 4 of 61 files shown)


packages/client/ui/react-ui/src/twind-instance.ts
jmderby, line 1: This file required a weird workaround as Library Mode didn't work out of the box for some reason.  https://twind.style/library-mode

but the fact that twind is an abandoned project now, i'm not surprised.

packages/client/ui/react-ui/src/utils/classNames.ts
```diff
@@ -1,12 +1,25 @@
...
  * @param inputs - The class names to combine.
  * @returns The combined class names as a string.
+ * applies internal twind instance to the class names
  */
 export function classNames(...inputs: ClassValue[]) {
-    return twMerge(clsx(inputs));
+    const processedInputs = inputs
+        .filter(Boolean)
+        .map((input) => {
+            if (typeof input === "string") {
```
AlbertoElias, line 16: what can input be if it's not a string?

PR #1313, opened 2025-07-24
feat: unifies api for both otp signers

AlbertoElias:
## Description

Our onAuthRequired callbacks were specific to email signer, this uses the `OtpSigner` naming to bring both email and phone under the same umbrella

## Test plan

Tested with demo apps

## Package updates

react-ui: minor
react-native: minor
wallets: minor

### DIFF
--- a/.changeset/breezy-mirrors-dance.md
+++ b/.changeset/breezy-mirrors-dance.md
@@ -1,0 +1,7 @@
+---
+"@crossmint/client-sdk-react-native-ui": minor
+"@crossmint/client-sdk-react-ui": minor
+"@crossmint/wallets-sdk": minor
+---
+
+Unifies callback calls for otp signers


apps/wallets/smart-wallet/expo/app/index.tsx
```diff
@@ -15,7 +16,7 @@ export default function Index() {
     const { experimental_customAuth } = useCrossmint();
     const loggedInUserEmail = experimental_customAuth?.email ?? null;
     const { wallet, getOrCreateWallet, status: walletStatus } = useWallet();
-    const { needsAuth, sendEmailWithOtp, verifyOtp } = useWalletEmailSigner();
+    const { needsAuth, sendOtp, verifyOtp } = useWalletOtpSigner();
```
alfonso-paella, line 19: needsAuth -- dont use Auth, name it something like isInitialized (and invert it)

packages/client/ui/react-native/src/hooks/useWalletOtpSigner.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function useWalletOtpSigner(): OtpSignerFunctions {
+    const context = useContext(CrossmintWalletEmailSignerContext);
+
+    if (context == null) {
+        throw new Error("useWalletOtpSigner must be used within CrossmintWalletProvider");
+    }
+
+    if (!context.sendOtp || !context.verifyOtp || !context.reject) {
+        throw new Error("Otp signer functions are not available. Make sure you're using an otp signer wallet.");
```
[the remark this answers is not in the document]
alfonso-paella: Somehting like "The wallet in context doesn't have any OTP signer configured. Register an OTP signer first before initializing this hook."

apps/wallets/smart-wallet/expo/app/index.tsx
AlbertoElias, line 19: `walletSignerNeedsSetup`
`otpSignerNeedsSetup`
`otpSignerInitialized`

alfonso-paella, line 19: I may be wrong but iirc its an anti pattern to have read and write properties in the same hook

PR #1319, opened 2025-07-29
bug: useRef for needsAuth to have latest state

AlbertoElias:
## Description

we need to make sure needsAuth is always up to date, so we also call onAuthRequired to trigger an update after verifying otp

## Test plan

tested with quickstart

## Package updates

react-ui: patch
react-native: patch
wallets: patch


jmderby (approved):
nice, uses same pattern as other refs so this makes sense.

PR #1326, opened 2025-07-31
Payments: Add more errors

VeniaminC:
## Description
Add more errors 
## Test plan
just adding thing to a const
## Package updates
```
---
"@crossmint/common-sdk-base": patch
---

Adding more payment error codes
```

### DIFF
--- a/packages/common/base/src/types/errors.ts
+++ b/packages/common/base/src/types/errors.ts
@@ -1,6 +1,7 @@
 export const PaymentErrors = {
     MINT_CONFIG_INVALID: "payments:mint-config.invalid",
     PAYMENT_METHOD_INVALID: "payments:payment-method.invalid",
+    EMAIL_INVALID: "payments:email.invalid",
     CLIENT_ID_INVALID: "payments:client-id.invalid",
     REQUIRED_RECIPIENT: "payments:recipient.required",
     MINTING_CONTRACT_MISSING: "payments:minting-contract.missing",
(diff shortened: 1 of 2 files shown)


packages/common/base/src/types/errors.ts
```diff
@@ -26,6 +25,17 @@ export const PaymentErrors = {
...
+    PAYMENT_AMOUNT_MISSING: "payment_amount_missing",
+    PAYMENT_SINGLE_PURCHASE_LIMIT_EXCEEDED: "payment_single_purchase_limit_exceeded",
+    PAYMENT_DAILY_TRANSACTION_LIMIT_EXCEEDED: "payment_daily_transaction_limit_exceeded",
+    TOKEN_DAILY_VOLUME_LIMIT_EXCEEDED: "token_daily_volume_limit_exceeded",
+    WALLET_ADDRESS_MISSING: "wallet_address_missing",
+    EMAIL_INVALID: "email_invalid",
+    PROJECT_ADDON_REQUIRED: "project_addon_required",
+    ASSET_NOT_AVAILABLE_IN_ENVIRONMENT: "asset_not_available_in_environment",
+    UNSUPPORTED_TOKEN: "unsupported_token",
+    PROJECT_FEATURE_NOT_ENABLED: "project_feature_not_enabled",
```
mPaella, line 38: why diverging the formats from all the existing codes...?

VeniaminC: https://docs.google.com/document/d/1Gc5jiB5sdnAx_qk9D-zBFLmAYUrUiuf8OknC8KhW0Sw/edit?tab=t.0

```
reason: snake_case_reason
```

mPaella: i think its too late to do that here, consistency is better than randomly starting snake now
and using namespaces like such `payments:limit-exceeded.daily-volume` is good

PR #1332, opened 2025-08-05
bug: Reverts back to needsAuth being a state

AlbertoElias:
## Description

Have to revert back to needsAuth being a state. When it's a ref, it doesn't update the state so remains false

## Test plan

Tested with demo expo app
## Package updates

wallets: patch
rn-ui: patch


apps/wallets/quickstart-devkit/components/approval-test.tsx
jmderby, line 1: this component will kinda break with stellar as were doing ternary operators for sol and evm. but all good this is also a testing component so anyone using stellar can just update, or we can fix later

PR #1359, opened 2025-08-22
feat: Adds Flow

AlbertoElias:
## Description

We are adding support for Flow in the API. This needs to be merged after.

## Test plan

No tests added

## Package updates

common-base: patch
wallets: patch

### DIFF
--- a/packages/wallets/src/api/types.ts
+++ b/packages/wallets/src/api/types.ts
@@ -13,7 +13,7 @@ import type {
     WalletNftsResponseDto,
     WalletV1Alpha2ErrorDto,
     WalletV1Alpha2TransactionErrorDto,
-    WalletV2025ResponseDtoReadable,
+    WalletV2025ResponseDto,
     WalletsMultipleTransactionV2025ResponseDto,
     WalletsSendTokenControllerSendToken2Response,
     WalletsSignatureV2025ResponseDto,
@@ -30,7 +30,7 @@ import type {
 } from "./gen/types.gen";
 
 export type CreateWalletParams = CreateWalletV2025Dto;
-export type GetWalletSuccessResponse = WalletV2025ResponseDtoReadable;
+export type GetWalletSuccessResponse = WalletV2025ResponseDto;
 export type CreateWalletResponse = GetWalletSuccessResponse | WalletV1Alpha2ErrorDto;
 export type GetWalletResponse = GetWalletSuccessResponse | WalletV1Alpha2ErrorDto;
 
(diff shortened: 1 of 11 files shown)


maxisch (approved):

packages/wallets/src/api/types.ts
```diff
@@ -13,7 +13,7 @@ import type {
     WalletNftsResponseDto,
     WalletV1Alpha2ErrorDto,
     WalletV1Alpha2TransactionErrorDto,
-    WalletV2025ResponseDtoReadable,
```
AlbertoElias, line 16: seems like it got renamed

packages/wallets/src/wallets/stellar.ts
```diff
@@ -65,7 +65,7 @@ export class StellarWallet extends Wallet<StellarChain> {
                     type: "contract-call",
                     contractId,
                     method,
-                    memo,
+                    memo: memo != null ? { type: "text", value: memo } : undefined,
```
AlbertoElias, line 68: @alberto-crossmint does this make sense? we were passing a string but the input now requires an object. only handling the `text` type

packages/wallets/src/wallets/wallet.ts
```diff
@@ -493,12 +493,8 @@ export class Wallet<C extends Chain> {
                 if (signer == null) {
                     throw new InvalidSignerError(`Signer ${pendingApproval.signer} not found in pending approvals`);
                 }
-                const transactionToSign =
```
AlbertoElias, line 496: does removign this make sense? @alberto-crossmint

alberto-crossmint: Hmmmm I don't think these two values are equal!

A Solana VersionedTransaction is composed by the _message_ (instructions, blockhash, sender,...) and then some other stuff as the already present signatures in the tx. What we have under `pendingApproval.message` is the serialization of this message, not the whole transaction, so the objects would be different and this would not work unless we change the implementation in `signTransaction`.

alberto-crossmint: So the fact that we're passing a whole transaction to this method, though it's convenient (once deserialized, this VersionedTransaction has a [sign](https://github.com/solana-foundation/solana-web3.js/blob/c26c13bf841821d9bbf83bb476c567fe29d13821/src/transaction/versioned.ts#L94) method that we can call to sign the transaction), it's not clean. We assume the payload the signer should sign is the one in approval.message, and that's the flow we have regression tests for. We should move towards that when we can.

 So, even when you're approach here about using `approval.message` is correct!, it wouldn't work unless we change the signTransaction method to mimic [the Solana impl](https://github.com/solana-foundation/solana-web3.js/blob/c26c13bf841821d9bbf83bb476c567fe29d13821/src/transaction/versioned.ts#L9), and I think this deserves its own PR!

PR #1366, opened 2025-08-27
Wallets TS SDK: Add Delegated Signer During Wallet Creation for Solana

devlyn37:
## Description

This PR adds the ability to specify delegated signers during wallet creation for Solana Smart wallets. To match the `addDelegateSigner` method, doing so looks like:

```TS
const wallet = await crossmintWallets.createWallet({
    chain: "solana",
    signer: {
        type: "email",
        email: "test@tester.com",
    },
    delegatedSigners: [{signer: "external-wallet:EbXL4e6XgbcC7s33cD5EZtyn5nixRDsieBjPQB7zf448"}]
});
```

Originally we discussed having the `delegatedSigners` array populated with objects similar to the one in `signer`, but that would be diverging from the types we already have established for `addDelegatedSigner` and `getDelegatedSigner`, which we should avoid for now. Improving delegated signer types is a separate discussion.

The configuration validation behavior when calling `getOrCreateWallet` or `getWallet` with the `delegatedSigners` option specified is that if there's an existing wallet fetched, it must have matching delegated signer config.

These are minimal changes to release for a client commitment, further work in this area includes:
- Adding this for different chains (EVM, Stellar)
- Supporting this new parameter in `CrossmintWalletBaseProvider` for React and RN packages.

### Local Testing

Tested the following cases:
- Happy path `createWallet`: https://www.loom.com/share/383c28c506ec4893bd96384a309b9d14
- Calling `getWallet` for an existing wallet with a different delegated signer configuration to confirm the error is handled correctly:

Screenshot 2025-08-27 at 1 14 53 AM

### Unit Tests

This PR adds extensive unit testing for delegate signer configuration matching in `validateExistingWalletConfig`

## Package updates

Patch `@crossmint/wallets-sdk`

### DIFF
--- a/packages/wallets/src/wallets/wallet-factory.ts
+++ b/packages/wallets/src/wallets/wallet-factory.ts
@@ -17,7 +17,7 @@ import type { DelegatedSigner, WalletArgsFor, WalletOptions } from "./types";
 import { compareSignerConfigs } from "../utils/signer-validation";
 
 const DELEGATED_SIGNER_MISMATCH_ERROR =
-    "When the 'delegatedSigners' argument is provided to a method that fetches an existing wallet, that wallet must have matching delegated signers.";
+    "When 'delegatedSigners' is provided to a method that may fetch an existing wallet, each specified delegated signer must exist in that wallet's configuration.";
 
 export class WalletFactory {
     constructor(private readonly apiClient: ApiClient) {}
@@ -55,10 +55,6 @@ export class WalletFactory {
         await args.options?.experimental_callbacks?.onWalletCreationStart?.();
 
         this.mutateSignerFromCustomAuth(args, true);
-
-        if (args.delegatedSigners != null && args.chain !== "solana") {
-            throw new WalletCreationError("Delegated signers are only supported for Solana smart wallets");
-        }
 
         const adminSigner =
             args.signer.type === "passkey" ? await this.createPasskeyAdminSigner(args.signer) : args.signer;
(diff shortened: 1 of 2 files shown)


packages/wallets/src/wallets/types.ts
```diff
@@ -101,6 +101,7 @@ export type WalletArgsFor<C extends Chain> = {
     owner?: string;
     plugins?: WalletPlugin<C>[];
     options?: WalletOptions;
+    delegatedSigners?: C extends "solana" ? Array<DelegatedSigner> : never;
```
AlbertoElias, line 104: Why limit to Solana though? We support this for all chains

packages/wallets/src/wallets/wallet-factory.ts
```diff
@@ -6,14 +6,18 @@ import type {
...
 import type { InternalSignerConfig, SignerConfigForChain } from "../signers/types";
 import { Wallet } from "./wallet";
 import { assembleSigner } from "../signers";
-import type { WalletArgsFor, WalletOptions } from "./types";
-import { compareSignerConfigs } from "@/utils/signer-validation";
+import type { DelegatedSigner, WalletArgsFor, WalletOptions } from "./types";
+import { compareSignerConfigs } from "../utils/signer-validation";
+
+const DELEGATED_SIGNER_MISMATCH_ERROR =
+    "When the 'delegatedSigners' argument is provided to a method that fetches an existing wallet, that wallet must have matching delegated signers.";
```
AlbertoElias, line 20: Why introduce this limitation? I think it should be okay if we allow only passing the delegated signers on original creation. Especially if this is used server side where the function we support is only createWallet

packages/wallets/src/wallets/wallet-factory.ts
```diff
@@ -52,6 +56,10 @@ export class WalletFactory {
 
         this.mutateSignerFromCustomAuth(args, true);
 
+        if (args.delegatedSigners != null && args.chain !== "solana") {
```
AlbertoElias, line 59: Would remove

packages/wallets/src/wallets/wallet-factory.ts
```diff
@@ -257,6 +268,48 @@ export class WalletFactory {
...
+        }
+
+        for (const argSigner of inputDelegatedSigners) {
+            const matchingExistingSigner = existingDelegatedSigners.find(
+                (existingSigner) => existingSigner.locator === argSigner.signer
+            );
+
+            if (matchingExistingSigner == null) {
+                const walletSigners = existingDelegatedSigners.map((s) => s.locator).join(", ");
+                throw new WalletCreationError(
```
AlbertoElias, line 308: This error does make sense. If the input does not have delegatedSigners and a wallet exists with them, should be okay. If the input has them, they need to match. Not match exactly though, because new signers could have been added after the fact

packages/wallets/src/wallets/types.ts
devlyn37: > These are minimal changes to release for a client commitment, further work in this area includes:
> Adding this for different chains (EVM, Stellar)

Making scope here as tight as possible to ship asap. Otherwise I'd need to spend the time testing all the other chains, and dealing with EVM's seemingly distinct signer format when specifying delegated signers on creation

packages/wallets/src/wallets/wallet-factory.ts
AlbertoElias: related yes

packages/wallets/src/wallets/wallet-factory.ts
devlyn37: adjusted

packages/wallets/src/wallets/wallet-factory.ts
devlyn37: Ok, then leaving for now!

AlbertoElias (approved):

PR #1372, opened 2025-08-28
feat: deprecates experimental_setCustomAuth in favour of setJwt

AlbertoElias:
## Description

We are going to be removing setCustomAuth in v1 of the sdk, this adds a deprecation notice and makes sure developers can start using setJwt 

## Test plan

Tested with quickstart using Firebase

## Package updates

react-base: patch


jcurbelo (approved):
lgtm

PR #1376, opened 2025-09-02
feat: Add getPublicKey method to NonCustodialSigner

jcurbelo:
## Description

Adds a new `getPublicKey()` method to the `NonCustodialSigner` abstract class that exposes secp256k1 public keys from non-custodial signers. This enables developers to retrieve public keys for verification and other cryptographic operations without needing to perform signing operations.

The implementation follows the existing pattern used in the codebase by leveraging the `get-status` event to retrieve signer information, then extracting the secp256k1 public key from the response.

**Note**: This is a work-in-progress implementation that will be refined locally.

**Link to Devin run**: https://app.devin.ai/sessions/30eef6d4bee94aa2b1c6b14a1c4eb35a

**Requested by**: @jcurbelo

## Test plan

⚠️ **No testing performed yet** - this is a draft implementation that requires:
- Unit tests to verify the method works correctly with valid responses
- Error handling tests for uninitialized connections and failed API calls
- Integration tests to ensure the API contract matches expectations
- Manual testing across different signer states

## Package updates

No package dependencies were added or modified.

## Human Review Checklist

**Critical items to verify:**
- [ ] **Type safety**: Confirm that the response type actually contains `publicKeys.secp256k1` property - there were TypeScript compilation errors suggesting type mismatches
- [ ] **API contract**: Verify that the `get-status` response reliably includes the `publicKeys` object when `signerStatus` is "ready"
- [ ] **Build verification**: Ensure TypeScript compilation passes after these changes
- [ ] **Error scenarios**: Test behavior when TEE connection is uninitialized, when signer is not ready, or when API calls fail
- [ ] **Authentication**: Verify the `experimental_customAuth` usage is correct and won't cause runtime errors

**Risk factors:**
- New untested code with assumptions about API response structure
- TypeScript type definition mismatches observed during development
- Dependency on experimental authentication features
- Generic error handling that may not cover all edge cases

### DIFF
--- a/packages/wallets/src/signers/non-custodial/ncs-signer.ts
+++ b/packages/wallets/src/signers/non-custodial/ncs-signer.ts
@@ -41,6 +41,37 @@ export abstract class NonCustodialSigner implements Signer {
 
     abstract signTransaction(transaction: string): Promise<{ signature: string }>;
 
+    /**
+     * Retrieves the public key from the non-custodial signer
+     * @returns The secp256k1 public key
+     * @throws Error if handshake parent is not initialized or if retrieval fails
+     */
+    public async getPublicKey(): Promise<string> {
+        if (this.config.clientTEEConnection == null) {
+            throw new Error("Handshake parent not initialized");
+        }
+
+        const handshakeParent = this.config.clientTEEConnection;
+
+        const response = await handshakeParent.sendAction({
+            event: "request:get-status",
+            responseEvent: "response:get-status",
+            data: {
+                authData: {
+                    jwt: this.config.crossmint.experimental_customAuth?.jwt ?? "",
+                    apiKey: this.config.crossmint.apiKey,
+                },
+            },
+            options: DEFAULT_EVENT_OPTIONS,
+        });
+
+        if (response?.status === "success" && response.signerStatus === "ready") {
+            return response.publicKeys.secp256k1;
+        }
+
+        throw new Error("Failed to get public key");
+    }
+
     protected async handleAuthRequired() {
         if (this.config.clientTEEConnection == null) {
             if (this.config.onAuthRequired == null) {


packages/wallets/src/signers/non-custodial/ncs-signer.ts
```diff
@@ -41,6 +47,52 @@ export abstract class NonCustodialSigner implements Signer {
...
+                authData: {
+                    jwt: this.config.crossmint.experimental_customAuth?.jwt ?? "",
+                    apiKey: this.config.crossmint.apiKey,
+                },
+            },
+            options: DEFAULT_EVENT_OPTIONS,
+        });
+
+        if (response?.status !== "success") {
+            console.error("[getPublicKey] Failed to get signer status:", response);
```
alberto-crossmint, line 69: I'd rephrase this error, as someone may be confused about why we're getting the signer status if they're trying to fetch the public key. Maybe something along the lines of "Failed to get signer public key"

packages/client/react-base/src/providers/CrossmintWalletBaseProvider.tsx
```diff
@@ -53,6 +57,14 @@ export function CrossmintWalletBaseProvider({
     const [wallet, setWallet] = useState<Wallet<Chain> | undefined>(undefined);
     const [walletStatus, setWalletStatus] = useState<"not-loaded" | "in-progress" | "loaded" | "error">("not-loaded");
 
+    const getSignerPublicKey = useCallback(async (): Promise<SignerPublicKey | null> => {
+        const signer = wallet?.signer;
+        if (!(signer instanceof NonCustodialSigner)) {
+            return null;
+        }
```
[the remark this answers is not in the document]
alberto-crossmint: That will ensure, that this is only called (and raise an error otherwise) for NCSs

packages/wallets/src/signers/non-custodial/ncs-signer.ts
```diff
@@ -41,6 +47,52 @@ export abstract class NonCustodialSigner implements Signer {
...
+
+        if (response?.status !== "success") {
+            console.error("[getPublicKey] Failed to get signer public key:", response);
+            throw new Error("Failed to get signer public key");
+        }
+
+        if (response.signerStatus === "ready") {
+            this._needsAuth = false;
+
+            if (!response.publicKeys || Object.keys(response.publicKeys).length === 0) {
```
alberto-crossmint, line 76: We usually do `response.publicKeys != null`. Not really important, just to avoid implicit comparison behavior, as a style pattern

packages/wallets/src/signers/non-custodial/ncs-signer.ts
```diff
@@ -41,6 +47,52 @@ export abstract class NonCustodialSigner implements Signer {
...
+
+        if (response.signerStatus === "ready") {
+            this._needsAuth = false;
+
+            if (response.publicKeys == null || Object.keys(response.publicKeys).length === 0) {
+                console.error("[getPublicKey] No public keys found in response:", response);
+                throw new Error("No public keys found in response");
+            }
+
+            return response.publicKeys;
```
alberto-crossmint, line 81: I don't think that, from the user perspective, it makes sense to return all public keys! For example, if they're using a EVM signer, returning a ed25519 key feels extremely unexpected, and may trigger some questions. Can we just make this method protected and, for each signer type, return the appropriate key type?

PR #1378, opened 2025-09-02
fix: Make export-signer events a specific type 

AlbertoElias:
## Description

We want the index and export page to handle different events, so we separate it here. Export should only listen to the export event and it should not be accessible from index.

## Test plan

Tested with the react apps

## Package updates

client-signers: patch

### DIFF
--- a/packages/client/signers/src/communications/schemas.ts
+++ b/packages/client/signers/src/communications/schemas.ts
@@ -132,7 +132,7 @@ export const ExportSignerPayloadSchema = {
                     .union([z.literal("ed25519"), z.literal("secp256k1")])
                     .describe("The cryptographic scheme to use"),
                 encoding: z
-                    .union([z.literal("base58"), z.literal("hex")])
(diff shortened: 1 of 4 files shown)


packages/client/signers/src/communications/schemas.ts
```diff
@@ -132,7 +132,7 @@ export const ExportSignerPayloadSchema = {
                     .union([z.literal("ed25519"), z.literal("secp256k1")])
                     .describe("The cryptographic scheme to use"),
                 encoding: z
-                    .union([z.literal("base58"), z.literal("hex")])
+                    .union([z.literal("base58"), z.literal("hex"), z.literal("strkey")])
```
devlyn37, line 135: Based on (this)[https://github.com/Crossmint/open-signer/pull/82#discussion_r2308971290] comment chain, maybe worth removing this now to prevent needing another SDK bump, but non-blocking!

devlyn37 (approved):

PR #1410, opened 2025-09-26
React Native sdk: add ui for signing flow

jmderby:
## Description

Support crossmint built-in ui for our react native signing flow, provide identical experience as our react web flow.

https://www.loom.com/share/1ca43d6f78fa4a2f9188772fead7a823

## Test plan

- Make sure changes to wallet provider work for both react web and native
- Make sure headless and non-headless mode works as expected via`headlessSigningFlow` 
- Make backwards compatible with no breaking changes (we default to headless ui) and follow up with a v1 release that will default to our ui.
- Tested adding a custom color theme via `appearance` prop 

## Package updates

@crossmint/client-sdk-react-ui-native

### DIFF
--- a/packages/client/react-base/src/providers/CrossmintWalletUIBaseProvider.tsx
+++ b/packages/client/react-base/src/providers/CrossmintWalletUIBaseProvider.tsx
@@ -76,7 +76,7 @@ export function CrossmintWalletUIBaseProvider({
     children,
     appearance,
     createOnLogin,
-    headlessSigningFlow = false,
+    headlessSigningFlow,
     showPasskeyHelpers = true,
     callbacks,
     initializeWebView,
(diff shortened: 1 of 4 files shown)


packages/client/react-base/src/providers/CrossmintWalletUIBaseProvider.tsx
```diff
@@ -0,0 +1,183 @@
...
+    type?: ValidPasskeyPromptType;
+    primaryActionOnClick?: () => void;
+    secondaryActionOnClick?: () => void;
+};
+
+export function CrossmintWalletUIBaseProvider({
+    children,
+    appearance,
+    createOnLogin,
+    headlessSigningFlow = false,
```
jmderby: By defaulting to false we technically don't create breaking changes but rather developers using this will just get our built-in ui by default.

PR #1415, opened 2025-09-29
Fix: EVM typed data signature type

alberto-crossmint:
## Description

This was not handled as a smart wallet signature but as a EOA instead, which was not correct

## Test plan

Nothing to test

### DIFF
--- a/packages/wallets/src/wallets/evm.ts
+++ b/packages/wallets/src/wallets/evm.ts
@@ -110,7 +110,6 @@ export class EVMWallet extends Wallet<EVMChain> {
                 },
                 signer: this.signer.locator(),
                 chain,
-                isSmartWalletSignature: false,
             },
         });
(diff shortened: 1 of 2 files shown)


packages/wallets/src/wallets/evm.ts
```diff
@@ -110,7 +110,6 @@ export class EVMWallet extends Wallet<EVMChain> {
                 },
                 signer: this.signer.locator(),
                 chain,
-                isSmartWalletSignature: false,
```
AlbertoElias, line 113: we need to make sure that when adding a delegated signer, this is set to false though

alberto-crossmint: Hmm but are we ever using this function for creating a delegated signer?

alberto-crossmint: Discussed offline! Resolving

PR #1418, opened 2025-09-29
SDK: Optimize providers re-renders

alberto-crossmint:

### DIFF
--- a/packages/client/react-base/src/providers/CrossmintProvider.tsx
+++ b/packages/client/react-base/src/providers/CrossmintProvider.tsx
@@ -1,74 +1,110 @@
-import { type ReactNode, createContext, useCallback, useMemo, useRef, useState } from "react";
-import { type Crossmint, type CrossmintConfig, type CustomAuth, createCrossmint } from "@crossmint/common-sdk-base";
+import {
+  type ReactNode,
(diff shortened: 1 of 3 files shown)


packages/client/react-base/src/providers/CrossmintProvider.tsx
```diff
@@ -58,17 +65,19 @@ export function CrossmintProvider({
...
-    const value = useMemo(
-        () => ({
+    const value = useMemo(() => {
+        if (!crossmintRef.current) {
+            throw new Error("CrossmintProvider is not initialized");
+        }
+        return {
             get crossmint() {
-                return crossmintRef.current;
+                return crossmintRef.current!;
```
AlbertoElias, line 74: shouldnt be necessary if we do == null above

AlbertoElias (approved):
should use == null checks everywhere

packages/client/react-base/src/providers/CrossmintProvider.tsx
alberto-crossmint: Unfortunately the linter would complain as non-null assertions do not propagate inside functions bc this can not ensure that the execution context is kept unmodified

PR #1465, opened 2025-10-28
shadow-signers: Support for React Native

guilleasz-crossmint:
### Overview
This PR implements Shadow Signer support for React Native by creating a **platform-agnostic abstraction layer** for the existing browser shadow signer implementation and adding a React Native-specific storage adapter that uses a hidden WebView to access browser crypto APIs.

---

### Core Abstraction Layer (New)

The shadow signer logic was **extracted and abstracted** into platform-agnostic components:

#### **1. `ShadowSignerStorage` Interface** (`packages/wallets/src/signers/shadow-signer/index.ts`)
```typescript
export interface ShadowSignerStorage {
    keyGenerator(): Promise<string>;
    sign(publicKey: string, data: Uint8Array): Promise<Uint8Array>;
    storeMetadata(walletAddress: string, data: ShadowSignerData): Promise<void>;
    getMetadata(walletAddress: string): Promise<ShadowSignerData | null>;
}
```

**Purpose**: Defines the contract that any platform (browser, React Native, Node.js) must implement.

#### **2. Shadow Signer Core Functions**
- `generateShadowSigner()` - Generates Ed25519 key pair and returns encoded public key
- `storeShadowSigner()` - Persists shadow signer metadata (wallet address ↔ public key mapping)
- `getShadowSigner()` - Retrieves shadow signer metadata for a wallet
- `hasShadowSigner()` - Checks if a shadow signer exists

**Key Abstraction**: These functions accept an optional `storage` parameter, allowing different storage implementations per platform.

#### **3. `BrowserShadowSignerStorage` Implementation**
Moved browser-specific logic into its own class:
- **Private Keys**: Stored in IndexedDB as non-extractable `CryptoKey` objects
- **Metadata**: Stored in localStorage (public keys, wallet addresses, chain info)
- **Crypto Operations**: Direct access to `window.crypto.subtle` API

#### **4. Stellar Key Encoding** (`encodeEd25519PublicKey.ts`)
Implemented custom Stellar key encoding because the `stellar-sdk` library is incompatible with React Native:
- Base32 encoding with CRC16 checksum
- Version byte prefix for Ed25519 public keys
- Pure JavaScript implementation with no native dependencies

---

### The Challenge
React Native **doesn't have direct access to**:
- `window.crypto.subtle` API (Web Crypto API)
- `indexedDB` for storing non-extractable keys
- Standard browser storage APIs

### The Solution: WebView Bridge Architecture

React Native implementation uses a **hidden WebView** as a cryptographic sandbox to access browser APIs:

```
┌──────────────────────────────────────────────────────┐
│          React Native Environment                     │
│                                                        │
│  ┌─────────────────────────────────────────────┐    │
│  │  WebViewShadowSignerStorage                  │    │
│  │  (React Native side)                         │    │
│  │                                               │    │
│  │  • Initialize WebView                        │    │
│  │  • Send crypto operations via postMessage    │    │
│  │  • Receive results via message handler       │    │
│  │  • Store metadata in SecureStore            │    │
│  └──────────────┬──────────────┬─────────────────┘    │
│                 │              │                       │
│                 ↓              ↑                       │
│         postMessage      onMessage                    │
│                 │              │                       │
│  ┌──────────────┴──────────────┴─────────────────┐    │
│  │       Hidden WebView               │    │
│  │                                                │    │
│  │  ┌──────────────────────────────────────┐    │    │
│  │  │  Injected JavaScript Code            │    │    │
│  │  │  (webview-storage-injected.ts)       │    │    │
│  │  │                                       │    │    │
│  │  │  • Access to crypto.subtle API       │    │    │
│  │  │  • Access to IndexedDB               │    │    │
│  │  │  • Generate non-extractable keys     │    │    │
│  │  │  • Sign data with stored keys        │    │    │
│  │  └──────────────────────────────────────┘    │    │
│  └───────────────────────────────────────────────┘    │
└──────────────────────────────────────────────────────┘
```

## Test plan
Manualy tested the flow of creating wallet with shadow signers, and checking it can do transactions without OTP verification


packages/wallets/src/signers/shadow-signer/encodeEd25519PublicKey.ts
```diff
@@ -0,0 +1,64 @@
+/**
+ * This code was copied from the stellar-sdk-base library.
+ * https://github.com/stellar/js-stellar-base/blob/5bac0b60bcd01794bb0b11fd8b6a45e486c28626/src/strkey.js#L400
+ * This is because the stellar-sdk library is not compatible with React Native.
+ */
+
+import base32 from "base32.js";
+
+function calculateChecksum(payload: Uint8Array): Uint8Array {
```
AlbertoElias: i would double check the implementation from `https://github.com/Crossmint/open-signer/blob/main/shared/cryptography/src/primitives/encoding/strkey.ts` to make sure it's aligned

packages/wallets/src/signers/shadow-signer/encodeEd25519PublicKey.ts
AlbertoElias: should we move this to utils?

packages/client/ui/react-native/src/utils/WebViewShadowSignerStorage.ts
```diff
@@ -0,0 +1,187 @@
...
+    async getMetadata(walletAddress: string): Promise<ShadowSignerData | null> {
+        try {
+            const key = `${this.SHADOW_SIGNER_STORAGE_KEY}_meta_${walletAddress}`;
+            console.log("[WebViewShadowSignerStorage] Getting metadata for key:", key);
+
+            const stored = await SecureStore.getItemAsync(key);
+
+            console.log("[WebViewShadowSignerStorage] Retrieved raw metadata:", stored);
+
+            if (!stored) {
```
AlbertoElias: == null

packages/client/ui/react-native/src/utils/webview-storage-injected.ts
AlbertoElias: wondering if we could re-use the code in browser storage and load it as a string to avoid duplication and changing it in two places

packages/client/ui/react-native/src/utils/WebViewShadowSignerStorage.ts
```diff
@@ -0,0 +1,187 @@
...
+            // Ignore parse errors - might be other messages
+        }
+    };
+
+    private async callWebViewFunction(
+        operation: string,
+        params: Record<string, unknown>
+    ): Promise<Record<string, unknown>> {
+        const webView = this.webViewRef?.current;
+        if (!webView) {
```
AlbertoElias: == null

AlbertoElias (commented):
overall looking good!! lefts some nits and questions

packages/wallets/src/signers/non-custodial/ncs-signer.ts
```diff
@@ -23,8 +25,13 @@ export abstract class NonCustodialSigner implements Signer {
     } | null = null;
     private _initializationPromise: Promise<void> | null = null;
     protected shadowSigner: ExternalWalletSigner<Chain> | null = null;
+    protected shadowSignerStorage?: ShadowSignerStorage;
```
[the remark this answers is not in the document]
guilleasz-crossmint: The shadowSigner is an ExternalWalletSigner not the config of the shadowSigner itself, the config is autogenerated in `getShadowSignerConfig` which is specific for each chain. We need access to the shadowSignerStorage before generating the shadowSignerConfig  as you can see in `initializeShadowSigner` in `packages/wallets/src/signers/non-custodial/ncs-signer.ts`

packages/wallets/src/wallets/wallet-factory.ts
```diff
@@ -129,9 +126,14 @@ export class WalletFactory {
...
         }
-
-        if (shadowSignerPublicKey != null && shadowSignerPrivateKey != null) {
-            await storeShadowSigner(walletResponse.address, args.chain, shadowSignerPublicKey, shadowSignerPrivateKey);
+        if (shadowSignerPublicKey != null && shadowSignerPublicKeyBase64 != null) {
+            await storeShadowSigner(
+                walletResponse.address,
+                args.chain,
+                shadowSignerPublicKey,
+                shadowSignerPublicKeyBase64,
```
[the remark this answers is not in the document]
guilleasz-crossmint: Because as we cannot retrieve the privatekey from the webview for storing it after wallet creation, we need to save the private key with the only identifier we have at that point that is the publickey in base 64, this public key later will be encoded according to each chain, so we need to save a reference of the base64 one. There were other solutions I though:
1. Do the encoded per chain in the storage, so we can already save it with the final public key, but this will require moving the encoding libraries into the webview
2. After returning the publickKeyBase64, encode it correctly and re-index the private key to the final publickey, but it seems more complicated to create all the re-index logic than to just save the publickeybase64

packages/wallets/src/utils/encodeEd25519PublicKey.ts
guilleasz-crossmint: Moved to utils

packages/wallets/src/signers/shadow-signer/index.ts
```diff
@@ -1,57 +1,54 @@
...
-        request.onerror = () => reject(request.error);
-        request.onsuccess = () => resolve(request.result);
-        request.onupgradeneeded = (event) => {
-            const db = (event.target as IDBOpenDBRequest).result;
-            if (!db.objectStoreNames.contains(SHADOW_SIGNER_DB_STORE)) {
-                db.createObjectStore(SHADOW_SIGNER_DB_STORE);
-            }
-        };
-    });
+export interface ShadowSignerStorage {
```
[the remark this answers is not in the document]
AlbertoElias: i just thought i had seen the interface definition in the react libraries too but couldnt find it now

packages/client/ui/react-native/src/utils/WebViewShadowSignerStorage.ts
```diff
@@ -0,0 +1,187 @@
...
+        this.injectStorageHandler();
+    }
+
+    private injectStorageHandler(): void {
+        if (this.isInjected || !this.webViewRef?.current) {
+            return;
+        }
+
+        try {
+            this.webViewRef.current.injectJavaScript(SHADOW_SIGNER_STORAGE_INJECTED_JS);
```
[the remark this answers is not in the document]
AlbertoElias: actually, this is a security issue with our RNWebView, we didn't know it had a method to injectJavaScript and we would prefer that it wasn't the case. We might have to see how to disable that. I would suggest not using RNWebview as this is only when needing to use the window-sdk to send messages with handshake etc

AlbertoElias (approved):
left some extra comments but this is looking really good! leaving it approved

packages/client/ui/react-native/src/utils/WebViewShadowSignerStorage.ts
kaspernj: Addressed in 3b6658132ac51b43c8913f81d37092b642dfa7ca. `src/backends/identifiers.js` now carries a deterministic PHP core/date/SPL/JSON/Reflection/PDO class and interface baseline instead of the previous tiny subset, so collisions such as `DateTime` fail before emission. The backend validation spec now covers `DateTime` across all targets. Validation: all directly affected Task 009 specs passed individually (117 tests total), followed by lint, typecheck, build, audit, both dependency-tree gates, package dry-run/content checks, public API, packed cold-consumer behavior, and diff/clean-tree checks.

PR #1763, opened 2026-04-02
fix: skip getSigner API call for admin/recovery signers in useSigner

albertoelias-crossmint:
## Description

When `useSigner()` is called with the recovery signer (admin signer in our API), `assembleFullSigner()` calls `getSigner()` to fetch the signer's status. However, the `getSigner` endpoint only works for **delegated** signers — it returns 404/400 for admin signers. Since admin signers are always approved for their wallet, this API call is unnecessary and causes errors.

This PR:
1. **Short-circuits `resolveNonDeviceSigner`** for non-passkey recovery signers — skips the `signerIsRegistered()` → `signers()` → `getSignerState()` chain entirely.
2. **Skips the `getSigner` call in `assembleFullSigner`** when the signer is an admin signer, setting status to `"active"` directly.
3. Applies the same optimization in `initDefaultSigner` when auto-assembling the recovery signer.
4. **Fixes a pre-existing bug in `isRecoverySigner`** for server-type recovery signers: when the recovery config is API-sourced (`{type: "server", address: "0x..."}` without `secret`), calling `deriveServerSignerDetails` would crash. Now uses the `address` directly when available, matching the intent described in the existing code comment.

Passkey signers are excluded from the early return because `isRecoverySigner` matches passkeys by type only, which could incorrectly match a delegated passkey. They still get the optimization when they fall through to the recovery fallback paths.

**Reviewer checklist:**
- [ ] `"active"` is the correct status to hardcode — `isApprovedSignerStatus()` accepts both `"success"` and `"active"`.
- [ ] Passing `undefined` as the second arg to `assembleFullSigner` still triggers the default (`this.#options?.deviceSignerKeyStorage`), which is correct per JS semantics.
- [ ] The passkey exclusion (`signer.type !== "passkey"`) in the early return of `resolveNonDeviceSigner` is necessary and correct — passkeys that are actually delegated signers must not be short-circuited.
- [ ] The `"address" in recovery` check in `isRecoverySigner` correctly handles API-sourced server recovery configs that lack a `secret`. Note: the `ServerSignerConfig` type doesn't declare `address`, so this is a runtime guard for the API-returned shape — verify this is acceptable vs. extending the type.

## Test plan

- Updated all 6 `useSigner` recovery-signer tests to assert `getSigner` is **not** called and signer status is `"active"`:
  - 4 non-passkey types: `api-key`, `email`, `phone`, `external-wallet`
  - 2 passkey recovery paths: passkey without id (auto-select fallback) and passkey with explicit id (registration-check fallback)
- All 264 existing unit tests pass (`pnpm test:vitest`).
- Lint passes (`pnpm lint`).
- Libs build succeeds (`pnpm build:libs`).
- ⚠️ No dedicated unit test for the server-signer `isRecoverySigner` fix (API-sourced recovery config without `secret`). The scenario is hard to reproduce in the existing test harness since the mock wallets use simple `recovery` shapes. Manual verification or an integration test would strengthen confidence here.

## Package updates

- `@crossmint/wallets-sdk`: patch — changeset added via `.changeset/skip-getsigner-admin-signers.md`

Link to Devin session: https://crossmint.devinenterprise.com/sessions/2b3dd5cfee21410387ff48862a94d65a

---

<a href="https://crossmint.devinenterprise.com/review/crossmint/crossmint-sdk/pull/1763" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


packages/wallets/src/wallets/wallet.ts
```diff
@@ -1254,12 +1267,15 @@ export class Wallet<C extends Chain> {
...
                 this.#apiClient.environment
             ).derivedAddress;
-            const recoveryDerived = deriveServerSignerDetails(
-                recovery,
-                this.chain,
-                this.#apiClient.projectId,
-                this.#apiClient.environment
-            ).derivedAddress;
+            const recoveryDerived =
+                "address" in recovery && recovery.address != null
```
albertoelias-crossmint: @devinai why this check?

panosinthezone (approved):
approving but definitely need to clean this codebase up

PR #1766, opened 2026-04-06
fix: handle API-sourced recovery configs without secret in isRecoverySigner

panosinthezone:
## Description

When a wallet is fetched via `getWallet()` without passing a `recovery` config, the SDK falls back to the API-sourced `adminSigner` config, which has the shape `{type: "server", address: "G...", locator: "server:G..."}` — no `secret` property. Calling `wallet.useSigner({ type: "server", secret: walletSecret })` on such a wallet crashes with:

```
TypeError: Cannot read properties of undefined (reading 'startsWith')
```

The crash happens in `isRecoverySigner` because `deriveServerSignerDetails()` is called on the recovery config, which flows into `deriveKeyBytes(signer.secret, ...)` → `stripAndValidateSecret(undefined)` → `undefined.startsWith(...)`.

**Fix:** Model the API-sourced shape in the type system and guard all paths that assume a `secret` is present.

### Changes

1. **New types in `signers/types.ts`** — `ApiSourcedServerSignerConfig` (the shape the API actually returns: `{ type: "server", address: string }`) and `RecoverySignerConfigForChain<C>` (union of `SignerConfigForChain<C> | ApiSourcedServerSignerConfig`). A runtime type guard `isApiSourcedServerSignerConfig` discriminates the two at runtime.

2. **`wallet-factory.ts`** — The `adminSigner` cast now uses `RecoverySignerConfigForChain<C>` instead of lying with `SignerConfigForChain<C>`.

3. **`isRecoverySigner`** — Uses a `resolveAddress` helper that checks `isApiSourcedServerSignerConfig` to decide whether to use `.address` directly or derive via `deriveServerSignerDetails`. On a successful match, **upgrades `#recovery`** with the user-provided signer config (which includes the `secret`), so downstream code has the complete config.

4. **`withRecoverySigner` and `resumePendingDeviceSignerApproval`** — Both now guard against API-sourced recovery configs with an explicit error directing the user to call `useSigner(...)` first, replacing the opaque `TypeError`.

5. **`isAutoAssemblableSignerConfig`** — Uses `isApiSourcedServerSignerConfig` to correctly report that API-sourced server configs cannot be auto-assembled.

### ⚠️ Things for reviewer to verify
- `isRecoverySigner` now has a **side effect**: it mutates `this.#recovery` on a successful match. This is intentional — it ensures the recovery config is upgraded with runtime-only fields (e.g. `secret`, `onSign`) — but it makes the method more than a pure predicate.
- The public `recovery` getter (line 237) casts `this.#recovery as SignerConfigForChain<C>`. This is a lossy cast if the recovery is still API-sourced. Verify no external caller depends on this getter returning a config with `secret`.
- Several `this.#recovery as SignerConfigForChain<C>` casts appear after `isApiSourcedServerSignerConfig` guards. These are safe (the guard throws/returns before the cast), but worth confirming the control flow.

## Test plan

* Six new unit tests in `wallet.test.ts` under the "recovery signer support" describe block:
  1. **API-sourced recovery (no secret) — `useSigner` succeeds:** Verifies no TypeError when recovery config lacks `secret` and derived address matches.
  2. **Address mismatch — `useSigner` rejects:** Derived address doesn't match API-sourced recovery → throws "is not registered" (not a TypeError).
  3. **User-provided recovery (has secret) — `useSigner` succeeds:** Existing derive-both-sides path still works.
  4. **`addSigner` happy path — API-sourced recovery:** `useSigner` with matching server signer, then `addSigner` succeeds (exercises `withRecoverySigner` after recovery upgrade).
  5. **`addSigner` rejects — no signer set:** API-sourced recovery, no `useSigner` call → throws "no secret available".
  6. **`addSigner` rejects — wrong signer type:** Email signer via `useSigner`, then `addSigner` → throws "no secret available" (email didn't upgrade `#recovery`).
* All existing tests continue to pass (`pnpm test:vitest` — 293 passed).
* Lint passes (`pnpm lint`).

## Package updates

Changeset added for `@crossmint/wallets-sdk` (patch).

Link to Devin session: https://crossmint.devinenterprise.com/sessions/e2e2dcf09d6e46f785d9dd65cc6e07b1
Requested by: @panosinthezone

---

<a href="https://crossmint.devinenterprise.com/review/crossmint/crossmint-sdk/pull/1766" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


packages/wallets/src/wallets/wallet.ts
```diff
@@ -915,7 +917,13 @@ export class Wallet<C extends Chain> {
 
     private async withRecoverySigner<T>(operation: () => Promise<T>): Promise<T> {
         const originalSigner = this.signer;
-        const recoveryInternalConfig = this.buildInternalSignerConfig(this.#recovery);
+        if (isApiSourcedServerSignerConfig(this.#recovery)) {
```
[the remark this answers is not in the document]
panosinthezone: We handle by throwing an internal error. Now for that method call we throw a better user facing error

packages/wallets/src/wallets/wallet.ts
```diff
@@ -971,7 +972,23 @@ export class Wallet<C extends Chain> {
...
+            throw new Error(
+                "Cannot assemble server signer: no secret available. " +
+                    'Call wallet.useSigner({ type: "server", secret: ... }) first with the recovery server secret.'
+            );
+        }
+        if (
+            this.#recovery != null &&
+            this.#recovery.type === "external-wallet" &&
```
albertoelias-crossmint, line 983: should we follow a similar approach and have an `isApiSourcedExternalWalletSigner`? we can also remove config for the name, doesn't add much value

albertoelias-crossmint (approved):

packages/wallets/src/wallets/wallet.ts
panosinthezone: > isApiSourcedExternalWalletSigner

yeah

PR #1775, opened 2026-04-09
Fix: ensure signer is authenticated

maxsch-xmint:
## Description

Porting a fix from 0.13.28. Context here: https://crossmint.slack.com/archives/C05KRUMULD6/p1775760283519089

## Test plan

Released and asked client to update

### DIFF
--- a/packages/wallets/src/wallets/wallet.ts
+++ b/packages/wallets/src/wallets/wallet.ts
@@ -582,11 +582,12 @@ export class Wallet<C extends Chain> {
         },
     })
     public async approve<T extends ApproveParams>(params: T): Promise<ApproveResult<T>> {
-        this.requireSigner();
         walletsLogger.info("wallet.approve.start", {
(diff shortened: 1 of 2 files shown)


packages/wallets/src/wallets/wallet.ts
```diff
@@ -582,12 +582,13 @@ export class Wallet<C extends Chain> {
         },
     })
     public async approve<T extends ApproveParams>(params: T): Promise<ApproveResult<T>> {
-        this.requireSigner();
         walletsLogger.info("wallet.approve.start", {
             transactionId: params.transactionId,
             signatureId: params.signatureId,
         });
 
+        await this.preAuthIfNeeded();
```
maxsch-xmint: this calles this.requireSigner(); internally, that's why I deleted it some lines above

albertoelias-crossmint (approved):

PR #1782, opened 2026-04-10
fix: make addSigner idempotent by resuming pending operations on retry

albertoelias-crossmint:
## Description

Fixes [WAL-9383](https://linear.app/crossmint/issue/WAL-9383/addsigner-retry-shouldnt-fail).

When `addSigner` is called and the approval step (step 2) fails mid-flight (network error, timeout, etc.), the signer is left in a pending state on the backend. On retry, `addSigner` would blindly call `registerSigner` again, which fails because the signer already exists.

This PR makes `addSigner` idempotent by checking for existing signer state before registration:
- **Already approved** → return immediately (no-op)
- **Pending operation** → resume the pending approval (transaction or signature) instead of re-registering
- **Failed / not found** → proceed with fresh registration as before

The approval logic (handling transactions for Solana/Stellar and signatures for EVM) is extracted into a shared `completeSignerRegistration` helper, used by both the fresh registration path and the resume path. This reuses the existing `getSignerState` / `getPendingSignerOperation` infrastructure that `recover()` already relies on.

### Things for reviewer to check
- Every `addSigner` call now makes an extra `getSigner` API call upfront to check existing state — confirm this latency tradeoff is acceptable
- Verify `isApprovedSignerStatus` (checks for `"success"` / `"active"`) covers all terminal approved states
- Race condition edge case: if a signer transitions from pending → approved between the upfront check and the approval attempt, the approval call should still be safe (idempotent on the backend), but worth confirming

## Test plan

- 6 new unit tests covering addSigner retry/idempotency:
  - Resume pending EVM signature instead of re-registering
  - Resume pending Solana transaction instead of re-registering
  - Early return when signer is already approved
  - True idempotency: calling `addSigner` twice with the same signer yields the same result (only one `registerSigner` call)
  - `prepareOnly` mode returns operation ID when resuming a pending operation
  - Fall through to fresh registration when signer is in failed state
- 3 existing `recover()` tests updated to account for the additional `getSigner` call from `addSigner`
- All 293 tests pass locally, lint passes

## Package updates

- `@crossmint/wallets-sdk`: patch bump (changeset added)

Link to Devin session: https://crossmint.devinenterprise.com/sessions/408fbed01afa4dae9266b753cae9bc78

---

<a href="https://crossmint.devinenterprise.com/review/crossmint/crossmint-sdk/pull/1782" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/packages/wallets/src/wallets/wallet.ts
+++ b/packages/wallets/src/wallets/wallet.ts
@@ -752,23 +752,39 @@ export class Wallet<C extends Chain> {
             if (pendingOperation == null) {
                 return { ...signer } as AddSignerReturnType<C>;
             }
-            const operationId =
-                pendingOperation.type === "transaction"
-                    ? { transactionId: pendingOperation.id }
-                    : { signatureId: pendingOperation.id };
-            walletsLogger.info("wallet.addSigner.prepared", operationId);
-            return { ...signer, ...operationId } as AddSignerReturnType<C>;
+            switch (pendingOperation.type) {
+                case "transaction": {
+                    const operationId = { transactionId: pendingOperation.id };
+                    walletsLogger.info("wallet.addSigner.prepared", operationId);
+                    return { ...signer, ...operationId } as AddSignerReturnType<C>;
+                }
+                case "signature": {
+                    const operationId = { signatureId: pendingOperation.id };
+                    walletsLogger.info("wallet.addSigner.prepared", operationId);
+                    return { ...signer, ...operationId } as AddSignerReturnType<C>;
+                }
+                default: {
+                    const _exhaustive: never = pendingOperation.type;
+                    throw new Error(`Unknown pending operation type: ${_exhaustive}`);
+                }
+            }
         }
 
         if (pendingOperation != null) {
-            if (pendingOperation.type === "transaction") {
-                await this.approveTransactionAndWait(pendingOperation.id);
-            } else {
-                await this.approveSignatureAndWait(pendingOperation.id);
+            switch (pendingOperation.type) {
+                case "transaction":
+                    await this.approveTransactionAndWait(pendingOperation.id);
+                    walletsLogger.info("wallet.addSigner.success", { transactionId: pendingOperation.id });
+                    break;
+                case "signature":
+                    await this.approveSignatureAndWait(pendingOperation.id);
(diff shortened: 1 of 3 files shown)


packages/wallets/src/wallets/wallet.ts
```diff
@@ -680,70 +707,73 @@ export class Wallet<C extends Chain> {
...
+        options?: AddSignerOptions
+    ): Promise<AddSignerReturnType<C> | WalletSigner> {
+        if (options?.prepareOnly) {
+            if (pendingOperation == null) {
+                return { ...signer } as AddSignerReturnType<C>;
+            }
+            const operationId =
+                pendingOperation.type === "transaction"
+                    ? { transactionId: pendingOperation.id }
+                    : { signatureId: pendingOperation.id };
```
panosinthezone: Lets not default here and process explicitly for the type

packages/wallets/src/wallets/wallet.ts
```diff
@@ -680,70 +707,73 @@ export class Wallet<C extends Chain> {
...
+            return { ...signer, ...operationId } as AddSignerReturnType<C>;
+        }
 
-                return { ...registeredSigner, status: "success" as const } as WalletSigner;
+        if (pendingOperation != null) {
+            if (pendingOperation.type === "transaction") {
+                await this.approveTransactionAndWait(pendingOperation.id);
+            } else {
+                await this.approveSignatureAndWait(pendingOperation.id);
             }
```
panosinthezone: same here

panosinthezone (approved):

packages/wallets/src/wallets/wallet.ts
albertoelias-crossmint: @devinai lets take a look. if you were to change it, what would be the cleanest and simplest approach?

packages/wallets/src/wallets/wallet.ts
albertoelias-crossmint: @devinai lets take a look. if you were to change it, what would be the cleanest and simplest approach?

packages/wallets/src/wallets/wallet.ts
albertoelias-crossmint: @devinai do it

PR #1788, opened 2026-04-13
Fix/e2e stellar prepared transaction

luans-qa:
## Description

 - Stellar prepared transaction always failing: approval-test.tsx used wallet?.chain !== "solana" to detect EVM      
  wallets, which incorrectly treated Stellar as EVM. This caused viem's isAddress() to reject all Stellar addresses   
  (G.../C...) with "Invalid recipient address" — the prepared transaction was never created and the test timed out    
  waiting for an element that never appeared. Fixed chain detection, added Stellar regex validation, and added a      
  stable data-testid="prepare-recipient" to the recipient input.                                                      
  - Phone OTP transfer test failing on retries: The sendCodePromise URL pattern (/signers/, /approvals,
  /transactions/, /confirm) never matches the phone OTP send endpoint, so it always burned 30s waiting for a response
  that never comes. On later retries with slow SMS delivery, the remaining budget wasn't enough. Reduced that wasted
  wait to 10s, extended the "Check your phone" message wait to 90s, and raised the global test timeout to 3 minutes.
  - webkit + phone authenticate tests failing on all retries: The toContainText("Wallets Quickstart") assertion uses
  Playwright's default 5s expect timeout. webkit on CI renders more slowly for phone signer configs, consistently
  timing out. Raised the global expect.timeout to 15s.

## Test plan

Tests should pass in CI - Example execution: https://github.com/Crossmint/crossmint-sdk/actions/runs/24347855334/job/71093624854

### DIFF
--- a/apps/wallets/quickstart-devkit/components/approval-test.tsx
+++ b/apps/wallets/quickstart-devkit/components/approval-test.tsx
@@ -8,12 +8,13 @@ import { isAddress } from "viem";
 export function ApprovalTest() {
     const { wallet } = useWallet();
 
-    const isEVMWallet = wallet?.chain !== "solana";
+    const isStellarWallet = wallet?.chain === "stellar";
     const isSolanaWallet = wallet?.chain === "solana";
+    const isEVMWallet = !isStellarWallet && !isSolanaWallet;
 
     // State for creating transactions that need approval
     const [prepareTransfer, setPrepareTransfer] = useState({
-        token: (isEVMWallet ? "eth" : "sol") as "eth" | "usdc" | "sol" | "usdxm",
+        token: (isEVMWallet ? "eth" : isSolanaWallet ? "sol" : "usdxm") as "eth" | "usdc" | "sol" | "usdxm",
         recipient: "",
         amount: "",
     });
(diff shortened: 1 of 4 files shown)


apps/wallets/quickstart-devkit/components/approval-test.tsx
```diff
@@ -8,12 +8,13 @@ import { isAddress } from "viem";
...
 
-    const isEVMWallet = wallet?.chain !== "solana";
+    const isStellarWallet = wallet?.chain === "stellar";
     const isSolanaWallet = wallet?.chain === "solana";
+    const isEVMWallet = !isStellarWallet && !isSolanaWallet;
 
     // State for creating transactions that need approval
     const [prepareTransfer, setPrepareTransfer] = useState({
-        token: (isEVMWallet ? "eth" : "sol") as "eth" | "usdc" | "sol" | "usdxm",
+        token: (isEVMWallet ? "eth" : isSolanaWallet ? "sol" : "usdxm") as "eth" | "usdc" | "sol" | "usdxm",
```
albertoelias-crossmint: why dont we use usdxm for all chains?

albertoelias-crossmint (approved):

apps/wallets/quickstart-devkit/components/approval-test.tsx
luans-qa: That makes sense, I pushed a change that use usdxm as default

PR #1808, opened 2026-04-22
feat: add key export event callback to SDK

panosinthezone:
## Description

Adds an `onExport` callback prop to `ExportPrivateKeyButton` (React **and** React Native) so host applications can react when a user successfully exports (copies) their private key. The iframe sends an `event:key-exported` PostMessage after clipboard copy; the SDK listens for it and invokes the callback.

This is the SDK-side counterpart to the event reporting added in [open-signer#199](https://github.com/Crossmint/open-signer/pull/199).

### Changes

1. **`KeyExportedPayloadSchema`** (`packages/client/signers/…/schemas.ts`) — new schema using `ResultResponse(z.object({}))`, consistent with all other event schemas. Imported and referenced in `exportSignerOutboundEvents` in `events.ts`.
2. **`onExport` prop** (`ExportPrivateKeyButton.tsx`, both React and React Native) — optional `() => void | Promise<void>` callback. Uses a `useRef` pattern to keep the callback current without causing iframe/WebView reloads when the parent re-renders with a new function reference.
3. **Listener wiring** (`ncs-signer.ts`) — registers a **fire-once** `on("event:key-exported")` listener on the TEE connection before sending the export request. The listener deregisters itself via `off(listenerId)` after the first invocation, then guards on `data.status === "success"` before invoking the callback. Error handling uses `Promise.resolve().then().catch()` to handle both sync throws and async rejections without blocking the export flow.
4. **Interface update** (`types.ts`) — `ExportableSignerAdapter._exportPrivateKey` signature updated to accept the optional callback.
5. **Conditional registration** — UI components only pass the callback wrapper when `onExportRef.current != null`, so the `if (onExport != null)` guard in `ncs-signer` correctly skips listener registration when no callback is provided.

### ⚠️ Review focus / human checklist

- [ ] **Schema validation**: The iframe sends `{ status: "success" }`. Verify `ResultResponse(z.object({}))` accepts this payload (it should — the success branch is `z.object({ status: z.literal("success") }).and(z.object({}))`).
- [ ] **Status guard + fire-once interaction**: The listener deregisters after the *first* `event:key-exported` regardless of status. If the first event were an error, `onExport` would never fire. In practice, the iframe only sends the event on successful clipboard copy, so this should be fine — but worth confirming.
- [ ] **Event timing**: `event:key-exported` fires from the iframe *after* `response:export-signer` (which resolves `sendAction`). The listener must still be reachable when the event arrives.
- [ ] **Late `onExport` prop**: The `onExportRef.current != null` check runs when `handleIframeLoad`/`handleWebViewLoadEnd` fires. If `onExport` is `undefined` at iframe load time but provided later, the listener won't be registered.
- [ ] **Error objects not logged**: `catch` blocks use static strings only (no error object) per Alberto's review feedback about avoiding sensitive info leaks. Trade-off is reduced debuggability.

## Test plan

Manually verified lint passes (`pnpm lint`). CI build & test and lint checks pass. No new unit tests — the callback is a fire-and-forget hook that depends on the iframe PostMessage flow, which cannot be exercised in isolation without mocking the transport layer. Integration testing recommended once both PRs (this + open-signer#199) are merged and deployed together.

## Package updates

- `@crossmint/client-sdk-react-ui` — minor
- `@crossmint/client-sdk-react-native-ui` — minor
- `@crossmint/wallets-sdk` — minor
- `@crossmint/client-signers` — minor

Changeset added via `.changeset/key-export-event.md`.

Link to Devin session: https://crossmint.devinenterprise.com/sessions/d079cc2486724c0d9c11a45c7d953750
Requested by: @panosinthezone

### DIFF
--- a/packages/client/signers/src/communications/events.ts
+++ b/packages/client/signers/src/communications/events.ts
@@ -1,4 +1,4 @@
-import type { z } from "zod";
+import { z } from "zod";
 
 import {
     GetStatusPayloadSchema,
@@ -38,6 +38,7 @@ export const exportSignerInboundEvents = {
 
 export const exportSignerOutboundEvents = {
     "response:export-signer": ExportSignerPayloadSchema.response,
+    "event:key-exported": z.object({}),
 } as const;
 
 export type ExportSignerInputEvent<E extends ExportSignerEventName> = z.infer<
--- a/packages/client/signers/src/communications/events.ts
+++ b/packages/client/signers/src/communications/events.ts
@@ -1,4 +1,4 @@
-import { z } from "zod";
+import type { z } from "zod";
 
 import {
     GetStatusPayloadSchema,
@@ -6,6 +6,7 @@ import {
     StartOnboardingPayloadSchema,
     CompleteOnboardingPayloadSchema,
     ExportSignerPayloadSchema,
+    KeyExportedPayloadSchema,
 } from "./schemas";
 
 export const SIGNER_EVENTS = ["start-onboarding", "complete-onboarding", "sign", "get-status"] as const;
@@ -38,7 +39,7 @@ export const exportSignerInboundEvents = {
 
 export const exportSignerOutboundEvents = {
     "response:export-signer": ExportSignerPayloadSchema.response,
-    "event:key-exported": z.object({}),
+    "event:key-exported": KeyExportedPayloadSchema.event,
 } as const;
 
 export type ExportSignerInputEvent<E extends ExportSignerEventName> = z.infer<
--- a/packages/client/ui/react-native/src/components/wallets/ExportPrivateKeyButton.tsx
+++ b/packages/client/ui/react-native/src/components/wallets/ExportPrivateKeyButton.tsx
@@ -27,7 +27,12 @@ export function ExportPrivateKeyButton({ appearance, onExport }: ExportPrivateKe
         typeof exportSignerOutboundEvents,
         typeof exportSignerInboundEvents
     > | null>(null);
+    const onExportRef = useRef(onExport);
     const [frameUrl, setFrameUrl] = useState<string>("");
+
+    useEffect(() => {
+        onExportRef.current = onExport;
+    }, [onExport]);
 
     useEffect(() => {
         if (crossmint != null) {
@@ -57,14 +62,14 @@ export function ExportPrivateKeyButton({ appearance, onExport }: ExportPrivateKe
                     });
                     connectionRef.current = connection;
                     await connection.handshakeWithChild();
-                    await wallet.signer._exportPrivateKey(connection, onExport);
+                    await wallet.signer._exportPrivateKey(connection, () => onExportRef.current?.());
                 }
             } catch (error) {
                 console.error("Failed to export private key:", error);
                 Alert.alert("Export Failed", "Failed to export private key. Please try again.");
             }
         },
-        [wallet, onExport]
+        [wallet]
     );
 
     const handleWebViewError = useCallback((syntheticEvent: { nativeEvent: unknown }) => {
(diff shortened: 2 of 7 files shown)


packages/wallets/src/signers/non-custodial/ncs-signer.ts
```diff
@@ -334,12 +334,26 @@ export abstract class NonCustodialSigner implements SignerAdapter {
...
 
+        if (onExport != null) {
+            const listenerId = exportTEEConnection.on("event:key-exported", () => {
+                exportTEEConnection.off(listenerId);
+                Promise.resolve()
+                    .then(() => onExport())
+                    .catch((err) => {
+                        console.error("[NCS Signer] onExport callback error:", err);
+                    });
+            });
```
panosinthezone: This effectively registers a  one-off listener. We call the onExport only once here so we don't spam if the user tries to export the key multiple times.

packages/client/ui/react-native/src/components/wallets/ExportPrivateKeyButton.tsx
```diff
@@ -55,7 +62,10 @@ export function ExportPrivateKeyButton({ appearance }: ExportPrivateKeyButtonPro
                     });
                     connectionRef.current = connection;
                     await connection.handshakeWithChild();
-                    await wallet.signer._exportPrivateKey(connection);
+                    await wallet.signer._exportPrivateKey(
+                        connection,
+                        onExportRef.current != null ? () => onExportRef.current?.() : undefined
```
panosinthezone, line 67: This always uses the latest version of the onExport method.

packages/client/signers/src/communications/events.ts
```diff
@@ -38,6 +38,7 @@ export const exportSignerInboundEvents = {
 
 export const exportSignerOutboundEvents = {
     "response:export-signer": ExportSignerPayloadSchema.response,
+    "event:key-exported": z.object({}),
```
alberto-crossmint: Nit: Shouldn't we return a Result type as for the rest of the events? 

```typescript
const ResultResponse = <T extends z.ZodTypeAny>(schema: T) =>
    ErrorResponse.or(z.object({ status: z.literal("success") }).and(schema));
```

With an `{}` response type we won't be able to signal the upstream app if something wen't wrong and it won't know if it needs to retry or it was just that the exporter was slow

alberto-crossmint: Nit: Should we move this to the schemas folder? It's the only one defined inline

packages/wallets/src/signers/non-custodial/ncs-signer.ts
```diff
@@ -334,12 +334,26 @@ export abstract class NonCustodialSigner implements SignerAdapter {
...
         const { scheme, encoding } = this.getChainKeyParams();
 
+        if (onExport != null) {
+            const listenerId = exportTEEConnection.on("event:key-exported", () => {
+                exportTEEConnection.off(listenerId);
+                Promise.resolve()
+                    .then(() => onExport())
+                    .catch((err) => {
+                        console.error("[NCS Signer] onExport callback error:", err);
+                    });
```
alberto-crossmint, line 354: Don't we need to await this ?

alberto-crossmint: Also I don't fully understand this. From my understanding

Promise.resolve() resolves immediately. On resolution it calls `(()=> onExport())` that trivially results in `onExport()`, the `.catch` would never trigger because we're not awaiting `onExport` in the `.then` callback, and the result of this whole line would be effectively the same as doing `onExport()` isn't it? and we would never be awaiting that so it would not print any log on failure

packages/client/signers/src/communications/events.ts
panosinthezone: Okay, didn't know that existed. Taking a look into it

panosinthezone: Yep!

packages/wallets/src/signers/non-custodial/ncs-signer.ts
panosinthezone: @devinai can you look into this?

packages/client/signers/src/communications/events.ts
panosinthezone: @devinai can you update using the suggestion

panosinthezone: @devinai can you also apply this

alberto-crossmint (approved):

PR #1819, opened 2026-04-30
fix(wallets): support locator-based token matching in balances filter

mPaella:
## Description

`wallet.balances(tokens)` filters returned tokens via `requestedTokens.includes(token.symbol)`, which only works when requested tokens are symbols (e.g. `"aaplx"`). When tokens are requested by chain locator (e.g. `solana:XsbEh...`), the API returns the token correctly but the SDK drops it because the response symbol (`AAPLx`) does not match the locator string.

This adds a private `matchesRequestedToken` method that performs case-insensitive matching against:
- Token symbol
- Solana `mintHash` (bare or `solana:<mint>` format)
- EVM `contractAddress` (bare or `<chain>:<address>` format)
- Stellar `contractId` (bare or `stellar:<contractId>` format)

The three chain branches are mutually exclusive and short-circuit, keeping the method flat and readable. `null` vs empty-array semantics are preserved: `undefined` means no filter (return all), `[]` means return none.

## Test plan

- Unit tests added for all new locator-matching paths:
  - EVM `contractAddress` via `base-sepolia:<address>` locator
  - EVM `contractAddress` via bare address
  - Solana `mintHash` via `solana:<mint>` locator
  - Stellar `contractId` via `stellar:<contractId>` locator
  - Case-insensitive matching across locator formats
- Existing symbol-based filtering test (`wallet.balances(["dai"])`) still passes
- All 103 tests in `wallet.test.ts` pass
- Lint passes (`pnpm lint`)

## Package updates

- `@crossmint/wallets-sdk`: patch — changeset added via `.changeset/locator-balance-fix.md`

Link to Devin session: https://crossmint.devinenterprise.com/sessions/b31034c19c8d4387adefb47237ba2953
Requested by: @mPaella


packages/wallets/src/wallets/wallet.ts
```diff
@@ -411,6 +413,53 @@ export class Wallet<C extends Chain> {
...
+        }
+
+        if (
+            this.chain !== "solana" &&
+            this.chain !== "stellar" &&
+            chainData != null &&
+            "contractAddress" in chainData
+        ) {
+            const addr = (chainData.contractAddress as string | undefined)?.toLowerCase();
+            if (addr != null) {
```
guilleasz-crossmint: @devinai in general there are many unnecessary nested ifs in this funtion couldn't we add chainData.contractAddress != null in the first if? This happens with all the three if statements lets try not to add so many if levels

packages/wallets/src/wallets/wallet.ts
```diff
@@ -411,6 +413,53 @@ export class Wallet<C extends Chain> {
...
+                if (requestedTokenSet.has(addr)) return true;
```
guilleasz-crossmint: @devinai if we got to this point where `this.chain !== "stellar"` we definitely are not going to enter `this.chain === "stellar"` so can we just return `requestedTokenSet.has(addr) || requestedTokenSet.has(`${this.chain}:${addr}`) instead of having to ifs that return true, and we shortcircuit the funciton. I think we can do this in the three if statements, all are mutually exclusive so once they got to that point we can already return

guilleasz-crossmint (approved):

PR #1823, opened 2026-05-08
feat(react-ui, react-native): auto-derive SDK reference docs from JSDoc

jmderby:
## Description
                                                                                 
  Auto-derives React and React Native SDK reference docs from JSDoc on the source types instead of hand-maintained `MANUAL_RETURNS` blocks in the generator. The`useWallet` and `useWalletOtpSigner` Returns sections now flow from JSDoc on `CrossmintWalletBaseContext` and `OtpSignerFunctions` — change a comment in source, the docs update on the next `pnpm generate:docs`.                         

Also adds a v1/v0 versioning banner on every generated page (per-page link to the v0 equivalent + V1 migration guide).
                                                                                    
Net: -363 LOC, eliminates a class of doc/source drift (already caught one — `DeviceSignerDescriptor` had been renamed to `DeviceSignerConfig` but the manual block still said the old name).                                                   
                       
## Test plan

 * Ran `pnpm generate:docs` in both `react-ui` and `react-native` — output matches prior MDX modulo intentional improvements (correct types from current source,`@internal` members hidden)                                                       
 * Verified `useWallet` Returns section renders all 6 fields and`useWalletOtpSigner` renders all 4, both end-to-end from JSDoc                    
* Confirmed v1/v0 banner appears on all 8 pages with correct per-page links
* Sanity-tested by editing a JSDoc comment in source and regenerating — change appeared without touching any script                                              
                                                                                    
## Package updates                                                                


jmderby (commented):
Note to reviewer: output is identical to before — we're just no longer hardcoding type definitions to achieve it. As a side benefit, this also surfaces a few SDK methods/fields that the hardcoded blocks had drifted out of sync with.

This is part of a bigger initiative that i'm personally taking on to achieve doc synchronization & automation.

albertoelias-crossmint (approved):

PR #1824, opened 2026-05-08
feat: persist identity key in native Keychain via postMessage bridge

albertoelias-crossmint:
## Description

On iOS 17+, WebKit can aggressively evict IndexedDB storage from a hidden 1×1 WKWebView, even while the host app remains in the foreground. When this happens mid-onboarding, the frame generates a new ECDH identity key with a different `deviceId`, causing `complete-onboarding` to fail with `"Authentication for device ... is not pending"` (TEE 400).

This PR adds the **native (SDK) side** of a backup/restore mechanism that persists the identity key JWK in `expo-secure-store` (iOS Keychain / Android Keystore), so the frame can recover its identity if IndexedDB is wiped.

**How it works:**

1. **On WebView init** (`initializeWebView`): loads any existing backup from SecureStore and injects it as a `window.__CROSSMINT_IDENTITY_KEY_BACKUP` global via `injectedJavaScriptBeforeContentLoaded`.
2. **On message from frame** (`handleMessage`): intercepts `{ type: "identity-key-backup", jwk }` postMessages, persists the JWK to SecureStore (scoped by environment), and updates React state so mid-session WebView reloads also inject the backup.
3. **Schema** (`RNWebView.tsx`): extends `AllowedGlobalsSchema` to accept the new backup global.

**Companion PR:** [Crossmint/open-signer#200](https://github.com/Crossmint/open-signer/pull/200) — frame-side changes that make the identity key extractable, send the backup message on key generation, and restore from the injected global when IDB is empty.

### Updates since initial revision

- **Fixed: in-session WebView reload now injects the backup.** `handleMessage` now calls `setIdentityKeyBackup(parsed.jwk)` alongside the SecureStore write, so `secureGlobals` stays in sync and `onContentProcessDidTerminate` / `onRenderProcessGone` reloads will include the backup global.
- **Fixed: "saved" log timing.** Moved the success log into `.then()` on `setItemAsync` so it only fires after the write actually completes, not optimistically.
- **Fixed: JWK shape validation on both write and read paths.** `handleMessage` now guards with `typeof parsed.jwk === "object" && !Array.isArray(parsed.jwk)` before accepting the JWK. `initializeWebView` applies the same guard when loading from SecureStore, so a corrupted/non-object value in SecureStore cannot crash the component tree via `AllowedGlobalsSchema.parse()`.

## Test plan

- This feature requires end-to-end testing on a real iOS device to simulate the WebKit IDB eviction scenario (start onboarding → trigger eviction → complete onboarding should recover the original deviceId from the backup)
- Manual verification that the `identity-key-backup` message is correctly intercepted and persisted in SecureStore
- Manual verification that on subsequent WebView init, the backup is loaded and injected as a global before the frame content loads
- Manual verification that a mid-session WebView process termination + reload still injects the backup (covers the `setIdentityKeyBackup` fix)

**Review checklist for humans:**

- [ ] **State batching in `initializeWebView`**: After `await SecureStore.getItemAsync()`, both `setIdentityKeyBackup` and `setNeedsWebView` are called synchronously. Verify React 18+ batches them so the WebView always mounts with the backup already set.
- [ ] **Fire-and-forget save**: `SecureStore.setItemAsync` failure is only logged as a warning. Confirm this is acceptable (vs. retrying or surfacing the error).
- [ ] **Permissive schema**: `z.record(z.unknown())` accepts any object shape. The JWK typeof/Array guards in `handleMessage` and `initializeWebView` add runtime narrowing, but there is no deep JWK field validation. Verify this passthrough approach is intentional.
- [ ] **Double JSON.parse in `handleMessage`**: `rawData` may be parsed once in the new backup handler and again in the existing console-message handler below. This is safe (first is in try/catch) but is a minor inefficiency to be aware of.

## Package updates

- `@crossmint/client-sdk-rn-window`: patch
- `@crossmint/client-sdk-react-native-ui`: patch

Changeset added via `.changeset/persist-identity-key-native.md`.

Link to Devin session: https://crossmint.devinenterprise.com/sessions/ead2b09763054a279d1682972c1f3ea0

---

<a href="https://crossmint.devinenterprise.com/review/crossmint/crossmint-sdk/pull/1824" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/client/ui/react-native/src/providers/CrossmintWalletProvider.tsx
+++ b/packages/client/ui/react-native/src/providers/CrossmintWalletProvider.tsx
@@ -281,6 +281,7 @@ function CrossmintWalletProviderInternal({
 
                 if (
                     parsed?.type === "identity-key-backup" &&
+                    parsed.jwk != null &&
                     typeof parsed.jwk === "object" &&
                     !Array.isArray(parsed.jwk)
                 ) {
--- a/packages/client/ui/react-native/src/providers/CrossmintWalletProvider.tsx
+++ b/packages/client/ui/react-native/src/providers/CrossmintWalletProvider.tsx
@@ -279,7 +279,7 @@ function CrossmintWalletProviderInternal({
             try {
                 const parsed = JSON.parse(rawData);
 
-                if (parsed?.type === "identity-key-backup" && typeof parsed.jwk === "object" && parsed.jwk != null) {
+                if (parsed?.type === "identity-key-backup" && typeof parsed.jwk === "object" && !Array.isArray(parsed.jwk)) {
                     const backupKey = `crossmint_identity_key_backup_${parsedAPIKey.environment}`;
                     SecureStore.setItemAsync(backupKey, JSON.stringify(parsed.jwk))
                         .then(() => logger.info("react-native.wallet.identity-key-backup.saved"))
--- a/packages/client/ui/react-native/src/providers/CrossmintWalletProvider.tsx
+++ b/packages/client/ui/react-native/src/providers/CrossmintWalletProvider.tsx
@@ -268,7 +268,12 @@ function CrossmintWalletProviderInternal({
             // so the identity key survives WebKit IndexedDB eviction.
             try {
                 const parsed = JSON.parse(rawData);
-                if (parsed?.type === "identity-key-backup" && parsed.jwk != null && typeof parsed.jwk === "object" && !Array.isArray(parsed.jwk)) {
+                if (
+                    parsed?.type === "identity-key-backup" &&
+                    parsed.jwk != null &&
+                    typeof parsed.jwk === "object" &&
+                    !Array.isArray(parsed.jwk)
+                ) {
                     const backupKey = `crossmint_identity_key_backup_${parsedAPIKey.environment}`;
                     SecureStore.setItemAsync(backupKey, JSON.stringify(parsed.jwk))
                         .then(() => logger.info("react-native.wallet.identity-key-backup.saved"))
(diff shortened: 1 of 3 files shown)


packages/client/ui/react-native/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -346,6 +378,26 @@ function CrossmintWalletProviderInternal({
 
     const initializeWebView = async () => {
         logger.info("react-native.wallet.webview.init.start");
+
+        // Load identity key backup from native before mounting the WebView.
```
albertoelias-crossmint: @devinai can remove this comment?

packages/client/ui/react-native/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -257,6 +264,31 @@ function CrossmintWalletProviderInternal({
 
             const rawData = event.nativeEvent.data;
 
+            // Handle identity key backup from the frame — persists the JWK in native SecureStore
+            // so the identity key survives WebKit IndexedDB eviction.
+            try {
```
albertoelias-crossmint, line 279: @devinai lets move this below the frame-ready if and only do json.parse once and then have ifs depending on type or console log. lets limit the checks in the if to the absolutely necessary

packages/client/ui/react-native/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -257,6 +264,31 @@ function CrossmintWalletProviderInternal({
 
             const rawData = event.nativeEvent.data;
 
+            // Handle identity key backup from the frame — persists the JWK in native SecureStore
```
albertoelias-crossmint: @devinai unnecessary comment

PR #1854, opened 2026-05-22
feat(wallets): normalize EVM server signer derivation to use "evm" chain type with legacy fallback

guilleasz-crossmint:
## Description

Normalizes EVM server signer key derivation to use `"evm"` as the chain type string (instead of chain-specific strings like `"base-sepolia"`) while maintaining full backward compatibility with existing wallets that were created with chain-specific derivation paths.

### Architecture

- **`deriveServerSignerCandidates()`** — returns both primary (`"evm"`) and legacy (chain-specific) derivations
- **`resolveServerSignerDerivation()`** — single sync decision point: cache → legacy if it matches a known on-chain address → primary
- **`resolveServerSignerApiLocator()`** — one-liner returning `server:<address>` via the shared resolution
- **`resolveServerSigner()`** — async registration check trying primary then legacy against the API
- **`isRecoverySigner()`** — matches against both derivation addresses

### Key behaviors

| Scenario | Derivation used |
|----------|----------------|
| New wallet creation (`createWallet`) | `"evm"` (primary) |
| New signer addition (`addSigner`) | `"evm"` (primary) |
| `useSigner` on existing wallet | Try `"evm"` first; if not registered, try legacy chain-specific |
| `isRecoverySigner` comparison | Match against either derivation |
| Signing (`buildInternalSignerConfig`) | Uses whichever was resolved via `resolveServerSignerDerivation` |
| `removeSigner` / `send(options.signer)` | Uses `resolveServerSignerDerivation` with cache + API-sourced address checks |
| Solana / Stellar | Unchanged — `getChainType` returns `"solana"`/`"stellar"` |

### Backward compatibility mechanisms

- **`#resolvedServerSigner`** cache with ownership validation (`matchResolvedServerSigner`)
- **`#apiSourcedRecoveryAddress`** — preserves recovery signer's on-chain address when factory replaces API config with user-provided config
- **`#apiSourcedDelegatedAddresses`** — same pattern for delegated server signers
- Both are propagated through wallet subclass constructors (EVM, Solana, Stellar)

### Files changed

- `packages/wallets/src/signers/server/helpers/derive-server-signer.ts` — `deriveServerSignerCandidates`, updated `deriveServerSignerDetails`
- `packages/wallets/src/signers/server/helpers/get-chain-type.ts` — new `getChainType()` helper
- `packages/wallets/src/wallets/wallet.ts` — core dual-derivation logic, `resolveServerSignerDerivation`, cache management
- `packages/wallets/src/wallets/wallet-factory.ts` — API address preservation for recovery + delegated signers
- `packages/wallets/src/wallets/evm.ts`, `solana.ts`, `stellar.ts` — propagate new constructor fields, use `resolveServerSignerApiLocator`
- `packages/wallets/src/utils/server-key-derivation.ts` — JSDoc update
- Test files — new tests for `deriveServerSignerCandidates`, `getChainType`, derivation verification

## Test plan

- 41 server signer and key derivation tests pass locally
- CI: lint, build & test, smoke tests all passing (3/3)
- Tests cover: dual derivation for EVM chains, null legacy for non-EVM, `getChainType` mapping, key derivation determinism

## Package updates

- `@crossmint/wallets-sdk` — minor bump (changeset already added)

Link to Devin session: https://crossmint.devinenterprise.com/sessions/e30fabadb48e41729a72901fcce40d37
Requested by: @guilleasz-crossmint

---

<a href="https://crossmint.devinenterprise.com/review/crossmint/crossmint-sdk/pull/1854" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


packages/wallets/src/wallets/wallet.ts
```diff
@@ -928,6 +976,47 @@ export class Wallet<C extends Chain> {
             }
         }
 
+        // For server signers, try primary (evm) derivation first, fall back to legacy (chain-specific)
```
guilleasz-crossmint: @devinai this adds alot of logic to the funciton `resolveNonDeviceSigner` can we move it to a new one `resolveServerSigner` and just run this function here?

packages/wallets/src/wallets/wallet.ts
```diff
@@ -972,6 +1061,20 @@ export class Wallet<C extends Chain> {
...
+            // Use cached derivation when it belongs to this signer (handles legacy wallets).
+            if (this.#resolvedServerDerivation) {
+                const { primary, legacy } = deriveServerSignerCandidates(
+                    signer,
+                    this.chain,
+                    this.#apiClient.projectId,
+                    this.#apiClient.environment
+                );
+                const cachedAddr = this.#resolvedServerDerivation.derivedAddress;
+                if (cachedAddr === primary.derivedAddress || (legacy != null && cachedAddr === legacy.derivedAddress)) {
```
guilleasz-crossmint: @devinai aren't we doing this comparation and locator building in very similar ways al around the file, could we extract it into a method, and re use it for everything to look cleaner?

packages/wallets/src/signers/server/helpers/derive-server-signer.ts
```diff
@@ -30,8 +30,39 @@ export function deriveServerSignerDetails(
...
+    const chainType = getChainType(chain);
+
+    // Primary: use normalized chain type ("evm" | "solana" | "stellar")
+    const primaryBytes = deriveKeyBytes(signer.secret, projectId, environment, chainType);
+    const primaryAddress = deriveServerSignerAddress(primaryBytes, chain);
+
+    // Legacy: chain-specific derivation (only matters for EVM where chain !== chainType)
+    const chainStr = typeof chain === "string" ? chain : String(chain);
+    let legacy: { derivedKeyBytes: Uint8Array; derivedAddress: string } | null = null;
+    if (chainType === "evm" && chainStr !== "evm") {
```
albertoelias-crossmint, line 65: i think we can clean this up a bit where we save `const primary = { .... }`. then in this if we return primary + legacy, and outside of the if, only primary

packages/wallets/src/wallets/evm.ts
```diff
@@ -28,6 +28,7 @@ export class EVMWallet extends Wallet<EVMChain> {
                 options: Wallet.getOptions(wallet),
                 alias: wallet.alias,
                 recovery: Wallet.getRecovery(wallet),
+                apiRecoveryAddress: Wallet.getApiRecoveryAddress(wallet),
```
albertoelias-crossmint: is this only used for server signers? is it the public key of the server signer? Ideally, we don't expose this in the wallet instance as developers won't need access to it. Can we find a different way?

packages/wallets/src/wallets/wallet-factory.ts
```diff
@@ -342,13 +350,16 @@ export class WalletFactory {
                     return true;
                 }
                 if (inputSigner.type === "server") {
-                    const { derivedAddress } = deriveServerSignerDetails(
+                    const { primary, legacy } = deriveServerSignerCandidates(
```
albertoelias-crossmint, line 361: do we still need `deriveServerSignerDetails`?

packages/wallets/src/wallets/wallet.ts
```diff
@@ -252,6 +264,45 @@ export class Wallet<C extends Chain> {
         return this.#options;
     }
 
+    /**
+     * Derive both primary ("evm") and legacy (chain-specific) server signer candidates.
+     */
+    private deriveServerCandidates(signer: ServerSignerConfig) {
```
albertoelias-crossmint: should we call it the same? deriveServerSignerCandidates?

packages/wallets/src/wallets/wallet.ts
```diff
@@ -252,6 +264,45 @@ export class Wallet<C extends Chain> {
...
+    }
+
+    /**
+     * Return the cached derivation if it belongs to the given signer (i.e. its address
+     * matches either the primary or legacy candidate). Returns null otherwise.
+     */
+    private getOwnedCachedDerivation(
+        signer: ServerSignerConfig
+    ): { derivedKeyBytes: Uint8Array; derivedAddress: string } | null {
+        if (!this.#resolvedServerDerivation) {
```
albertoelias-crossmint: == null

should we call it resolvedServerSigner for clarity?

packages/wallets/src/wallets/wallet.ts
```diff
@@ -252,6 +264,45 @@ export class Wallet<C extends Chain> {
...
+     */
+    private deriveServerCandidates(signer: ServerSignerConfig) {
+        return deriveServerSignerCandidates(signer, this.chain, this.#apiClient.projectId, this.#apiClient.environment);
+    }
+
+    /**
+     * Return the cached derivation if it belongs to the given signer (i.e. its address
+     * matches either the primary or legacy candidate). Returns null otherwise.
+     */
+    private getOwnedCachedDerivation(
```
albertoelias-crossmint: dont think this funciton name is obvious enough for what it does

packages/wallets/src/wallets/wallet.ts
```diff
@@ -944,6 +1007,46 @@ export class Wallet<C extends Chain> {
...
+                legacy.derivedAddress === this.#apiSourcedRecoveryAddress
+            ) {
+                this.#resolvedServerDerivation = legacy;
+            } else {
+                this.#resolvedServerDerivation = primary;
+            }
+            this.#needsRecovery = false;
+            return true;
+        }
+        const tried = legacy
```
albertoelias-crossmint: the legacy checks should be != null

albertoelias-crossmint (commented):
Damn, we really need to spend time refactoring this (my bad lol), I'm finding it hard to review. I feel like there's so many places where we fetch the candidates and decide which ones to use. I left some nits that can help readability but doesnt fix the overall problem. can you see if there's a small iterative refactor or a better way to apply this change without setting it up in so many different places? no need to block this pr if not possible

packages/wallets/src/wallets/evm.ts
albertoelias-crossmint: should we set this for all signer types then? the name is too generic and wont apply only to server signers

packages/wallets/src/wallets/wallet-factory.ts
```diff
@@ -204,6 +205,19 @@ export class WalletFactory {
             signers = createArgs.signers as SignerResponse[];
         }
 
+        // Preserve the API-sourced recovery address so the wallet can identify
+        // legacy derivations even when the user-provided config replaces the API one.
+        const apiRecoveryAddress =
```
albertoelias-crossmint: shouldnt we do this for any recovery signer?

packages/wallets/src/wallets/wallet-factory.ts
```diff
@@ -204,6 +205,19 @@ export class WalletFactory {
...
+            apiRecovery.type === "server" && "address" in apiRecovery && !("secret" in apiRecovery)
+                ? (apiRecovery as { address: string }).address
+                : undefined;
+
+        // Preserve the API-sourced delegated signer addresses so the wallet can identify
+        // legacy derivations for delegated server signers (same pattern as apiRecoveryAddress).
+        const apiDelegatedSignerAddresses = (apiDelegatedSigners ?? [])
```
albertoelias-crossmint: same here

packages/wallets/src/wallets/wallet.ts
```diff
@@ -68,7 +68,7 @@ import type {
 import { type ApiSourcedServerSignerConfig, isApiSourcedServerSignerConfig, AuthRejectedError } from "../signers/types";
 import { assembleSigner } from "../signers";
 import { NonCustodialSigner } from "../signers/non-custodial";
-import { deriveServerSignerDetails } from "../signers/server";
+import { deriveServerSignerCandidates as deriveServerSignerCandidatesHelper } from "../signers/server";
```
albertoelias-crossmint: why do we rename?

packages/wallets/src/wallets/wallet.ts
```diff
@@ -98,18 +100,38 @@ export class Wallet<C extends Chain> {
     #initialSigners: SignerConfigForChain<C>[];
     #needsRecovery = false;
     #deviceSignerApproved = false;
+    #resolvedServerSigner: { derivedKeyBytes: Uint8Array; derivedAddress: string } | null = null;
```
albertoelias-crossmint: can we have a common type and re-use it?

packages/wallets/src/wallets/wallet.ts
```diff
@@ -944,6 +1054,47 @@ export class Wallet<C extends Chain> {
         throw new Error(`Signer "${locator}" is not registered in this wallet.`);
     }
 
+    /**
+     * Resolve a server signer: try primary ("evm") derivation first, fall back to legacy
+     * (chain-specific) derivation when checking registration. Sets #resolvedServerSigner
+     * to whichever derivation is on-chain. Returns true if the signer is the admin (recovery) signer.
+     */
+    private async resolveServerSigner(signer: ServerSignerConfig): Promise<boolean> {
```
albertoelias-crossmint, line 1060: confused as to why we need both this and resolveServerSignerDerivation

packages/wallets/src/wallets/wallet.ts
```diff
@@ -252,6 +282,74 @@ export class Wallet<C extends Chain> {
...
+            return this.#resolvedServerSigner;
+        }
+        return null;
+    }
+
+    /**
+     * Resolve which derivation (primary "evm" or legacy chain-specific) to use for a server signer.
+     * Priority: cached resolution → legacy if it matches a known on-chain address → primary.
+     */
+    private resolveServerSignerDerivation(signer: ServerSignerConfig): {
```
guilleasz-crossmint: @devinai is this function only used in resolveServerSignerApiLocator? can we then move all this function inside of it to not have naming confusions

packages/wallets/src/wallets/wallet.ts
guilleasz-crossmint: Yeah they are very similar but work differently and are used in different moments, `resolveServerSigner` is used when setting the signer, it checks the signer actually exists in the wallet making an api call, and has side effects letting the wallet know that if it doesn't need recovery. 
`resolveServerSignerDerivation` is used when doing operations such as send, where we have the information locally so we can derive the address without doing api calls and no side effects are necessary.

packages/wallets/src/wallets/evm.ts
guilleasz-crossmint: Actually this is only applied for server signers, for the other signers it would be ignored as we don't need it, we just need it for getting the derivative address. @devinai can we maybe add server to the names to make it more explicit that is for server signers

packages/wallets/src/wallets/wallet-factory.ts
guilleasz-crossmint: this is not necessary for the other signers, we won't need to use it. @devinai change this to `apiServerRecoveryAddress` to make it more explicit

packages/wallets/src/wallets/wallet-factory.ts
guilleasz-crossmint: @devinai same here

packages/wallets/src/wallets/wallet.ts
guilleasz-crossmint: It's because there's a private method on the Wallet class with the same name that wraps the imported function

packages/wallets/src/wallets/wallet.ts
guilleasz-crossmint: @devin apply this

packages/wallets/src/wallets/evm.ts
```diff
@@ -28,6 +28,8 @@ export class EVMWallet extends Wallet<EVMChain> {
                 options: Wallet.getOptions(wallet),
                 alias: wallet.alias,
                 recovery: Wallet.getRecovery(wallet),
+                apiServerRecoveryAddress: Wallet.getApiServerRecoveryAddress(wallet),
```
albertoelias-crossmint: lets go with apiRecoveryServerSignerAddress for clarity. api and server can clash easily and confuse

albertoelias-crossmint (approved):
leaving approved with a small comment

PR #1859, opened 2026-05-26
feat(expo): add semantic locators for React native SDK to be used in E2E Mobile tests

luans-qa:
## Summary

Adds `testID` props to interactive elements in the Expo wallet playground app so Maestro's `id:` selector can target them in E2E tests.

The Maestro flows, CI workflow, and test utilities live in the [crossmint-mobile-e2e-tests](https://github.com/Crossmint/crossmint-mobile-e2e-tests) repo. This PR only ships the semantic locators the tests depend on.

### testIDs added

| File | IDs |
|------|-----|
| `apps/wallets/expo/src/App.tsx` | `logout-button` |
| `apps/wallets/expo/snippets/02-login-screen.tsx` | `email-input`, `send-code-button`, `otp-input`, `verify-button` |
| `apps/wallets/expo/snippets/03-wallet-display.tsx` | `copy-address-button`, `wallet-address` |
| `apps/wallets/shared/src/components/BalanceCard.tsx` | `{token}-balance`, `fund-button`, `refresh-balance-button` |
| `apps/wallets/shared/src/components/TransferForm.tsx` | `token-selector-{t}`, `recipient-input`, `amount-input`, `transfer-button`, `successful-tx-link` |
| `apps/wallets/shared/src/components/ActivityList.tsx` | `load-activity-button` |
| `apps/wallets/shared/src/components/ApprovalTest.tsx` | `approval-recipient-input`, `approval-amount-input`, `prepare-tx-button`, `transaction-id-input`, `approve-tx-button`, `approve-sig-button` |

### Other changes

- `metro.config.js` — adds `apps/wallets/shared/` to `watchFolders` so Metro hot-reloads shared component changes
- `.gitignore` — excludes `.env.local`
- `README.md` — updates env file reference from `.env` to `.env.local`

## Test plan

- [x] Login flow passes on iOS simulator (iPhone 16e, iOS 26.2) — verified locally
- [x] Login flow passes on Android emulator (Medium_Phone_API_36.1, API 36) — verified locally


.github/workflows/e2e-mobile-regression-tests.yml
```diff
@@ -0,0 +1,371 @@
...
+              with:
+                  api-level: 29
+                  arch: x86_64
+                  profile: pixel_2
+                  avd-name: test
+                  force-avd-creation: false
+                  emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim
+                  disable-animations: true
+                  script: |
+                      adb install apps/wallets/expo/android/app/build/outputs/apk/debug/app-debug.apk
```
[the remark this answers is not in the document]
luans-qa: Fixed in f15a11a4 — bumped to `api-level: 35` (Android 15), which is the closest available level to the locally-verified API 36. `reactivecircus/android-emulator-runner@v2` doesn't support API 36 yet, so 35 is the best option.

.github/workflows/e2e-mobile-regression-tests.yml
```diff
@@ -0,0 +1,371 @@
...
+                      adb reverse tcp:8081 tcp:8081
+                      cd apps/wallets/expo
+                      CI=1 npx expo start --port 8081 --localhost &
+                      sleep 15
```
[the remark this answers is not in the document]
luans-qa: Fixed in f15a11a4 — replaced the fixed `sleep 15` with the same polling loop used by the iOS job. Metro is now polled every 2s for up to 60s; tests start as soon as the bundler is ready.

apps/wallets/expo/tests/specs/auth/login.yaml
```diff
@@ -0,0 +1,43 @@
...
+  MAILOSAUR_API_KEY: ${MAILOSAUR_API_KEY}
+  MAILOSAUR_SERVER_ID: ${MAILOSAUR_SERVER_ID}
+---
+# Step 1: Generate a unique test email for this run (mirrors web test pattern)
+- runScript:
+    file: ../../../shared/utils/generateEmail.js
+
+# Step 2: Launch app with clean state (clear keychain to force login screen)
+- launchApp:
+    clearKeychain: true
```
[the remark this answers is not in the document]
luans-qa: Good catch — fixed in f9289448. Added `clearState: true` alongside `clearKeychain: true`. Without it, Android auth tokens persisted in AsyncStorage/SecureStore across runs, causing the login screen to never appear and the test to fail with a misleading "element not visible" error.

.github/workflows/e2e-mobile-regression-tests.yml
```diff
@@ -0,0 +1,377 @@
...
+            - name: Install Maestro
+              run: |
+                  curl -Ls "https://get.maestro.mobile.dev" | bash
+                  echo "$HOME/.maestro/bin" >> $GITHUB_PATH
+
+            - name: Run iOS Maestro tests
+              id: run-ios-tests
+              working-directory: apps/wallets/expo
+              env:
+                  MAILOSAUR_API_KEY: ${{ secrets.MAILOSAUR_API_KEY }}
```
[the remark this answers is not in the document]
luans-qa: Valid concern. The `mobile-dev-inc/action-maestro-cloud` action runs tests on Maestro Cloud (their SaaS), not on our own runners, so it's not a direct substitute here. The practical mitigation is to pin a specific version in the install URL, e.g. `curl -Ls "https://get.maestro.mobile.dev" | VERSION=1.40.0 bash`, so the runner always pulls a known release rather than latest. Happy to apply that if you want to enforce it — otherwise the risk profile is similar to other `curl | bash` installers already in use in the repo (pnpm, etc.) and is bounded to the CI sandbox with no persistent access.

.github/workflows/e2e-mobile-regression-tests.yml
albertoelias-crossmint, line 1: are we running this already? can we run this without running on a real device or simulator?

.github/workflows/e2e-mobile-regression-tests.yml
```diff
@@ -0,0 +1,375 @@
...
+    maestro-android:
+        name: Android E2E Tests
+        runs-on: ubuntu-latest
+        timeout-minutes: 60
+
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v4
+
+            - name: Install pnpm
```
albertoelias-crossmint: why do we need to run this and other steps twice? cant we have a combined setup?

.github/workflows/e2e-mobile-regression-tests.yml
```diff
@@ -0,0 +1,375 @@
...
+            - name: Parse results and notify Slack
+              env:
+                  SLACK_WEBHOOK: ${{ secrets.SLACK_QA_ALERTS_WEBHOOK }}
+                  GITHUB_RUN_ID: ${{ github.run_id }}
+                  GITHUB_REPOSITORY: ${{ github.repository }}
+                  GITHUB_SHA: ${{ github.sha }}
+                  IOS_OUTCOME: ${{ needs.maestro-ios.result }}
+                  ANDROID_OUTCOME: ${{ needs.maestro-android.result }}
+              run: |
+                  node << 'EOF'
```
albertoelias-crossmint: cant we move this to a script in the repo that then gets called from here?

apps/wallets/expo/tests/specs/auth/login.yaml
albertoelias-crossmint, line 1: do we need tests/e2e/specs? cant they all fall under tests/specs? i feel like we're repeating too much and so many folders arent providing useful information

albertoelias-crossmint, line 1: where are the actual tests? these feel like steps to run but we're not making any assertions

albertoelias-crossmint (commented):
left some questions and comments to clean it up a little

.github/workflows/e2e-mobile-regression-tests.yml
luans-qa: Not yet — the workflow file needs to be on `main` before GitHub enables `workflow_dispatch` for it. Once the PR is merged we can trigger it manually from the Actions tab.

Regarding running without a device: Maestro requires a live simulator or emulator — it drives the actual native app UI. There's no headless mode for mobile E2E. The CI jobs use macos-latest (iOS Simulator, no physical device) and ubuntu-latest + KVM-accelerated Android emulator, so no real hardware is needed.

.github/workflows/e2e-mobile-regression-tests.yml
luans-qa: GitHub Actions parallel jobs run on completely isolated runners (separate VMs), so setup steps can't be shared between them — each job must install its own dependencies. This is a fundamental GHA constraint, not a design choice.

Making them sequential (single job) would fix the duplication but would require the whole job to run on macOS (since iOS needs it), losing KVM acceleration for Android and making total runtime significantly longer. Parallel with duplicated setup is the right tradeoff here.

.github/workflows/e2e-mobile-regression-tests.yml
luans-qa: Done in 13aaddf8 — extracted to `apps/wallets/expo/scripts/notify-slack.js`. The workflow step is now a single line: `node apps/wallets/expo/scripts/notify-slack.js`.

apps/wallets/expo/tests/specs/auth/login.yaml
luans-qa: Agreed — fixed in 13aaddf8. Flattened to `tests/specs/auth/login.yaml`, removing the redundant `e2e/` level.

luans-qa: The assertions are the two `extendedWaitUntil` steps — they fail the entire flow if the expected element doesn't appear within the timeout. In Maestro, every step is an implicit assertion: if any step fails (element not found, tap target missing, script throws), the test fails.

Concretely:
- `extendedWaitUntil id: email-input timeout: 30000` — asserts the login screen loaded
- `extendedWaitUntil id: logout-button timeout: 30000` — asserts auth succeeded and the wallet dashboard is visible

The intermediate steps (inputText, tapOn) also assert implicitly — they fail if the target element doesn't exist.

.github/workflows/e2e-mobile-regression-tests.yml
luans-qa, line 1: Fixed in ab1cb5dd. Added an explicit readiness check after each polling loop in both the iOS step and the Android emulator-runner script:

```bash
if ! curl -sf http://localhost:8081/status 2>/dev/null | grep -q "packager-status:running"; then
  echo "::error::Metro did not start within 60s — aborting to avoid misleading test failures"
  exit 1
fi
```

Both jobs now abort early with a clear diagnostic instead of proceeding into Maestro with a broken JS bundle.

luans-qa:
Good catch from Devin — the `notify-slack` job was missing `actions/checkout@v4`, so `node apps/wallets/expo/scripts/notify-slack.js` would fail with `MODULE_NOT_FOUND` on a fresh runner. Fixed in 73edd59a.

albertoelias-crossmint (approved):
leaving approved. i think there are some pending ai comments that look relevant

luans-qa:
> leaving approved. i think there are some pending ai comments that look relevant

Thanks! the scope for this PR was changed, so no new changes are needed

PR #1870, opened 2026-05-29
fix(ci): parallelize nightly E2E across browsers to prevent timeout

luans-qa:
## Problem

The nightly E2E job has been hitting the 60-minute job limit every day and cancelling before any tests complete.

**Root cause**: 60 tests (20 per browser × 3 browsers) run serially on `workers: 1`. With `retries: 2` and transfer tests that individually time out at 2 minutes, the full suite needs ~3 hours — 3× the job limit.

Timeline from today's cancelled run ([26626516698](https://github.com/Crossmint/crossmint-sdk/actions/runs/26626516698)):
- `08:26:19` — "Running 60 tests using 1 worker"
- `08:26:55` — first test fails, triggers retry cycle
- `08:33:51` — solana transfer timeout (2 min) consumes retry budget
- `09:21:32` — job cancelled after exactly 60 min with ~17/60 tests completed

## Fix

Split the single job into a **matrix of 3 parallel browser jobs** (chromium, firefox, webkit). Each job runs only its 20 tests and has a 90-minute limit, which it comfortably fits within.

| Before | After |
|--------|-------|
| 1 job, 60 serial tests, 60 min limit | 3 parallel jobs, 20 tests each, 90 min limit |
| Cancelled every day before finishing | Each browser completes independently |
| One timed-out Slack notification | One combined Slack message with per-browser breakdown |

### Other improvements
- **Install only the needed browser per job** (`playwright install --with-deps ${{ matrix.browser }}`) instead of all three — saves ~2 min of setup per job
- **`fail-fast: false`** so a webkit failure doesn't cancel the chromium and firefox jobs
- **Separate `notify` job** aggregates results from all 3 artifacts into a single Slack message

## Test plan
- [ ] Trigger via `workflow_dispatch` and confirm 3 parallel jobs appear
- [ ] Confirm all 3 jobs complete within their 90-minute limit
- [ ] Confirm a single combined Slack notification fires after all 3 finish


.github/workflows/e2e-regression-tests.yml
```diff
@@ -67,240 +69,228 @@ jobs:
...
+              continue-on-error: true
+
+            - name: Parse results and send to Slack
               env:
                   SLACK_WEBHOOK: ${{ secrets.SLACK_QA_ALERTS_WEBHOOK }}
                   GITHUB_RUN_ID: ${{ github.run_id }}
                   GITHUB_REPOSITORY: ${{ github.repository }}
                   GITHUB_SHA: ${{ github.sha }}
+                  MATRIX_RESULT: ${{ needs.e2e-regression-tests.result }}
               run: |
```
albertoelias-crossmint: can we move this to a script like we did in the other repo?

luans-qa:
Done — extracted to `apps/wallets/quickstart-devkit/scripts/notify-slack.js`, alongside the existing `generate-test-report.js`. The workflow step is now a single `node apps/wallets/quickstart-devkit/scripts/notify-slack.js` call (c62c5d2e).

albertoelias-crossmint (approved):

PR #1878, opened 2026-06-01
fix(react-native): prevent iOS WKWebView JS throttling from zero-area clipping

panosinthezone:
## Description

On iOS, a WKWebView hidden via `width: 0, height: 0, overflow: "hidden"` gets its `ActivityState::IsVisible` flag cleared by WebKit. This causes `updateThrottleState()` to drop the foreground assertion on the WebContent process, making it eligible for suspension, timer throttling, and priority termination under memory pressure — which explains intermittent JS execution failures in the signer relay WebView.

**Fix:** Replace the zero-area clipped wrapper with `width: 1, height: 1, opacity: 0, pointerEvents: "none"`. This keeps the compositing layer present so iOS treats the WebView as visible, maintaining full JS execution priority. Matches the Flutter SDK's `Offstage` pattern.

**WebKit source references** (from [WebKit/WebKit](https://github.com/WebKit/WebKit)):
- `WebPageProxy::isViewVisible()` → checks `ActivityState::IsVisible` flag (`WebPageProxy.cpp:3257`)
- `updateThrottleState()` → `isViewVisible() ? takeVisibleActivity() : dropVisibleActivity()` (`WebPageProxy.cpp:3546`)
- `PageClientImpl::isActiveViewVisible()` → `isViewInWindow() && !_isBackground` (`PageClientImplIOS.mm:163`)

See [detailed source code walkthrough](https://github.com/Crossmint/crossmint-sdk/pull/1878#issuecomment-4596988817) in PR comments.

## Test plan

* Verify on physical iOS devices that wallet creation/signing flows work consistently
* Monitor for reduced `react-native.wallet.webview.process.terminated` events
* Test under memory pressure scenarios (open many apps, then return)
* Confirm WebView remains invisible and non-interactive to users

## Package updates

* `@crossmint/client-sdk-react-native-ui`: patch (changeset added)

Link to Devin session: https://crossmint.devinenterprise.com/sessions/aa42293821a24fee86c25e350715efd5
Requested by: @panosinthezone

### DIFF
--- a/.changeset/fix-rn-webview-ios-visibility.md
+++ b/.changeset/fix-rn-webview-ios-visibility.md
@@ -2,4 +2,4 @@
 "@crossmint/client-sdk-react-native-ui": patch
 ---
 
-Fix iOS WKWebView JS throttling caused by zero-area clipping. Replace `overflow: hidden` + `width: 0, height: 0` wrapper with `opacity: 0` + `pointerEvents: none` so the WebView backing layer stays composited and iOS does not deprioritize the WebContent process. Add diagnostic instrumentation (mount timing, bridge first-message, heartbeat).
+Fix iOS WKWebView JS throttling caused by zero-area clipping. Replace `overflow: hidden` + `width: 0, height: 0` wrapper with `opacity: 0` + `pointerEvents: none` so the WebView backing layer stays composited and iOS does not deprioritize the WebContent process.
--- a/.changeset/fix-rn-webview-ios-visibility.md
+++ b/.changeset/fix-rn-webview-ios-visibility.md
@@ -1,0 +1,5 @@
+---
+"@crossmint/client-sdk-react-native-ui": patch
+---
+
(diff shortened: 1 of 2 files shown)


albertoelias-crossmint (approved):

calderaro:
Confirmed on the Flutter SDK — the hidden signer WebView is already 1×1, not 0×0, so it is not exposed to the zero-area throttling path this PR fixes.

`hidden_signer_bridge_host.dart` build():
- `SizedBox(width: 1, height: 1)` — never 0×0.
- `Opacity(0.01)`, not `0.0` — Flutter's `RenderOpacity` skips painting the child at alpha 0, which would detach the platform-view WKWebView from compositing (the equivalent failure).
- `Opacity` + `IgnorePointer`, not the `Offstage` widget — `Offstage` gives zero layout size + skipped paint, same failure mode.

One correction: this PR description (and the Flutter dartdoc it cites) say the fix matches Flutter's `Offstage` pattern. Flutter does not use `Offstage` here — it uses the Opacity/IgnorePointer wrapper above. Opening a Flutter doc fix (branch `fix/hidden-webview-visibility-docs`) to correct the dartdoc and add a why-comment so a refactor to real `Offstage`/opacity-0 doesn't silently reintroduce this bug.

PR #1911, opened 2026-06-09
fix(wallets): wipe server signer secret from memory after key derivation

albertoelias-crossmint:
## Description

Resolves [WAL-10250](https://linear.app/crossmint/issue/WAL-10250/review-secret-retention-in-server-signer-memory-model).

The server signer memory model retained both the plaintext master secret and derived private key bytes in memory for the wallet's entire lifetime. The master secret is strictly more sensitive — it can derive keys for all chains/environments. This PR reduces secret exposure by wiping key material as early as possible after use.

### Changes

**Dedicated recovery cache (`wallet.ts`)**
- Split the server signer cache into two purpose-built fields:
  - `#resolvedServerSigner` — delegated signer cache, reset on each `useSigner()` call
  - `#resolvedRecoveryServerSigner` — recovery signer cache, set once during recovery resolution, never cleared
- This ensures `stripSecretFromRecovery()` is safe: the recovery key material persists in `#resolvedRecoveryServerSigner` even when `useSigner()` is called with a different (delegated) signer.

**Secret stripping (`wallet.ts`)**
- `resolveServerSigner()`: when the signer resolves as the **recovery** signer, caches to `#resolvedRecoveryServerSigner` and calls `stripSecretFromRecovery()` to replace `#recovery` with address-only config.
- `resolveServerSignerDerivation()`: for `ApiSourcedServerSignerConfig`, returns `#resolvedRecoveryServerSigner` directly. For `ServerSignerConfig`, derives candidates and checks both caches before falling through to heuristic picking. Wipes temporary candidate key bytes in all paths.
- `wipeNonSelectedCandidate()`: zeros the non-selected candidate's `derivedKeyBytes`.

**Guard updates (`wallet.ts`)**
- `withRecoverySigner()` and `resumePendingDeviceSignerApproval()`: check `#resolvedRecoveryServerSigner == null` to determine if recovery is available.
- `isAutoAssemblableSignerConfig()`: server case checks `#resolvedRecoveryServerSigner != null`.
- `buildInternalSignerConfig()`: copies `derivedKeyBytes`, then wipes the source if it's a short-lived temporary (not either cached reference).

**Key material wiping (`wallet.ts`)**
- `useSigner()`: wipes old `#resolvedServerSigner.derivedKeyBytes` before nulling.
- `resolveServerSigner()` error path: wipes both candidates before throwing.
- `isRecoverySigner()`/`resolveAddresses`: wipes candidates after extracting addresses.

**Signer adapters** (`evm-server-signer.ts`, `solana-server-signer.ts`, `stellar-server-signer.ts`)
- Constructors call `secureWipe(config.derivedKeyBytes)` after building internal key material.

**New utility** (`utils/secure-wipe.ts`)
- `secureWipe(...buffers)`: zeros `Uint8Array` buffers, skips `null`/`undefined`.

## Test plan

* Pre-existing unit tests (server-signers, server-key-derivation, signer-mapping) all pass
* Lint passes (`pnpm lint`)
* CI build & test, lint, and smoke tests all pass

## Package updates

- `@crossmint/wallets-sdk`: patch (changeset: `.changeset/wipe-server-signer-secret-from-memory.md`)

Link to Devin session: https://crossmint.devinenterprise.com/sessions/45a9e564ea2d4fd7ab2f14e4d2d62972

---

<a href="https://crossmint.devinenterprise.com/review/crossmint/crossmint-sdk/pull/1911" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/wallets/src/wallets/wallet.ts
+++ b/packages/wallets/src/wallets/wallet.ts
@@ -1153,8 +1153,8 @@ export class Wallet<C extends Chain> {
     /**
      * Wipe the derivedKeyBytes of the non-selected candidate to reduce key material in memory.
      */
-    private wipeNonSelectedCandidate(_selected: DerivedServerSigner, other: DerivedServerSigner | null): void {
-        if (other != null && other !== _selected) {
+    private wipeNonSelectedCandidate(selected: DerivedServerSigner, other: DerivedServerSigner | null): void {
+        if (other != null && other !== selected) {
             secureWipe(other.derivedKeyBytes);
         }
     }
@@ -1477,7 +1477,7 @@ export class Wallet<C extends Chain> {
         pendingOperation: { type: "signature" | "transaction"; id: string }
     ): Promise<void> {
         const originalSigner = this.#signer;
-        if (isApiSourcedServerSignerConfig(this.#recovery)) {
+        if (isApiSourcedServerSignerConfig(this.#recovery) && this.#resolvedServerSigner == null) {
             throw new Error(
                 "Cannot resume pending approval: no secret available. " +
                     'Call wallet.useSigner({ type: "server", secret: ... }) first with the recovery server secret.'
--- a/packages/wallets/src/wallets/wallet.ts
+++ b/packages/wallets/src/wallets/wallet.ts
@@ -1951,6 +1951,9 @@ export class Wallet<C extends Chain> {
                 // Copy the derived key bytes so the signer adapter can safely wipe its copy
                 // without corrupting the cached #resolvedServerSigner reference.
                 const keyBytesCopy = new Uint8Array(resolved.derivedKeyBytes);
+                if (resolved !== this.#resolvedServerSigner) {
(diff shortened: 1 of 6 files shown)


packages/wallets/src/wallets/wallet.ts
```diff
@@ -1190,7 +1233,10 @@ export class Wallet<C extends Chain> {
 
     private async withRecoverySigner<T>(operation: () => Promise<T>): Promise<T> {
         const originalSigner = this.signer;
-        if (isApiSourcedServerSignerConfig(this.#recovery)) {
+        if (
+            isApiSourcedServerSignerConfig(this.#recovery) &&
+            (this.#resolvedServerSigner == null || this.#resolvedServerSigner.derivedAddress !== this.#recovery.address)
```
albertoelias-crossmint: @devinai we do this check many times, can we abstract it?

guilleasz-crossmint (approved):

PR #1926, opened 2026-06-16
refactor(wallets): introduce ChainAdapter to centralize per-chain behavior

albertoelias-crossmint:
## What

Phase 2 of the `wallet.ts` decomposition: moves the per-chain `switch (this.chain)` branches into a `ChainAdapter` strategy, one per chain. wallet.ts: 2,109 → ~2,050 lines.

- `chains/chain-adapter.ts` — `ChainAdapter` interface, `getChainAdapter(chain)`, and `getChainAdapterByType(chainType)` (returns `null` for unsupported)
- `chains/adapters/{evm,solana,stellar}.ts` — the three implementations
- `wallet.ts` + `services/balance-formatter.ts` dispatch through the adapter; `getSignerRegistrationChain` and `isSolanaWallet` removed
- `PendingSignerOperation` defined once in `wallets/types.ts` (deferred Phase-1 cleanup)

## Included fixes (from review)

Two correctness items surfaced in review are folded in:

1. **Signatures are EVM-only.** Investigation showed Stellar message-signing isn't a usable feature — no `createSignature` entry point exists for Stellar (only `EVMWallet` has `signMessage`/`signTypedData`), and the Stellar external-wallet + non-custodial signers reject `signMessage`. The original guard blocked only Solana, letting Stellar fall through. The Stellar adapter's `supportsSignatures` is now `false`, so the `"only supported for EVM smart wallets"` error is accurate. Transaction approval is unaffected. *(Was briefly PR #1927; folded here.)*
2. **`signers()` is now chain-agnostic + accurate error.** The hard-coded `chainType` literals are replaced with `getChainAdapterByType(chainType) == null`, and the unsupported-chain case now throws `Wallet chain type <x> not supported` instead of the misleading `Wallet type smart not supported` (WAL-10673).

## Design note

`transactionPayloadToSign` and transaction-hash extraction were deliberately **not** moved into the adapter — they dispatch on the API *response* shape, not the wallet chain, so a per-chain adapter would be wrong. They stay in wallet.ts / operation-poller.ts.

## Guarantees

- Error classes, message strings, and `walletsLogger` payloads preserved byte-for-byte for the pure-refactor parts; the two behavior changes above are each captured in their own changeset.
- Validated against the local characterization suite (~200 Wallet-behavior tests) + an adversarial diff review.

## Test plan

- `pnpm --filter @crossmint/wallets-sdk test:vitest` → **612 passed / 0 failed / 68 skipped**
- `pnpm --filter @crossmint/wallets-sdk build` → green (DTS heap raised to 8GB; see commit)
- biome clean


packages/wallets/src/chains/adapters/evm.ts
```diff
@@ -0,0 +1,50 @@
+import type { AddSignerChain, AddSignerContext, ChainAdapter } from "../chain-adapter";
+import type { RegisterSignerChain, RegisterSignerResponse, Signer as APISigner } from "../../api";
+import type { Chain } from "../chains";
+import type { PendingSignerOperation, TokenBalance } from "../../wallets/types";
+import { walletsLogger } from "../../logger";
+import { InvalidSignerError } from "../../utils/errors";
+import { getPendingSignerOperation } from "../../utils/signer-mapping";
+
+export const evmChainAdapter: ChainAdapter = {
```
guilleasz-crossmint, line 9: Why chain adapters are an object and not a class?

packages/wallets/package.json
```diff
@@ -17,7 +17,7 @@
     "types": "./dist/index.d.ts",
     "files": ["dist", "LICENSE", "README.md"],
     "scripts": {
-        "build": "pnpm generate && tsup",
+        "build": "pnpm generate && cross-env NODE_OPTIONS=--max-old-space-size=8192 tsup",
```
guilleasz-crossmint, line 20: Why?

packages/wallets/src/wallets/wallet.ts
```diff
@@ -1941,7 +1885,7 @@ export class Wallet<C extends Chain> {
     }
 
     protected async approveSignatureInternal(signatureId: string, options?: ApproveOptions) {
-        if (this.isSolanaWallet) {
+        if (!this.chainAdapter.supportsSignatures) {
             throw new Error("Approving signatures is only supported for EVM smart wallets");
```
guilleasz-crossmint, line 1887: Should we include stellar in the error message, or should stellar not support signatures?

packages/wallets/src/chains/adapters/evm.ts
guilleasz-crossmint: Actually the way it is being used later as a getter it kind of makes sense to use an object so we are not instantiated an instance each time is being used, unless everything is static, as there is actually no reference to `this` in the object, whatever you prefer for me is ok

guilleasz-crossmint (approved):

packages/wallets/package.json
albertoelias-crossmint: as we add more files, we went over. i added  a task for the end of the refactor to look into this. we have the same issue with react sdks

packages/wallets/src/wallets/wallet.ts
guilleasz-crossmint: I really liked the fact that the wallet.ts is completly chain agnostic, this is the only place of the code that still the chain appear I think we can change this to be `getChainAdapter(walletResponse.chainType) == null` instead.
Also the errror message seems off as it only shows the wallet type and not the chain. I think it is impossible for the type to be smart but fail because of an invalid chain, but as the code is there we can or remove the chain check or improve the error message

PR #1947, opened 2026-06-24
Re-onboard the iOS signer per signature with ephemeral webview storage

tomas-martins-crossmint:
On iOS the non-custodial signer runs in a hidden `WebView`, and that webview's storage isn't reliable across launches, so the device identity can disappear and break signing. This stops relying on that storage on iOS, leaving Android unchanged.

The signer frame now uses a non-persistent store with in-memory device storage, and reloads before each signature so it re-onboards with a fresh OTP. It also covers the frame reloading mid-onboarding: if the web content process is terminated between sending and verifying the code, the signer requests a new code and keeps the prompt open instead of failing the signature.

## Changes
- On iOS, build the signer frame URL with `deviceStorage=memory`, set the `WebView` to non-persistent storage via `incognito`, and reload the frame before each signature. Only applied on iOS.
- Add an iOS-only `resetSignerFrame` option to `wallets-sdk`, threaded through the signer descriptors.
- Re-issue the OTP when the frame reloads mid-onboarding, surfacing `OnboardingSessionExpiredError` so the prompt can show a fresh code instead of rejecting.
- Track a `connectionGeneration` on `HandshakeParent` to detect that reload.
- Keep the OTP prompt open and tell the user a new code was sent.
- Add unit tests for the per-signature reset and the mid-onboarding recovery.

---

<a href="https://crossmint.devinenterprise.com/review/crossmint/crossmint-sdk/pull/1947" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/rn-ios-ephemeral-signer-frame.md
+++ b/.changeset/rn-ios-ephemeral-signer-frame.md
@@ -1,0 +1,9 @@
+---
+"@crossmint/wallets-sdk": patch
+"@crossmint/client-sdk-react-base": patch
+"@crossmint/client-sdk-react-native-ui": patch
+---
+
+On iOS, the React Native non-custodial signer stops relying on the signer webview's storage, which isn't reliable across launches and could drop the device identity and break signing. The signer frame now uses non-persistent storage with in-memory device storage, and reloads to re-onboard with a fresh OTP before each signature.
+
+Android keeps its existing persistent behavior.
--- a/.changeset/rn-ios-ephemeral-signer-frame.md
+++ b/.changeset/rn-ios-ephemeral-signer-frame.md
@@ -1,9 +1,11 @@
 ---
-"@crossmint/wallets-sdk": patch
-"@crossmint/client-sdk-react-base": patch
-"@crossmint/client-sdk-react-native-ui": patch
+"@crossmint/wallets-sdk": minor
+"@crossmint/client-sdk-react-native-ui": minor
+"@crossmint/client-sdk-react-ui": minor
+"@crossmint/client-sdk-react-base": minor
+"@crossmint/client-sdk-window": minor
 ---
 
-On iOS, the React Native non-custodial signer stops relying on the signer webview's storage, which isn't reliable across launches and could drop the device identity and break signing. The signer frame now uses non-persistent storage with in-memory device storage, and reloads to re-onboard with a fresh OTP before each signature.
+On iOS the non-custodial signer stops relying on the signer webview's storage, which isn't reliable across launches and could drop the signer and break signing. The frame now uses non-persistent storage with in-memory key storage, and reloads to re-onboard with a fresh OTP before each signature. Android keeps its existing persistent behavior.
 
-Android keeps its existing persistent behavior.
+It also recovers the OTP flow when the frame reloads mid-onboarding: the signer detects the reload, requests a fresh code, and keeps the prompt open so the user can enter the new one.
--- a/.changeset/signer-otp-reissue-on-frame-reload.md
+++ b/.changeset/signer-otp-reissue-on-frame-reload.md
@@ -1,7 +1,0 @@
----
-"@crossmint/client-sdk-window": patch
-"@crossmint/wallets-sdk": patch
-"@crossmint/client-sdk-react-base": patch
----
-
-Recover the non-custodial signer OTP flow when the signer frame reloads between sending and verifying the code. This happens with the ephemeral iOS frame if the web content process is terminated mid-onboarding, which would otherwise lose the in-memory device and dead-end with an OTP error. The signer now detects the reload, requests a fresh code, and keeps the prompt open so the user can enter the new one.
--- a/.changeset/signer-otp-reissue-on-frame-reload.md
+++ b/.changeset/signer-otp-reissue-on-frame-reload.md
@@ -1,0 +1,7 @@
+---
+"@crossmint/client-sdk-window": patch
+"@crossmint/wallets-sdk": patch
+"@crossmint/client-sdk-react-base": patch
+---
+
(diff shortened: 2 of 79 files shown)


.changeset/signer-otp-reissue-on-frame-reload.md
albertoelias-crossmint, line 1: do we need 2 changesets if we're modifying the same modules in the same pr?

.changeset/rn-ios-ephemeral-signer-frame.md
```diff
@@ -0,0 +1,9 @@
+---
+"@crossmint/wallets-sdk": patch
```
albertoelias-crossmint: this should be a minor release as it's a significant change

packages/client/ui/react-native/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -50,6 +50,10 @@ export interface CrossmintWalletProviderProps {
 
 const MAX_HANDSHAKE_RETRIES = 2;
 
+const USES_EPHEMERAL_DEVICE_STORAGE = Platform.OS === "ios";
```
albertoelias-crossmint: would specify this is for tee signers as device is normally associate with device signer

packages/wallets/src/signers/non-custodial/ncs-signer.ts
```diff
@@ -334,13 +353,41 @@ export abstract class NonCustodialSigner implements SignerAdapter {
...
+        if (frameReloaded) {
+            walletsLogger.warn("complete-onboarding: signer frame reloaded mid-onboarding, re-issuing OTP", {
+                onboardingGeneration: this._onboardingConnectionGeneration,
+                currentGeneration: connection?.connectionGeneration,
+            });
+            // Re-run onboarding so the backend issues a fresh OTP against the reloaded frame. Do not
+            // reject the auth promise: the signing flow continues so the user can enter the new code.
+            await this.sendMessageWithOtp();
+            throw new OnboardingSessionExpiredError();
+        }
```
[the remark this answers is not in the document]
tomas-martins-crossmint: Addressed in 3c99993c. If the re-issue itself fails, we now reject the flow (and surface the error) instead of throwing OnboardingSessionExpiredError, so the dialog no longer stays open over a dead promise.

packages/client/react-base/src/hooks/useSignerAuth.ts
```diff
@@ -74,6 +74,12 @@ export function useSignerAuth(): SignerAuthState & SignerAuthHandlers {
             setEmailSignerDialogStep("initial");
             setActiveAuthEmail(undefined);
         } catch (error) {
+            if (error instanceof OnboardingSessionExpiredError) {
+                // The signer frame was reloaded mid-onboarding and a fresh code was already sent.
+                // Keep the dialog open so the user can enter the new code.
+                console.warn("Signer session expired, a new code was sent");
+                return;
+            }
```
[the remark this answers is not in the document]
tomas-martins-crossmint: Addressed in 386b063a. OnboardingSessionExpiredError is now re-thrown to the OTP input, which shows "That code expired. We sent a new one. Enter it below." instead of silently returning.

packages/client/ui/react-native/src/providers/CrossmintWalletProvider.tsx
tomas-martins-crossmint: Good call. Renamed the flag to `USES_EPHEMERAL_SIGNER_STORAGE` and added a comment noting it's the non-custodial (TEE) signer webview, not the device signer (979ccba9).

packages/client/ui/react-native/src/providers/CrossmintWalletProvider.tsx
```diff
@@ -50,6 +50,12 @@ export interface CrossmintWalletProviderProps {
 
 const MAX_HANDSHAKE_RETRIES = 2;
 
+// Applies to the non-custodial (TEE) signer webview, not the device signer. On iOS that webview's
+// storage isn't reliable across launches, so we keep its key in memory and reload before each signature.
+const USES_EPHEMERAL_SIGNER_STORAGE = Platform.OS === "ios";
```
albertoelias-crossmint, line 56: would specify all these constants only apply to tee signers

albertoelias-crossmint (approved):

packages/wallets/src/signers/non-custodial/ncs-signer.ts
```diff
@@ -334,13 +342,38 @@ export abstract class NonCustodialSigner implements SignerAdapter {
...
+                currentGeneration: connection?.connectionGeneration,
+            });
+            try {
+                await this.sendMessageWithOtp();
+            } catch (reissueError) {
+                this._authPromise?.reject(reissueError as Error);
+                throw reissueError;
+            }
+            throw new OnboardingSessionExpiredError();
+        }
```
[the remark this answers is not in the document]
albertoelias-crossmint: @tomas-martins-crossmint unsure about this one, please look into it and check whats right

packages/client/ui/react-native/src/providers/CrossmintWalletProvider.tsx
tomas-martins-crossmint: done, clarified that in the comment

packages/wallets/src/signers/non-custodial/ncs-signer.ts
```diff
@@ -307,9 +317,7 @@ export abstract class NonCustodialSigner implements SignerAdapter {
             });
         } catch (err) {
             walletsLogger.error("complete-onboarding: error", { error: err });
-            this._needsAuth = true;
-            this._authPromise?.reject(err as Error);
-            throw err;
+            return await this.handleOnboardingVerificationFailure(err as Error);
         }
```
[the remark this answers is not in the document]
albertoelias-crossmint: @tomas-martins-crossmint we should address this

PR #1960, opened 2026-06-30
fix: inject React context into DeviceSignerModule for Android

tomas-martins-crossmint:
## Description

On some low-end Android devices (e.g. HOTWAV Cyber 13 with Spreadtrum T606), the `DeviceSignerContextHolder.Initializer` ContentProvider may not have set `appContext` by the time the Expo native module accesses `KeystoreKeyStorage`. Because `prefs` was evaluated via `by lazy`, it cached `null` permanently — making every key lookup fail, triggering an infinite recovery loop that generated 55+ device signers.

This PR changes `DeviceSignerModule` to pass `appContext.reactContext` directly to `DeviceSignerStorageFactory.create(context)`, bypassing the ContentProvider entirely. The Expo module framework guarantees `reactContext` is available when the module is instantiated.

```kotlin
// Before
private val storage by lazy {
    DeviceSignerStorageFactory.create()  // relies on ContentProvider
}

// After
private val storage by lazy {
    val context = appContext.reactContext
        ?: throw IllegalStateException("React context not available")
    DeviceSignerStorageFactory.create(context)  // direct injection
}
```

Companion PR: https://github.com/Crossmint/crossmint-kotlin-sdk/pull/119 (adds `create(context)` overload)

## Test plan

* Verified the change compiles against the existing `sdk-device-signer` API with the new `create(context)` overload
* The fix eliminates the ContentProvider race condition that caused the SpotPay P0 incident

## Package updates

* `@crossmint/client-sdk-react-native-ui`: patch — changeset added via `.changeset/rn-android-inject-context-device-signer.md`
* Requires `com.crossmint:sdk-device-signer` >= 0.0.16 (published from companion Kotlin SDK PR)

Link to Devin session: https://crossmint.devinenterprise.com/sessions/b48df64129a448fb97a5d8d614ed9688
Requested by: @tomas-martins-crossmint

### DIFF
--- a/.changeset/rn-android-inject-context-device-signer.md
+++ b/.changeset/rn-android-inject-context-device-signer.md
@@ -1,5 +1,5 @@
 ---
-"@crossmint/client-sdk-react-native-ui": patch
+"@crossmint/client-sdk-react-native-ui": minor
 ---
 
 Fix Android device signer key lookup failures on low-end devices by injecting the React context directly into `KeystoreKeyStorage` instead of relying on a `ContentProvider` to set it via a static singleton.
--- a/.changeset/rn-android-inject-context-device-signer.md
+++ b/.changeset/rn-android-inject-context-device-signer.md
@@ -1,0 +1,9 @@
+---
+"@crossmint/client-sdk-react-native-ui": patch
+---
+
+Fix Android device signer key lookup failures on low-end devices by injecting the React context directly into `KeystoreKeyStorage` instead of relying on a `ContentProvider` to set it via a static singleton.
+
+The `DeviceSignerModule` now passes `appContext.reactContext` to `DeviceSignerStorageFactory.create(context)`, bypassing the `DeviceSignerContextHolder` initialization race that caused `SharedPreferences` to be permanently unavailable on some devices.
+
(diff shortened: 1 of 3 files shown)


.changeset/rn-android-inject-context-device-signer.md
```diff
@@ -0,0 +1,9 @@
+---
+"@crossmint/client-sdk-react-native-ui": patch
```
albertoelias-crossmint: we should use minor releases if we're updating native code

albertoelias-crossmint (approved):

.changeset/rn-android-inject-context-device-signer.md
```diff
@@ -0,0 +1,9 @@
+---
+"@crossmint/client-sdk-react-native-ui": minor
```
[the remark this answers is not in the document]
tomas-martins-crossmint: this is intended

.changeset/rn-android-inject-context-device-signer.md
tomas-martins-crossmint: done!

PR #1968, opened 2026-07-06
fix(e2e): reuse persistent test wallets to stop devnet faucet failures

luans-qa:
## Summary

Fixes the chronic `Wallet E2E › crossmint - solana - email/phone › transfers funds` failures in the scheduled [E2E Regression Tests](https://github.com/Crossmint/crossmint-sdk/actions/runs/28778603481) (failing on all 3 browsers).

**Root cause:** every run created a brand-new wallet and requested a SOL airdrop from the public devnet faucet, which is permanently rate-limited (`429 Too Many Requests`) from shared GitHub Actions IPs — ~18 airdrop requests per run (2 configs × 3 retries × 3 browsers).

**Fix — reuse wallets so faucets are only needed rarely:**
- Test emails are now deterministic (`test-email-e2e@…`, `test-phone-e2e@…`) instead of timestamped, so the same wallets — and their funding — persist across runs, retries, and browsers. Rotate to a fresh identity via `TESTS_WALLET_EMAIL_SUFFIX` if ever needed. The Stellar alias is derived deterministically as well.
- `fundWalletWithSolAirdrop` checks the on-chain SOL balance first and skips the airdrop when the wallet is already funded. When funding is needed it requests the maximum with fallbacks (5 → 2 → 1 → 0.1 SOL), so a single successful airdrop keeps the wallet funded for thousands of transfers. Devnet RPC is overridable via `SOLANA_DEVNET_RPC_URL` (e.g. a Helius key can be slotted in later with no code change).
- The USDXM faucet is only called when the balance is below the transfer amount, and transfers were reduced from 10 to 0.01 USDXM so one 10 USDXM funding lasts ~1000 runs.

Faucet pressure drops from ~18 requests/run to at most 4, and to 0 once the wallet is seeded.

## Test plan

- [x] `tsc --noEmit` and `biome check` clean on changed files (only pre-existing warnings)
- [x] `playwright test --list` discovers all 75 tests
- [x] Ran `crossmint - solana - email` locally: deterministic-email auth passes, persistent wallet received 10 USDXM, balance-check + airdrop fallback chain executes as designed (airdrop itself still 429'd — faucet was dry from this IP today; first successful airdrop anywhere seeds the wallet permanently)
- [ ] Monitor the next scheduled E2E regression runs

### DIFF
--- a/apps/wallets/quickstart-devkit/tests/e2e/specs/wallets/e2e.spec.ts
+++ b/apps/wallets/quickstart-devkit/tests/e2e/specs/wallets/e2e.spec.ts
@@ -54,7 +54,7 @@ test.describe("Wallet E2E", { tag: "@critical" }, () => {
                 const walletAddress = await getWalletAddress(authenticatedPage);
                 // Transfer the minimum practical amount so the reused wallet's funds
                 // last across many runs without needing the faucet.
-                const transferAmount = "0.01";
+                const transferAmount = "0.001";
 
                 // Only hit the faucet when the reused wallet doesn't have enough USDXM
                 const initialBalance = await getWalletBalance(authenticatedPage);
@@ -117,7 +117,7 @@ test.describe("Wallet E2E", { tag: "@critical" }, () => {
 
                 // Transfer the minimum practical amount so the reused wallet's funds
                 // last across many runs without needing the faucet.
-                const transferAmount = "0.01";
+                const transferAmount = "0.001";
                 const walletAddress = await getWalletAddress(authenticatedPage);
 
                 // Only hit the faucet when the reused wallet doesn't have enough USDXM
--- a/apps/wallets/quickstart-devkit/tests/e2e/specs/wallets/e2e.spec.ts
+++ b/apps/wallets/quickstart-devkit/tests/e2e/specs/wallets/e2e.spec.ts
@@ -52,27 +52,27 @@ test.describe("Wallet E2E", { tag: "@critical" }, () => {
 
             test("transfers funds", async ({ authenticatedPage, testConfig }) => {
                 const walletAddress = await getWalletAddress(authenticatedPage);
+                // Transfer the minimum practical amount so the reused wallet's funds
+                // last across many runs without needing the faucet.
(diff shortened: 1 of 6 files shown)


luans-qa:
Re: greptile review

- **Issue 1 (stale balance in final error log)** — fixed in cba81cd2: the balance is now re-read after the retry loop, and if an airdrop landed on-chain despite a confirmation timeout the test continues instead of failing.
- **Issue 2 (concurrent browsers may triple-fund on first run)** — accepted as-is: it only affects the one-time seeding run for a brand-new wallet identity (the wallet ends up with 30 USDXM instead of 10, which just lasts longer). Cross-job coordination isn't worth the complexity; once seeded, all browsers skip the faucet entirely.

apps/wallets/quickstart-devkit/tests/e2e/specs/wallets/e2e.spec.ts
```diff
@@ -115,21 +115,20 @@ test.describe("Wallet E2E", { tag: "@critical" }, () => {
                     throw new Error(`Unknown chain: ${(testConfig as { chain: string }).chain}`);
                 }
 
-                const transferAmount = "10";
+                // Transfer the minimum practical amount so the reused wallet's funds
+                // last across many runs without needing the faucet.
+                const transferAmount = "0.01";
```
maxsch-xmint: I would lower this as much as possible if this test will run on every commit

maxsch-xmint (approved):

apps/wallets/quickstart-devkit/tests/e2e/specs/wallets/e2e.spec.ts
luans-qa: Lowered to 0.001 USDXM in 6d10b921 (all 4 transfer spots in e2e + smoke). That's the proven default of the transferFunds util, so each 10 USDXM faucet seed now lasts ~5,000 transfers. Didn't go to the absolute floor (0.000001 = 1 base unit at 6 decimals) since any rounding in the UI/SDK would zero it out.

PR #1975, opened 2026-07-09
fix(wallets): expose SDK error details for geo-blocked requests

albertoelias-crossmint:
## Description

Fixes [WAL-11049](https://linear.app/crossmint/issue/WAL-11049/expose-sdk-error-details-for-geo-blocked-requests). When a user's requests are geo-blocked by Cloudflare (403 with an HTML body), the SDK previously surfaced only a generic `status: "error"` — integrators (e.g. SpotPay in Myanmar) couldn't tell a permanent block from a transient network blip, and the underlying failure was an opaque `SyntaxError` from calling `.json()` on HTML.

This is the **minimal** version of the fix. Rather than the ticket's "throw a typed error for all non-`ok` responses", which would break the ~20 call sites that rely on 4xx returning a JSON error body (`if ("error" in result)`), the `throwIfCrossmintApiAuthError` JWT mapping, and the 422 idempotency path in `approveSignature`, this only newly throws for non-`ok` responses whose body **isn't JSON** — which is exactly the Cloudflare-HTML case that was producing the `SyntaxError`.

Two changes:

1. **`ApiClient.makeRequest`** (`packages/common/base`) — keep the existing `>= 500` throw, and additionally throw a typed `ApiClientError` (carrying HTTP `status` + body) when `!response.ok` **and** the response content-type is not `application/json`. JSON 4xx responses still pass through unchanged.

2. **`CrossmintWalletBaseProvider`** (`packages/client/react-base`) — add an `error` field to the wallet context beside `status`:
   ```ts
   error: { code: "region-blocked" | "network" | "unknown"; status?: number; message: string } | null
   ```
   A `mapWalletError()` helper classifies the thrown error: `403 → region-blocked` (permanent, don't retry); fetch reject/`TypeError` / `5xx` / `429 → network` (transient, retryable); else `unknown`. It's set in the `getWallet`/`getOrCreateWallet`/`createWallet` catch blocks and cleared on start/success/logout. Because `getOrCreateWallet` auto-retries via the `createOnLogin` effect when `status` flips to `"error"`, the loop is now gated to bail out on a `region-blocked` code so it stops hammering a permanently-blocked endpoint.

`error` flows through `useWallet()` automatically, so react-ui and react-native consumers get it with no further changes.

## Test plan

* New unit tests in `packages/common/base/src/apiClient/api-client.test.ts` cover `makeRequest`: non-ok + non-JSON body → throws `ApiClientError` with status/body; 5xx → throws; non-ok + JSON body → passes through; ok → passes through.
* `pnpm --filter @crossmint/common-sdk-base test:vitest` and `pnpm --filter @crossmint/client-sdk-react-base test:vitest` both pass.
* Build (`pnpm --filter @crossmint/client-sdk-react-base... build`) and lint pass with no new warnings.

## Package updates

* `@crossmint/common-sdk-base`: patch
* `@crossmint/client-sdk-react-base`: minor (new `error` context field)

Changeset added: `.changeset/geo-blocked-wallet-error-details.md`.

Link to Devin session: https://crossmint.devinenterprise.com/sessions/6eb653772dbe4a10bdaff1435299a166

### DIFF
--- a/packages/client/react-base/src/providers/CrossmintWalletBaseProvider.tsx
+++ b/packages/client/react-base/src/providers/CrossmintWalletBaseProvider.tsx
@@ -277,8 +277,6 @@ export function CrossmintWalletBaseProvider({
             if (wallet != null) {
                 return wallet;
             }
-            // Don't hammer a permanently-blocked endpoint: this callback is re-fired by the
-            // createOnLogin effect whenever status flips to "error", so stop retrying region blocks.
             if (walletError?.code === "region-blocked") {
                 return undefined;
             }
--- a/packages/client/react-base/src/providers/CrossmintWalletBaseProvider.tsx
+++ b/packages/client/react-base/src/providers/CrossmintWalletBaseProvider.tsx
@@ -16,7 +16,7 @@ import {
 } from "@crossmint/wallets-sdk";
 import type { HandshakeParent } from "@crossmint/client-sdk-window";
 import type { signerInboundEvents, signerOutboundEvents } from "@crossmint/client-signers";
-import type { UIConfig } from "@crossmint/common-sdk-base";
+import { ApiClientError, type UIConfig } from "@crossmint/common-sdk-base";
 import { useCrossmint, useSignerAuth } from "@/hooks";
 import type { CreateOnLogin } from "@/types";
 import cloneDeep from "lodash.clonedeep";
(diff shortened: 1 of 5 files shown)


packages/client/react-base/src/providers/CrossmintWalletBaseProvider.tsx
```diff
@@ -238,9 +277,15 @@ export function CrossmintWalletBaseProvider({
             if (wallet != null) {
                 return wallet;
             }
+            // Don't hammer a permanently-blocked endpoint: this callback is re-fired by the
```
albertoelias-crossmint: @devinai unnecessary comment

packages/client/react-base/src/providers/CrossmintWalletBaseProvider.tsx
```diff
@@ -116,6 +132,28 @@ export type PasskeyPromptState = {
...
+    if (error instanceof ApiClientError) {
+        if (error.status === 403) {
+            return { code: "region-blocked", status: error.status, message: error.message };
+        }
+        if (error.status >= 500 || error.status === 429) {
+            return { code: "network", status: error.status, message: error.message };
+        }
+        return { code: "unknown", status: error.status, message: error.message };
+    }
+    // A rejected fetch (no response) surfaces as a TypeError; treat as transient network failure.
```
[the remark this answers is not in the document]
albertoelias-crossmint: @devinai propose a more specific fix

packages/client/react-base/src/providers/CrossmintWalletBaseProvider.tsx
```diff
@@ -116,6 +132,41 @@ export type PasskeyPromptState = {
...
 };
 
+/**
+ * Cloudflare serves an HTTP 403 with an HTML body when it bans the requester's country/region.
+ * Its error code 1009 ("Access denied: ... has banned the country or region your IP address is
+ * in") is the distinguishing signal — it lets us treat a genuine geo-block as permanent without
+ * also labelling every other non-JSON 403 (nginx access controls, a CDN/reverse-proxy rule) as
+ * `region-blocked` and permanently suppressing the auto-retry loop.
+ */
+const CLOUDFLARE_REGION_BLOCK_PATTERN = /error\s*1009|banned\s+(?:the\s+)?(?:country|region)/i;
```
alberto-crossmint, line 142: A bit weak comparison, but if no dedicated error field, then it lgtm!

alberto-crossmint (approved):

PR #1976, opened 2026-07-10
feat(wallets): auto-inject a device signer on Solana wallet creation

alberto-crossmint:
## What

Makes Solana `createWallet()` auto-inject a device signer by default (when `deviceSignerKeyStorage` is configured), reaching parity with EVM and Stellar. Previously a `chain !== "solana"` guard in `WalletFactory.createWallet` excluded Solana, so the recovery signer ended up as the default for Solana transactions.

## Why the guard existed — and why it's now safe to remove

The guard was added in #1830 because a Solana wallet's provider isn't known until after the wallet is created, and not every provider supports device signers — those that don't reject a device signer at creation time. Naively removing the guard would break wallet creation for those wallets.

This PR removes the guard **with a defensive retry**: when creation is rejected with the stable `DEVICE_SIGNER_NOT_SUPPORTED` error code — and only when the SDK auto-injected the device signer — creation is retried once without it. This reuses the same error code the SDK already handles in `addSigner` / `recover()`.

## How

- Removed the `&& chain !== "solana"` guard; the device signer is injected for all chains when `deviceSignerKeyStorage` is set.
- New `createSmartWallet()` helper: attempts creation with the device signer; on a `DEVICE_SIGNER_NOT_SUPPORTED` response for an auto-injected signer, strips it and retries once. The wallet's existing post-creation `recover()` flow then handles the recovery-signer fallback, as before.
- Guardrails: a generic error surfaces **without** retry (never masks unrelated failures); an **explicitly-supplied** device signer is never silently stripped.

## Tests

`wallet-factory.test.ts` (32 → 35 passing):
- Solana now injects a device signer at creation.
- Retries without the device signer on `DEVICE_SIGNER_NOT_SUPPORTED`.
- Non-device errors are not retried.
- An explicit device signer is not stripped on rejection.

## Notes

- On the retry path, the rejected eager attempt leaves one orphaned temporary device key (unmapped, since the wallet doesn't exist yet) — this mirrors existing EVM-on-failure behavior; there's no `deleteKey`-by-temp-id API, so no cleanup complexity was added.

### DIFF
--- a/packages/wallets/src/wallets/wallet-factory.ts
+++ b/packages/wallets/src/wallets/wallet-factory.ts
@@ -126,15 +126,8 @@ export class WalletFactory {
             );
         }
 
-        // Include a device signer in the wallet's initial signer set when a device-key storage is
-        // available (client-side). This is done uniformly across EVM, Stellar and Solana so the
-        // device signer is the default across chains. Solana providers without device-signer support
-        // (e.g. Squads, the current production default) reject a device signer at creation with the
-        // stable DEVICE_SIGNER_NOT_SUPPORTED error code, and the SDK cannot tell which provider backs
-        // the wallet before it is created. We therefore attempt eager registration and, if the backend
-        // rejects it, transparently retry creation without the device signer (see createSmartWallet).
-        // The wallet's post-creation flow then registers a device signer for supported providers or
-        // falls back to the recovery signer for those that don't.
+        // Inject a device signer as the default on every chain when key storage is available.
+        // Some providers reject it at creation; createSmartWallet retries without it in that case.
         const eagerlyAddDeviceSigner = validatedArgs.options?.deviceSignerKeyStorage != null;
         const explicitSigners = validatedArgs.signers ?? [];
         const didAutoInjectDeviceSigner = eagerlyAddDeviceSigner && !explicitSigners.some((s) => s.type === "device");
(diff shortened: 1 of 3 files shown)


packages/wallets/src/wallets/wallet-factory.ts
```diff
@@ -182,6 +176,51 @@ export class WalletFactory {
...
+        const signersWithoutDeviceSigner = builtSigners.filter((s) => !this.isBuiltDeviceSigner(s));
+        return await this.apiClient.createWallet(buildParams(signersWithoutDeviceSigner));
+    }
+
+    /** Matches the auto-injected device signer (object form); locator-based device signers are strings. */
+    private isBuiltDeviceSigner(
+        builtSigner: { signer: string } | RegisterSignerParams | { signer: PasskeySignerConfig }
+    ): boolean {
+        const signer = builtSigner.signer;
+        return typeof signer === "object" && signer != null && "type" in signer && signer.type === "device";
```
albertoelias-crossmint: Don't we always have a type?

albertoelias-crossmint (approved):

packages/wallets/src/wallets/wallet-factory.ts
alberto-crossmint: Good catch — every object-form built signer (passkey or device) always carries a `type`, so `"type" in signer` was redundant. Dropped it in b142cc12; the check is now just `typeof signer === "object" && signer != null && signer.type === "device"`.

PR #1997, opened 2026-07-29
refactor(wallets): extract shared quorum member matching (M4-4 part 1/4)

panosinthezone:
Linear: [WAL-11292](https://linear.app/crossmint/issue/WAL-11292/m4-4-sdk-quorum-aware-recovery-identity-usesigner-disambiguation) · EDD §6.5 · Stacked on M4-3 (#1993) · **Part 1/4 of the M4-4 stack** (split from #1996)

## Summary

Pure refactor + new shared utility, no behavior change. Extracts the quorum-member matching that `WalletFactory.isMatchingQuorumMember` did inline into `src/utils/quorum-members.ts`, so the `useSigner` selection flows (part 4) and `SignerManager` (part 2) can reuse the exact same per-type semantics:

- `matchesQuorumMember(candidate, member, serverCandidateAddresses)` — per-type matching (email normalized, phone exact, external-wallet by address, passkey by id → name → permissive, server by derived addresses incl. legacy). Server derivation is injected so each caller plugs its own provider.
- `getQuorumMemberLocator(member)` — API locator when present, derived fallback otherwise.
- `ResolvedQuorumMember` type: API identity fields + index signature admitting runtime fields (server `secret`, external-wallet `onSign`, passkey callbacks) that later parts graft on.

The factory now delegates to the shared matcher and — the one incidental fix — wipes the derived key bytes after the address comparison (`secureWipe`), which the inline version omitted.

## Testing

- New `quorum-members.test.ts` matcher table (all member types, normalization, permissive passkey, server candidate injection).
- Full `packages/wallets` suite: 708 passed.

### DIFF
--- a/packages/wallets/src/signers/types.ts
+++ b/packages/wallets/src/signers/types.ts
@@ -122,6 +122,22 @@ export function isApiSourcedServerSignerConfig(config: { type: string }): config
 export type RecoverySignerConfigForChain<C extends Chain> = SignerConfigForChain<C> | ApiSourcedServerSignerConfig;
 
 /**
+ * A member of a resolved quorum admin signer. API-sourced members carry the identity fields
+ * (per-member `locator`, passkey `id`, server/external-wallet `address`); the index signature
+ * additionally admits runtime fields grafted from the caller's config (server `secret`,
+ * external-wallet `onSign`, passkey `onSignWithPasskey`), which the API cannot store.
+ */
+export type ResolvedQuorumMember = Record<string, unknown> & {
+    type: string;
+    locator?: string;
+    address?: string;
+    email?: string;
+    phone?: string;
+    id?: string;
+    name?: string;
+};
+
+/**
  * A quorum admin signer as **resolved by the API** (read side): members live under `signers`
  * and the derived `quorum:<id>` locator is present. The create-side input counterpart is
  * `QuorumRecoveryConfig`, whose members live under `methods`.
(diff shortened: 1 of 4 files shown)


packages/wallets/src/signers/types.ts
```diff
@@ -121,6 +121,22 @@ export function isApiSourcedServerSignerConfig(config: { type: string }): config
...
+ */
+export type ResolvedQuorumMember = Record<string, unknown> & {
+    type: string;
+    locator?: string;
+    address?: string;
+    email?: string;
+    phone?: string;
+    id?: string;
+    name?: string;
+};
```
alberto-crossmint, line 138: Shouldn't this be a discriminated union?

packages/wallets/src/utils/quorum-members.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Locator of a resolved quorum member. API-sourced members carry their locator; for members
+ * that don't (e.g. caller-supplied configs on the no-API-config fallback path), fall back to
+ * deriving one from the config fields.
+ */
+export function getQuorumMemberLocator(member: ResolvedQuorumMember): string {
+    if (typeof member.locator === "string") {
+        return member.locator;
+    }
```
alberto-crossmint, line 14: How can it not have a locator?

alberto-crossmint (approved):

PR #1998, opened 2026-07-29
feat(wallets): quorum member plumbing in SignerManager and server resolver (M4-4 part 2/4)

panosinthezone:
Linear: [WAL-11292](https://linear.app/crossmint/issue/WAL-11292/m4-4-sdk-quorum-aware-recovery-identity-usesigner-disambiguation) · EDD §6.5 · **Part 2/4 of the M4-4 stack** (split from #1996)

## Summary

The `SignerManager` + `ServerSignerResolver` plumbing the `useSigner` member-selection flows (part 4) sit on. Dormant until part 4 lands — nothing can select a quorum member yet — but each piece is unit-tested here on its own contract:

- **`SignerManager` member APIs**: `quorumMemberLocators()`, `matchQuorumMembers()` (via the part-1 shared matcher), `adoptQuorumMemberConfig()` (replaces one member with the API-merged config and records the selection in an adopted-locator set — deliberately never the quorum-wide `adoptRecoveryConfig`), `adoptedAssemblableQuorumMember()` (only members the caller explicitly selected this session, and only if they can auto-assemble).
- **Per-member secret strip**: `stripSecretFromRecovery()` gains the quorum arm — resolved server members drop their `secret`, members without an API address pass through.
- **Actionable errors replace `QuorumSignerNotSupportedError`** in `require()` and `withRecoverySigner()`: they now list the member locators and point at `wallet.useSigner()`. `withRecoverySigner()` (used by `addSigner`/`removeSigner`) runs with the active signer when it is a quorum member — no reassembly, no silent member picking (EDD-strict).
- **`ServerSignerResolver`**: quorum server-member addresses join `apiRecoveryAddress` as admin addresses (`#isAdminAddress`), so a legacy derivation matching a member wins the pick and is cached as the recovery resolution. `Wallet`'s constructor feeds the member addresses in.

## Testing

- `signer-manager.test.ts`: require/withRecoverySigner quorum arms, adoption + assemblable-member gating, per-member secret strip.
- `resolver.test.ts`: legacy derivation via quorum member address, cached as recovery resolution.
- Full `packages/wallets` suite: 719 passed.

### DIFF
--- a/packages/wallets/src/signers/server/resolver.ts
+++ b/packages/wallets/src/signers/server/resolver.ts
@@ -52,11 +52,16 @@ export class ServerSignerResolver {
 
     resolveDerivation(config: ServerSignerConfig | ApiSourcedServerSignerConfig): DerivedServerSigner {
         if (isApiSourcedServerSignerConfig(config)) {
-            if (this.#resolvedRecoveryServerSigner != null) {
-                return this.#resolvedRecoveryServerSigner;
+            // The cache holds one admin identity; with quorum server members there can be
+            // several, so serve it only for the address it was resolved for — never
+            // substitute another member's key material.
+            const cached = this.#resolvedRecoveryServerSigner;
+            if (cached != null && cached.derivedAddress === config.address) {
+                return cached;
             }
             throw new Error(
-                "Cannot resolve server signer derivation: no secret available and no cached recovery resolution. " +
+                `Cannot resolve server signer derivation for "server:${config.address}": ` +
+                    "no secret available and no cached recovery resolution for that address. " +
                     'Call wallet.useSigner({ type: "server", secret: ... }) first.'
             );
         }
--- a/packages/wallets/src/signers/server/resolver.ts
+++ b/packages/wallets/src/signers/server/resolver.ts
@@ -14,6 +14,7 @@ export class ServerSignerResolver {
     readonly #projectId: string;
     readonly #environment: string;
     readonly #apiRecoveryAddress: string | null;
+    readonly #apiQuorumServerMemberAddresses: string[];
     readonly #apiDelegatedAddresses: string[];
     readonly #knownOnChainAddresses: () => string[];
 
@@ -25,6 +26,7 @@ export class ServerSignerResolver {
         projectId: string;
         environment: string;
         apiRecoveryAddress: string | null;
+        apiQuorumServerMemberAddresses?: string[];
         apiDelegatedAddresses: string[];
         knownOnChainAddresses: () => string[];
     }) {
@@ -32,8 +34,16 @@ export class ServerSignerResolver {
         this.#projectId = params.projectId;
         this.#environment = params.environment;
         this.#apiRecoveryAddress = params.apiRecoveryAddress;
+        this.#apiQuorumServerMemberAddresses = params.apiQuorumServerMemberAddresses ?? [];
         this.#apiDelegatedAddresses = params.apiDelegatedAddresses;
         this.#knownOnChainAddresses = params.knownOnChainAddresses;
+    }
+
+    // The admin identity is either the single recovery signer's address or any quorum
+    // member's — a legacy derivation matching either must win over the primary, so the
+    // assembled adapter's locator lines up with the API's member locator.
+    #isAdminAddress(address: string): boolean {
+        return address === this.#apiRecoveryAddress || this.#apiQuorumServerMemberAddresses.includes(address);
     }
 
     deriveCandidates(config: ServerSignerConfig): { primary: DerivedServerSigner; legacy: DerivedServerSigner | null } {
@@ -62,7 +72,7 @@ export class ServerSignerResolver {
         }
 
         if (legacy != null) {
-            if (legacy.derivedAddress === this.#apiRecoveryAddress) {
+            if (this.#isAdminAddress(legacy.derivedAddress)) {
                 secureWipe(primary.derivedKeyBytes);
                 return legacy;
             }
@@ -129,7 +139,7 @@ export class ServerSignerResolver {
     }
 
     #selectRecovery(primary: DerivedServerSigner, legacy: DerivedServerSigner | null): DerivedServerSigner {
-        if (this.#apiRecoveryAddress != null && legacy != null && legacy.derivedAddress === this.#apiRecoveryAddress) {
+        if (legacy != null && this.#isAdminAddress(legacy.derivedAddress)) {
             this.#resolvedRecoveryServerSigner = legacy;
             this.#wipeNonSelectedCandidate(legacy, primary);
             return legacy;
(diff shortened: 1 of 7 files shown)


packages/wallets/src/signers/server/resolver.ts
alberto-crossmint, line 1: Claude found:

**The resolver returns the cached member's key for a different member's address**

  packages/wallets/src/signers/server/resolver.ts:52-62 + signers/descriptors/server.ts

  resolveDerivation()'s API-sourced branch returns #resolvedRecoveryServerSigner without comparing config.address:

  if (isApiSourcedServerSignerConfig(config)) {
      if (this.#resolvedRecoveryServerSigner != null) {
          return this.#resolvedRecoveryServerSigner;   // ← no address check
      }

  Pre-PR this was sound: there was exactly one admin server signer, so the single slot could only ever hold it. #apiQuorumServerMemberAddresses breaks that invariant — a quorum now has several admin-side server identities sharing one slot.

  Reproduced. After selecting member A, asking for member B's key material returns A's:

  ✗ B: the resolver hands back the cached member's key for a DIFFERENT member's address
      - Expected  0xc4FE5bABae8C7e9210E1FcA549eFf650C157F632   ← member B
      + Received  0x532b3a57A2765EB3DcBcd2348AA5e91F78D6b6fC   ← member A

  ✗ C: canAutoAssemble reports a secret-less member assemblable off another member's cache
      expected { type: 'server', … } to be null

  Repro C is the gate that lets this through: serverSignerDescriptor.canAutoAssemble returns ctx.serverSigners.hasRecoveryResolution — a global boolean, not a per-address check. So adoptedAssemblableQuorumMember() green-lights member B off member A's cache, and assembly silently substitutes A. In a threshold-2 quorum that means the second approval goes in as
  A again and the transaction never reaches threshold.

  Suggested guard — and note the secret-bearing branch 8 lines below already compares addresses, so this is consistent with the surrounding design, not a new concept:

  if (isApiSourcedServerSignerConfig(config)) {
      const cached = this.#resolvedRecoveryServerSigner;
      if (cached != null && cached.derivedAddress === config.address) {
          return cached;
      }
      throw new Error(...);
  }
  …plus the same predicate in canAutoAssemble. Longer term, key the cache by address instead of one slot.

packages/wallets/src/wallets/services/signer-manager.ts
alberto-crossmint, line 1: Colude found:

**stripSecretFromRecovery() wipes secrets from members that were never resolved**

packages/wallets/src/wallets/services/signer-manager.ts:152-170

  The quorum arm strips every server member that has an address. The comment justifies it with "once a server member's derivation is resolved (and cached in the resolver), its secret is no longer needed" — but the code never checks that it was resolved. The single-signer arm immediately below does check (resolvedRecoveryAddress != null).

  This matters because ServerSignerResolver has exactly one recovery cache slot (#resolvedRecoveryServerSigner), so at most one member can ever be "resolved."

  Reproduced. Quorum with server members A and B, both carrying secrets — exactly what part 3 (#1999) mergeQuorumRuntimeConfig produces via { ...method, ...member }:

  ✗ A: stripSecretFromRecovery() wipes the secret of a member that was never resolved
      expected undefined to be 'bbbb…'   ← B's secret, gone

  Mirror the single-signer arm:

  member.type === "server" &&
  "secret" in member &&
  member.address === this.#serverSignerResolver.resolvedRecoveryAddress

packages/wallets/src/signers/server/resolver.ts
panosinthezone: Confirmed and fixed in 936bd5e8 — both call sites are now per-address:

- `resolveDerivation()`'s API-sourced branch serves the cache only when `cached.derivedAddress === config.address`, otherwise it throws the existing actionable error (now naming the address).
- The server descriptor's `canAutoAssemble` compares `ctx.serverSigners.resolvedRecoveryAddress === config.address` instead of the global `hasRecoveryResolution`.

So an unresolved member now degrades to the honest "call wallet.useSigner() with the member you hold" path instead of silently substituting the cached member's key. Safe for single-admin flows: `matchesRecovery`/`matchQuorumMembers` both guarantee the selected derivation's address equals the recovery/member API address, so the cache always matches its own config there.

Added regressions: cached-A-refuses-B in `resolver.test.ts`, the wrong-address `canAutoAssemble` case in `descriptors.test.ts`, and an adopted-but-unresolved member returning null from `adoptedAssemblableQuorumMember()` in `signer-manager.test.ts`.

Agreed on keying the cache by address as the root-cause fix — deferring that to a follow-up (M4-7 candidate) since threshold >1 is API-gated at launch and the guard closes the substitution path.

packages/wallets/src/wallets/services/signer-manager.ts
panosinthezone: Confirmed and fixed in 936bd5e8 — the quorum arm now mirrors the single-signer arm exactly as you suggested: only the member whose `address === resolvedRecoveryAddress` is stripped, so a member the resolver never resolved keeps its grafted secret until it's selected (at which point the slot holds its derivation and it gets stripped in turn).

Updated the existing strip test to pin this: a quorum with a resolved member (`0xM`), an unresolved secret-bearing member (`0xOther`), and a no-API-address member — only `0xM` loses its secret.

packages/wallets/src/wallets/services/signer-manager.ts
```diff
@@ -84,7 +89,85 @@ export class SignerManager<C extends Chain> {
...
+            return;
+        }
+        this.#recovery = {
+            ...this.#recovery,
+            signers: this.#recovery.signers.map((member) =>
+                getQuorumMemberLocator(member) === memberLocator ? merged : member
+            ),
+        };
+        this.#adoptedQuorumMemberLocators.add(memberLocator);
+    }
```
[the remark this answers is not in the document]
panosinthezone: Taken in 936bd5e8: `adoptQuorumMemberConfig` now no-ops and logs `signerManager.adoptQuorumMemberConfig.noSuchMember` when the locator matches no member, so nothing is recorded on a miss. No current call site can hit it (callers derive the locator from the matched member itself), but the invariant is now explicit and observable. Test added.

packages/wallets/src/signers/server/resolver.ts
panosinthezone: @greptileai can you verify the fix?

packages/wallets/src/wallets/services/signer-manager.ts
panosinthezone: @greptileai can you verify the fix?

packages/wallets/src/wallets/services/signer-manager.ts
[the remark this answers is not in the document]
panosinthezone: @greptileai can you verify the fix?

PR #2001, opened 2026-07-30
Remove the stale device signer after a successful recovery

tomas-martins-crossmint:
This pull request adds a new step to the recovery flow to delete the stale device signer once a new one has been registered.

Removal only happens when we can be certain which signer is stale. If there's exactly one device signer registered, that one is stale by definition and gets removed. If there's more than one (another device already using this wallet, for example), we only remove it if exactly one of them matches this device's current name.

This pull request also bumps the CrossmintDeviceSigner module to 1.1.3, which includes a fix to report local keys that can't be used anymore as not present, which marks the wallet as needing recovery.

## Changes

- **`device-recovery-service.ts`**: `recover()` now removes the stale device signer after successfully registering its replacement, using the single-signer or name-match rule above
- **`CrossmintReactNativeUI.podspec`**: bumped the pinned `CrossmintDeviceSigner` version to 1.1.3


packages/wallets/src/wallets/wallet.ts
```diff
@@ -1038,6 +1040,19 @@ export class Wallet<C extends Chain> {
             configSigners.map(async (configSigner) => {
                 try {
                     const signerState = await this.#signerManager.getSignerState(configSigner.locator as SignerLocator);
+                    if (signerState.signer == null) {
+                        return null;
+                    }
+                    // The per-signer GET response doesn't include `name`, but the wallet-level
```
albertoelias-crossmint: how come the get response includes something different to the get wallet?

tomas-martins-crossmint: ah, that's a leftover from claude while I was testing the recovery flow and the app wasn't fetching the signers name at first, but never managed to reproduce it again

tomas-martins-crossmint: I'll remove this for now

packages/wallets/src/wallets/services/device-recovery-service.ts
```diff
@@ -161,6 +164,13 @@ export class DeviceRecoveryService<C extends Chain> {
             }
         }
 
+        const existingSigners = await this.#signers();
+        const existingDeviceSigners = existingSigners.filter((s) => s.type === "device");
+        const staleDeviceSignerLocator = this.#resolveStaleDeviceSignerLocator(
+            existingDeviceSigners,
+            deviceSignerKeyStorage
+        );
```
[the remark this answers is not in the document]
albertoelias-crossmint: @tomas-martins-crossmint worth looking into properly. dont assume that what greptile says is always true, but if it is, we should see if we can fix without making the code messy

albertoelias-crossmint (approved):

albertoelias-crossmint (approved):

PR #2005, opened 2026-08-05
feat(checkout): add identityVerificationHandling so a merchant can own the verification step

AngelPaella:
The SDK half of the merchant escape hatch for KYC, §4.5 of the Identity SDK PRD. ENG4-367.

The checkout page half is [Paella-Labs/crossbit-main#28380](https://github.com/Paella-Labs/crossbit-main/pull/28380), which has to reach production before this ships. Unknown query params are stripped, so a newer SDK against an older page loses the flag and both checkout and the merchant render a widget for the same inquiry.

## What this adds

**`identityVerificationHandling?: "external"`** on `CrossmintEmbeddedCheckoutV3WebProps`, which only the react-ui component takes. It rides to the checkout page as a query param through `appendObjectToQueryParams`, which appends strings bare, so the page receives `identityVerificationHandling=external` and its zod enum accepts it. It intersects the shared props type, so both the existing-order and new-order shapes keep it.

Web only, deliberately. It started on the props type react-native shares with web, which meant an RN merchant could set the flag, typecheck, reach the page, and have checkout suppress the step with nothing able to replace it: `CrossmintIdentityVerification` renders a DOM iframe, and RN's barrel exports neither it nor the credentials reader. The order would sit at `requires-kyc` behind a collapsed WebView with no message. RN now rejects the flag at compile time, and gets it back when there is an RN component to mount.

**`getIdentityVerificationCredentials(order)`**, taking `Order | undefined` and returning `IdentityVerificationCredentials | undefined`. It has to accept undefined: `useCrossmintCheckout()` returns `order?: Order`, so a narrower parameter would have forced every merchant into `order!`, discarding exactly the defensiveness the reader exists to provide. A merchant taking the step over reads the order from the hook and passes the result to `CrossmintIdentityVerification`:

```tsx
const { order } = useCrossmintCheckout();
const credentials = getIdentityVerificationCredentials(order);

<CrossmintEmbeddedCheckout orderId={orderId} identityVerificationHandling="external" payment={...} />
{credentials && <CrossmintIdentityVerification credentials={credentials} onComplete={...} />}
```

Re-exported from `@crossmint/client-sdk-react-ui` so merchants do not have to depend on `client-sdk-base` directly.

## Why one value and not two

The prop started as `kycHandling?: "checkout" | "external"`. Two things changed.

The name is now `identityVerificationHandling`, matching the vocabulary the rest of this surface already uses (`CrossmintIdentityVerification`, `getIdentityVerificationCredentials`). KYC is the provider's term for one step of identity verification, and it was the odd one out. This diverges from §4.5 of the PRD, which writes `kycHandling`.

The union is a single value. Nothing branches on the non-external value on the checkout page, where the only read is `=== "external"`. Widening an optional union later is backward compatible and free, so a second value buys nothing today and commits us to a name before the third mode's real semantics are known. If the in-house UI never ships, we never paid for it. `"embedded"` was the leading candidate and is actively misleading besides, since the merchant's replacement is an iframe too: ownership is the axis, not where it renders.

Consequence for the page: `identityVerificationHandling="embedded"` is now rejected rather than ignored, so it hits checkout's validation error screen like any unknown value. That is the existing loud-rejection behaviour, and it means a second mode has to reach production on the page before it can ship here.

## A transport fix rides along

Running this flow in a browser turned up a bug in `WindowTransport`, and the fix ships here because `identityVerificationHandling="external"` is what exposes it.

`addMessageListener` subscribes to the global `message` event and filtered on `event.origin` alone. One Crossmint iframe per page kept that harmless, since no other frame could send from that origin. This flag puts two on the page: the checkout iframe, and the `CrossmintIdentityVerification` iframe the merchant mounts. Both serve from the same origin, and `ui:height.changed` is the one event name their maps share, so each client received the other's heights.

The result contradicted what this PR and #28380 both claim. Checkout suppressed its step and reported `height: 0`, the verification iframe reported 660, and the checkout iframe took that, so the merchant got 660px of blank space where the box should have collapsed. The failure flips with message order: checkout's 0 arriving last sets the merchant's Persona form to 0, and the buyer cannot verify at all. An early harness run recorded that version.

Captured in the parent with a listener that tags each message with its sending frame:

```
height= 146  from=crossmint-embedded-checkout.iframe
height=   0  from=crossmint-embedded-checkout.iframe      <- checkout collapses
height= 660  from=crossmint-identity-verification.iframe  <- and this overwrites it
```

`WindowTransport` now compares `event.source` to the peer window it already holds for sending. `SignersWindowTransport` does this today. `IFrameWindow`, `PopupWindow` and `NewTabWindow` construct the transport with the window they talk to, so the check holds for them. A message whose sending window has closed carries a null source and drops.

Worth a reviewer's attention: this narrows a trust boundary that has been loose since the transport was written, and it affects every product using these iframes, not only checkout.

## The one caller that pointed at the wrong window

`useOAuthWindowListener` built a `ChildWindow` over `window.opener || window.parent`. On a merchant's top-level page `window.opener` is null and `window.parent` is that same page, so the peer was the page's own window, while `authMaterialFromPopupCallback` arrives from the OAuth popup. Those cannot match, and social login would have hung with the popup open.

That client only listened, never sent, and its `targetOrigin` was `"*"`, so nothing read the peer until this change put it on the receive path. Callers passing `"*"` filtered nothing at all before, which is the second gap this closes: any frame on the merchant's page could post an `authMaterialFromPopupCallback` carrying a secret it chose, and the page handed it to `handleRefreshAuthMaterial`.

Both listeners now attach to the popup the same function opens, which is the window that sends them. That also clears a leak older than this PR: `off("authMaterialFromPopupCallback")` passes an event name to an API that takes a listener id, so each login attempt stacked another live handler that was never removed.

The changeset bumps `client-sdk-react-ui` for this half.

## Why a reader instead of typing the Order

`packages/client/base/src/lib/hosted-checkout/Order.ts` says at the top that it is kept in sync with `@crossmint/products-payments-headless-checkout` by hand, and it has drifted: the backend models the kyc preparation variant, this mirror has no `kyc` key anywhere in `preparation`. Editing a union inside 5000 lines of declared types is a bigger change than this feature and is not KYC-specific, so the reader holds the one cast inside the SDK instead of leaving it in every integration. Typing the mirror properly is filed separately, and it will simplify the reader's body without touching its signature.

The reader also optional-chains `order.payment`, because an order reaches the merchant over postMessage where `order` is typed `z.any()` on both ends. A shape the type forbids should return undefined rather than throw in the merchant's app.

## Test plan

- `get-identity-verification-credentials.test.ts`, new, 5 cases: returns the credentials, and returns undefined for a non-verification preparation, for no preparation, for an order with no payment at all, and for no order yet. All pass.
- `window-transport.test.ts`, new, 4 cases: a same-origin non-peer frame gets dropped, a null source gets dropped, a foreign origin still gets dropped, and `removeMessageListener` unsubscribes. Take the source check back out and the first two fail. This package had no tests before, so it also gains a `test:vitest` script and a `vitest.config.ts` copied from `client-sdk-base`.
- `use-oauth-window-listener.test.ts`, new, 1 case: a callback from a stranger window is ignored, the same callback from the popup reaches `handleRefreshAuthMaterial`. Restore the old `ChildWindow` and it fails with zero calls.
- `pnpm turbo build` on `client-sdk-base`, `client-sdk-react-ui` and `client-sdk-react-native-ui` passes, which is what typechecks this repo. In the built output I checked the three things that matter: base's declarations export `getIdentityVerificationCredentials` and `CrossmintEmbeddedCheckoutV3WebProps` carrying `identityVerificationHandling?: "external"`, react-ui's two embed components declare the web props type, and react-native's `dist/index.d.ts` contains no mention of the flag.
- `pnpm lint` is clean on the touched files. The three warnings biome reports on the props file are pre-existing `Record<string, any>` line items.

Run in a browser, against `~/dev/kyc-e2e-harness` §F: the quickstart on :3006 with `identityVerificationHandling="external"`, crossbit-main #28380 on :3000, driven by Playwright. Checkout's iframe measures 0 and its body is empty, the merchant's slot mounts a real Persona sandbox form at 660 as a sibling rather than a nested frame, `kyc:ready` fires, checkout polls `GET /orders` on its 3s cadence, and the verification iframe's `src` stays byte-identical across those ticks so the fresh-per-poll credentials object does not remount Persona.

Every other consumer of the transport was driven in a browser against the same local stack. Embedded checkout with the flag off reaches 660px with no identity iframe on the page. Payment method management in `apps/payments/nextjs` receives `ui:height.changed` at 214, 334 and 392, each with `event.source === iframe.contentWindow`. OAuth login ran twice in one page load: the second popup closed itself and one callback reached the page. An abandoned popup left no session and re-armed the button, which only the `checkWindowClosure` interval can do, and that block calls `stopListening()`.

Reloading an iframe in place kept `contentWindow` identical and the client kept receiving, so navigating an iframe does not mint a new peer. Re-parenting the element would, and nothing here re-parents.

Three consumers need no run. Hosted checkout popup and new tab register no listener: `PopupWindow.initSync` and `NewTabWindow.initSync` get no `incomingEvents`, and the overlay polls `windowClient.window.closed`. React Native uses `RNWebViewTransport`, a separate class. The signer iframes compare `event.source` today through `SignersWindowTransport`.

`getUrl()` on the built dist emits `?identityVerificationHandling=external`, omits the param when the prop is unset, and never emits the old `kycHandling` name. The receiving end has a schema test in #28380 that parses a bare `identityVerificationHandling=external`, plus a browser check that the page rejects any other value with its validation screen.

Not in this PR: docs for the takeover flow, and react-native support, which needs an identity verification component that is not an iframe before the flag can mean anything there.

The changeset states the deploy-order requirement too, since the changelog is where a merchant reads it rather than this description.

### DIFF
--- a/packages/client/base/src/services/identity-verification/get-identity-verification-credentials.test.ts
+++ b/packages/client/base/src/services/identity-verification/get-identity-verification-credentials.test.ts
@@ -8,6 +8,8 @@ function order(preparation: unknown): Order {
 }
 
 describe("getIdentityVerificationCredentials", () => {
+    // environmentId is on the wire but off the type on purpose: the backend hardcodes it to null and
+    // PAY-11803 deletes it. The fixture keeps it to pin that the reader passes the object through as-is.
     test("returns the credentials an order carries for its verification step", () => {
         const credentials = { provider: "persona", inquiryId: "inq-1", sessionToken: "tok-1", environmentId: null };
 
--- a/packages/client/base/src/services/identity-verification/get-identity-verification-credentials.test.ts
+++ b/packages/client/base/src/services/identity-verification/get-identity-verification-credentials.test.ts
@@ -27,4 +27,9 @@ describe("getIdentityVerificationCredentials", () => {
     test("returns undefined for an order carrying no payment at all", () => {
         expect(getIdentityVerificationCredentials({ orderId: "order-1" } as Order)).toBeUndefined();
     });
+
+    // useCrossmintCheckout() hands back `order?: Order`, so the documented call site passes this.
+    test("returns undefined before the checkout hook has an order", () => {
+        expect(getIdentityVerificationCredentials(undefined)).toBeUndefined();
+    });
 });
--- a/packages/client/base/src/services/identity-verification/get-identity-verification-credentials.test.ts
+++ b/packages/client/base/src/services/identity-verification/get-identity-verification-credentials.test.ts
@@ -1,0 +1,30 @@
+import { describe, expect, test } from "vitest";
+
+import type { Order } from "@/lib/hosted-checkout/Order";
+import { getIdentityVerificationCredentials } from "./getIdentityVerificationCredentials";
+
+function order(preparation: unknown): Order {
+    return { orderId: "order-1", payment: { status: "requires-kyc", preparation } } as Order;
+}
+
+describe("getIdentityVerificationCredentials", () => {
+    test("returns the credentials an order carries for its verification step", () => {
+        const credentials = { provider: "persona", inquiryId: "inq-1", sessionToken: "tok-1", environmentId: null };
+
+        expect(getIdentityVerificationCredentials(order({ kyc: credentials }))).toEqual(credentials);
+    });
+
+    // A crypto order's preparation shape, which is what the type actually models today.
+    test("returns undefined when the preparation is not a verification one", () => {
+        expect(getIdentityVerificationCredentials(order({ chain: "base", payerAddress: "0x1" }))).toBeUndefined();
+    });
+
+    test("returns undefined before the order needs verification", () => {
+        expect(getIdentityVerificationCredentials(order(undefined))).toBeUndefined();
+    });
+
+    // Orders reach a merchant over postMessage, so a shape the type forbids still has to not throw.
+    test("returns undefined for an order carrying no payment at all", () => {
+        expect(getIdentityVerificationCredentials({ orderId: "order-1" } as Order)).toBeUndefined();
+    });
+});
--- a/packages/client/base/src/types/embed/v3/CrossmintEmbeddedCheckoutV3Props.ts
+++ b/packages/client/base/src/types/embed/v3/CrossmintEmbeddedCheckoutV3Props.ts
@@ -7,6 +7,12 @@ interface CrossmintEmbeddedCheckoutV3CommonProps {
     appearance?: EmbeddedCheckoutV3Appearance;
     payment: EmbeddedCheckoutV3Payment;
     jwt?: string;
+    /**
+     * Who renders the identity verification step. `"external"` stops checkout from rendering it, so
+     * you have to mount `CrossmintIdentityVerification` yourself with the credentials from
+     * `getIdentityVerificationCredentials(order)`, or the buyer cannot finish the order.
+     */
+    kycHandling?: "checkout" | "external";
 }
 
 export interface CrossmintEmbeddedCheckoutV3ExistingOrderProps extends CrossmintEmbeddedCheckoutV3CommonProps {
--- a/packages/client/base/src/types/embed/v3/CrossmintEmbeddedCheckoutV3Props.ts
+++ b/packages/client/base/src/types/embed/v3/CrossmintEmbeddedCheckoutV3Props.ts
@@ -34,7 +34,7 @@ export type CrossmintEmbeddedCheckoutV3Props =
 export type CrossmintEmbeddedCheckoutV3WebProps = CrossmintEmbeddedCheckoutV3Props & {
     /** `"external"`: you mount `CrossmintIdentityVerification` from
      * `getIdentityVerificationCredentials(order)`, or the buyer cannot finish. */
-    kycHandling?: "checkout" | "external";
+    identityVerificationHandling?: "external";
 };
 
 export type EmbeddedCheckoutV3Recipient = EmbeddedCheckoutV3EmailRecipient | EmbeddedCheckoutV3WalletAddressRecipient;
--- a/packages/client/base/src/types/embed/v3/CrossmintEmbeddedCheckoutV3Props.ts
+++ b/packages/client/base/src/types/embed/v3/CrossmintEmbeddedCheckoutV3Props.ts
@@ -7,12 +7,6 @@ interface CrossmintEmbeddedCheckoutV3CommonProps {
     appearance?: EmbeddedCheckoutV3Appearance;
     payment: EmbeddedCheckoutV3Payment;
     jwt?: string;
-    /**
-     * Who renders the identity verification step. `"external"` stops checkout from rendering it, so
-     * you have to mount `CrossmintIdentityVerification` yourself with the credentials from
-     * `getIdentityVerificationCredentials(order)`, or the buyer cannot finish the order.
-     */
-    kycHandling?: "checkout" | "external";
 }
 
 export interface CrossmintEmbeddedCheckoutV3ExistingOrderProps extends CrossmintEmbeddedCheckoutV3CommonProps {
@@ -34,6 +28,14 @@ export interface CrossmintEmbeddedCheckoutV3NewOrderProps extends CrossmintEmbed
 export type CrossmintEmbeddedCheckoutV3Props =
     | CrossmintEmbeddedCheckoutV3ExistingOrderProps
     | CrossmintEmbeddedCheckoutV3NewOrderProps;
+
+// Web only: taking over the verification step needs CrossmintIdentityVerification, which renders an
+// iframe. React Native has no equivalent yet, so the flag stays off the shared props.
+export type CrossmintEmbeddedCheckoutV3WebProps = CrossmintEmbeddedCheckoutV3Props & {
+    /** `"external"`: you mount `CrossmintIdentityVerification` from
+     * `getIdentityVerificationCredentials(order)`, or the buyer cannot finish. */
+    kycHandling?: "checkout" | "external";
+};
 
 export type EmbeddedCheckoutV3Recipient = EmbeddedCheckoutV3EmailRecipient | EmbeddedCheckoutV3WalletAddressRecipient;
 
--- a/packages/client/base/src/types/embed/v3/CrossmintEmbeddedCheckoutV3Props.ts
+++ b/packages/client/base/src/types/embed/v3/CrossmintEmbeddedCheckoutV3Props.ts
@@ -29,8 +29,8 @@ export type CrossmintEmbeddedCheckoutV3Props =
     | CrossmintEmbeddedCheckoutV3ExistingOrderProps
     | CrossmintEmbeddedCheckoutV3NewOrderProps;
 
-// Web only: taking over the verification step needs CrossmintIdentityVerification, which renders an
-// iframe. React Native has no equivalent yet, so the flag stays off the shared props.
+// Web only: the takeover needs CrossmintIdentityVerification, a DOM iframe with no react-native
+// counterpart. RN checkout keeps running the verification step itself.
 export type CrossmintEmbeddedCheckoutV3WebProps = CrossmintEmbeddedCheckoutV3Props & {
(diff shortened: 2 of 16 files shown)


packages/client/base/src/services/identity-verification/get-identity-verification-credentials.test.ts
```diff
@@ -0,0 +1,30 @@
...
+import { getIdentityVerificationCredentials } from "./getIdentityVerificationCredentials";
+
+function order(preparation: unknown): Order {
+    return { orderId: "order-1", payment: { status: "requires-kyc", preparation } } as Order;
+}
+
+describe("getIdentityVerificationCredentials", () => {
+    test("returns the credentials an order carries for its verification step", () => {
+        const credentials = { provider: "persona", inquiryId: "inq-1", sessionToken: "tok-1", environmentId: null };
+
```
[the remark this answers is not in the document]
AngelPaella: Real divergence, and I am keeping it, with the reasoning now in the file.

`environmentId` is genuinely on the wire. `kycRequiredPaymentPreparationSchema` declares it as `z.string().nullable()`, and `buildOrderPayment.ts:320` hardcodes it to `null` under `TODO(PAY-11803): delete environmentId in next API version`.

So adding it to `IdentityVerificationCredentials` would put a field on our public type that is always null, that nothing in the SDK reads, and that the backend is already removing. `CrossmintIdentityVerification` does not want it either: the standalone route takes `provider` and `inquiryId`, and Persona resolves template and environment from the inquiry itself.

The fixture keeps the field on purpose. The reader returns `preparation.kyc` by reference, so whatever the order carries rides along, and asserting `toEqual` against the full wire shape pins that we hand the object back unreshaped rather than picking fields off it. Added a comment saying so, since the divergence reads as an oversight otherwise: 81af5c6.

A merchant who does need it today is reading a field we do not promise, which is the correct signal while PAY-11803 is open.

packages/client/base/src/types/embed/v3/CrossmintEmbeddedCheckoutV3Props.ts
```diff
@@ -29,6 +29,14 @@ export type CrossmintEmbeddedCheckoutV3Props =
     | CrossmintEmbeddedCheckoutV3ExistingOrderProps
     | CrossmintEmbeddedCheckoutV3NewOrderProps;
 
+// Web only: taking over the verification step needs CrossmintIdentityVerification, which renders an
+// iframe. React Native has no equivalent yet, so the flag stays off the shared props.
+export type CrossmintEmbeddedCheckoutV3WebProps = CrossmintEmbeddedCheckoutV3Props & {
+    /** `"external"`: you mount `CrossmintIdentityVerification` from
+     * `getIdentityVerificationCredentials(order)`, or the buyer cannot finish. */
+    identityVerificationHandling?: "external";
+};
```
vmc-crossmint: What does it mean, we our sdk cannot work in react native?

vmc-crossmint (approved):
Approved for me, but i prefer @maxwellfortney to review too as he has context on sdk details

packages/client/base/src/types/embed/v3/CrossmintEmbeddedCheckoutV3Props.ts
AngelPaella: RN checkout keeps working, and this PR leaves it alone. Verification still runs inside checkout's WebView there.

Only the opt-out is web. `identityVerificationHandling="external"` tells the checkout page to skip its verification step because the merchant renders it instead. The one component for that job, `CrossmintIdentityVerification`, is a DOM iframe, and RN has no counterpart to it yet.

Put the flag on the shared props and an RN merchant's code compiles clean all the way to the page. Checkout suppresses the step and nothing replaces it. The order sits at `requires-kyc` behind a collapsed WebView, no message shown. Keeping the flag off the shared type makes that a compile error instead. RN gets it back when there is a WebView verification component to mount.

AngelPaella:
On `event.source !== this.otherWindow`, since that is the line worth being careful about.

## Why source rather than origin

`postMessage` delivers to a window, not to a connection. The browser queues the message on the
target window and fires one `message` event, and every listener on that window sees it whatever
frame sent it. A receiver has three fields to filter on: `origin`, `source`, `data`.

`origin` names the site that sent the message. `source` names the window. We filtered on the site
while `send` had picked one window:

    send(...) { this.otherWindow.postMessage(message, this.targetOrigin); }

Receiving accepted anything from the peer's whole site, a wider set than the single peer we send
to. `SignersWindowTransport` overrides `addMessageListener` to compare source. This moves the
check into the base class.

With one Crossmint iframe per page, origin stood in for identity and nothing broke.
`identityVerificationHandling="external"` puts two on the page, both on `crossmint.com`, sharing
`ui:height.changed`:

    height= 146  from=crossmint-embedded-checkout.iframe
    height=   0  from=crossmint-embedded-checkout.iframe      <- checkout collapses
    height= 660  from=crossmint-identity-verification.iframe  <- and this overwrites it

Both frames carry the same origin, so no origin filter tells them apart. Callers passing
`targetOrigin: "*"` filtered nothing at all, which this closes too.

## What it touches

| Consumer | Transport | Reached |
|---|---|---|
| Embedded checkout iframe | `WindowTransport` | yes |
| Identity verification iframe | `WindowTransport` | yes |
| Payment method management iframe | `WindowTransport` | yes |
| OAuth login popup | `WindowTransport` | yes, and it pointed at the wrong window |
| NCS signer, export private key | `SignersWindowTransport` | no, compares source today |
| Hosted checkout popup and new tab | `WindowTransport` | no, registers no listener; the overlay polls `window.closed` |
| React Native WebViews | `RNWebViewTransport` | no, separate class |
| Order intent, agentic enrollment | Basis Theory | no, not our transport |

## OAuth login broke, and this PR fixes it

`useOAuthWindowListener` built its client over `window.opener || window.parent`. On a merchant's
top-level page `window.opener` is null and `window.parent` is that same page, so the peer was the
page's own window while `authMaterialFromPopupCallback` arrives from the popup. Those cannot
match, and social login would hang with the popup open.

The old code worked because `targetOrigin` was `"*"` and nothing read the peer when receiving.
That client only listened, so the mismatch stayed out of sight until this change put the peer on
the receive path.

Both listeners now attach to the popup that the same function opens. That clears a leak older
than this PR: `off("authMaterialFromPopupCallback")` passes an event name to an API that takes a
listener id, so each login attempt stacked another live handler.

## What I ran

Local crossbit-main on :3000, SDK dist from this branch, the KYC harness quickstart on :3006, and
`apps/payments/nextjs` on :3005.

| Check | Result |
|---|---|
| Embedded checkout, flag off | **pass**. `checkout iframe 660px`, `src flag none`, `identity iframe absent`, Persona form inside checkout |
| Identity verification, flag on | **pass**. checkout 0px, Persona at 660 as a sibling, `kyc:ready`, `src` byte-identical across the 3s polls |
| Payment method management | **pass on height**. `ui:height.changed` at 214, 334, 392, 392, each with `event.source === iframe.contentWindow`. `payment-method:selected` untested, the account has no saved card |
| OAuth login, twice in one page load | **pass**. Second login completed, popup closed itself, one callback reached the page |
| OAuth popup abandoned | **pass on recovery**. No session created, button re-armed. Latency unmeasured: Chrome throttled my poller while the tab sat behind the popup |

On the abandoned popup: no auth callback and no error callback arrived, and those are the only
other paths that clear `activeOAuthProvider`. The `checkWindowClosure` interval cleared it, and
that block calls `stopListening()`.

I also reloaded an iframe in place. `contentWindow` stayed the identical object and the client
kept receiving, so navigating an iframe does not mint a new peer. Re-parenting the element would,
and nothing here re-parents.

## Automated

`window-transport.test.ts`: a same-origin non-peer frame gets dropped, a null source gets dropped,
a foreign origin still gets dropped, `removeMessageListener` unsubscribes. Take the source check
out and the first two fail.

`use-oauth-window-listener.test.ts`: a callback from a stranger window goes nowhere, the same
callback from the popup reaches `handleRefreshAuthMaterial`. Restore the old `ChildWindow` and it
fails with zero calls.

PR #2019, opened 2026-08-11
feat(react-native): identity verification component and takeover flag

AngelPaella:
Ships `CrossmintIdentityVerification` for React Native, plus the `identityVerificationHandling="external"` takeover flag, so an RN merchant can render Crossmint's hosted verification step in their own screen instead of inside embedded checkout.

The hosted page at `/sdk/unstable/identity-verification` already detects `window.ReactNativeWebView` and speaks the RN bridge protocol, so this is client-only: no crossbit-main change, no new wire contract, no new dependency.

## What's here

**`CrossmintIdentityVerification`** hosts that page in `RNWebView`, builds its URL through the existing `createIdentityVerificationService(...).iframe.getUrl(props)`, and subscribes the five shipped events (`ui:height.changed`, `kyc:ready`, `kyc:completed`, `kyc:cancelled`, `kyc:error`) through `WebViewParent`, surfacing them as the same props the web component takes.

Details worth review:

- A failed load reports through `onError`. Without it a 500 or an offline device produces no `kyc:ready` and no `kyc:error`, and the view stays 0px tall, so the merchant cannot distinguish a dead page from a loading one. Both `onError` and `onHttpError` map to the shipped contract's non-retriable `widget-unavailable`. No main-frame guard is needed, unlike Flutter's equivalent: `RNCWebViewClient` overrides only the legacy 4-arg `onReceivedError` (main-frame only by Android contract, never the modern `WebResourceRequest` overload) and wraps `onReceivedHttpError` in `if (request.isForMainFrame())`, so a Persona sub-frame hiccup cannot reach these. `retriable: false` matches Flutter and is honest: the component does not retry on its own.
- Listener cleanup uses the ids `client.on()` returns, not the event name. `off(name)` removes nothing, which is the leak `EmbeddedCheckoutV3WebView.tsx` has today. The unmount test covers it.
- Callbacks are read off a ref, so listeners subscribe once and a late event still calls the current render's props. This is the component's subtlest mechanism and it now has a dedicated test: removing the ref leaves every other test green, because they each render once.
- `domStorageEnabled` is set explicitly. Android's `WebSettings` default is off and react-native-webview applies no default of its own: the codegen spec declares `domStorageEnabled?: boolean` with no `WithDefault`, and the JS component has no `defaultProps`, so an unset prop means the native setter is never called. (The prop's own JSDoc claiming "The default value is `true`" is stale.) The hosted Persona widget touching `localStorage` would then fail on Android only. Checkout's WebView and `ExportPrivateKeyButton` both set it for the same reason. Not copying the rest of checkout's Android block: `mixedContentMode="always"`, `allowFileAccess` and `allowUniversalAccessFromFileURLs` are security loosenings with no identified need here.
- `WebViewParent` receives `webViewRef` itself rather than a `{ current: node }` snapshot, so the transport reads the live node. Inert today (`identityVerificationOutgoingEvents` is `{}` so nothing sends, and no `recovery` is configured so nothing reloads), but it is also less code than the pattern it was copied from.

`height` starting at 0 is deliberate, matching both in-repo precedents: the hosted page reports ~660 off a fixed-pixel Persona container regardless of the WebView's own height, and a 0-height RN WebView still loads and runs JS.

**`getIdentityVerificationCredentials`** is re-exported from `client-sdk-base`. It is platform-neutral and already tested there, so this is a re-export rather than an implementation, and the build is what proves the symbol exists.

**`identityVerificationHandling`** moves from `CrossmintEmbeddedCheckoutV3WebProps` onto the shared common props. RN builds its checkout URL with the same `getUrl(props)` and `appendObjectToQueryParams`, so moving the field is the entire change: no forwarding code. `CrossmintEmbeddedCheckoutV3WebProps` stays as a deprecated alias because deleting an exported type is breaking.

## Test runner

The package had a `vitest.config.ts` and a `turbo.json` declaring `test:vitest`, but no script, no test deps and no test files, so CI ran nothing here. Added the script and deps; this is the package's first suite, 10 tests.

Two test-environment notes:

- `react-native-get-random-values` is stubbed via a `vitest.config.ts` alias rather than `vi.mock`. rn-window's transport imports that polyfill, which requires Flow-typed react-native and is unparseable by vitest, and rn-ui and rn-window resolve **different installed copies** (2.0.0 vs 1.11.0), so a `vi.mock` in the test file never intercepts the copy rn-window's dist actually imports. The alias works regardless of importer and covers future tests in this package.
- The `react-native-webview` mock forwards `ref`. React 19 does not assign refs to function components, and without it `webViewRef.current` stays null, the client is never constructed, and every event assertion passes vacuously.

## Known and accepted

An event arriving before the post-mount effect constructs the client is dropped, since `onMessage` forwards through `client?.handleMessage`. The window is one render against a full page load, and checkout has the same shape. `onMessage` itself stays a defined function across that window on purpose: react-native-webview derives `messagingEnabled` from `typeof onMessage === "function"`, so passing `client?.handleMessage` directly would create the native view with messaging off and no `window.ReactNativeWebView` for the hosted page to detect.

## Not here

Camera permission *code* (`mediaCapturePermissionGrantType` on iOS, a `CAMERA` declaration in `withCrossmintUI`) is gated on Leg 1 of the device-test protocol. The *documentation* is not: `RNCWebChromeClient.onPermissionRequest` maps `RESOURCE_VIDEO_CAPTURE` to `Manifest.permission.CAMERA` and calls `Activity.requestPermissions`, which cannot succeed unless the host app declares it, and that holds regardless of what Leg 1 finds. So the reference snippet now carries a note naming both the Android manifest entry and `NSCameraUsageDescription`. Run A may show `react-native-webview@13.15.0` already grants capture, in which case the deliverable is a docs note. If it shows Android denies it, the fix belongs in `EmbeddedCheckoutV3WebView.tsx` too, since the shipped in-checkout KYC step has the same gap. Writing it before the test means guessing.

Manual end-to-end verification of the flag is also not possible yet: no RN app with embedded checkout exists (the repo's only Expo app is wallets-only, and `embedded-checkout-playground` is Next.js).

## Release note

The flag is a footgun while the component is unpublished: a merchant setting `"external"` with nothing to mount leaves the buyer unable to finish. Both changesets are in this PR so the component and the flag go out in the same release.

Separately, published `client-sdk-base@2.7.0` still builds the dead `/sdk/unstable/kyc-verification` route. The fix is already on main; it is waiting in CI's release PR #2013.


packages/client/ui/react-native/src/components/identity-verification/CrossmintIdentityVerification.tsx
```diff
@@ -0,0 +1,85 @@
...
+            ref={webViewRef}
+            source={{ uri: identityVerificationService.iframe.getUrl(props) }}
+            // Stays a function before the client exists: react-native-webview derives
+            // messagingEnabled from `typeof onMessage === "function"`, and without it the
+            // page has no window.ReactNativeWebView to detect.
+            onMessage={(event) => client?.handleMessage(event)}
+            style={{ width: "100%", height, backgroundColor: "transparent" }}
+            allowsInlineMediaPlayback={true}
+            mediaPlaybackRequiresUserAction={false}
+            domStorageEnabled={true}
```
luans-qa, line 85: A load failure is silent here. If the hosted page 500s or the device is offline there's no kyc:ready and no kyc:error, and height stays 0, so the merchant renders an invisible view forever with no callback and no way to tell it from "still loading". identityVerificationIncomingEvents["kyc:error"] already has reason: "widget-unavailable" and retriable — wiring onError/onHttpError into latestProps.current.onError uses the shipped contract and is a few lines.

packages/client/ui/react-native/src/components/identity-verification/CrossmintIdentityVerification.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const identityVerificationService = createIdentityVerificationService({ apiClient });
+
+    useEffect(() => {
+        const webView = webViewRef.current;
+        if (!webView || client) {
+            return;
+        }
+        setClient(
+            new WebViewParent(
+                { current: webView },
```
luans-qa: WebViewParent's constructor takes RefObject<WebView | null> — pass webViewRef directly. { current: webView } is a copy of one render's node, so transport.reload() and every send would target a stale instance if the node is recreated. Nothing exercises it today (no recovery configured), but it's copied from EmbeddedCheckoutV3WebView and there's no reason to keep it.

packages/client/ui/react-native/scripts/generate-reference.mjs
```diff
@@ -82,10 +82,13 @@ see the [previous version of this page](/sdk-reference/wallets/v0/react-native/{
             "CrossmintCheckoutProvider",
             "useCrossmintCheckout",
             "CrossmintEmbeddedCheckout",
+            "CrossmintIdentityVerification",
```
luans-qa, line 86: getIdentityVerificationCredentials is exported from the package and used by the CrossmintIdentityVerification snippet in examples.json, but isn't in this list — the generated page will show an undocumented symbol. Add it here plus a descriptions entry

packages/client/ui/react-native/examples.json
```diff
@@ -76,5 +76,9 @@
...
         "language": "tsx",
         "note": "In React Native, when payment.crypto.enabled is true, you must provide a payment.crypto.payer object with the required callback handlers for signing transactions and switching chains."
+    },
+    "CrossmintIdentityVerification": {
+        "code": "import {\n    CrossmintIdentityVerification,\n    getIdentityVerificationCredentials,\n    useCrossmintCheckout,\n} from \"@crossmint/client-sdk-react-native-ui\";\n\nfunction VerificationStep() {\n    const { order } = useCrossmintCheckout();\n    const credentials = getIdentityVerificationCredentials(order);\n\n    if (credentials == null) {\n        return null;\n    }\n\n    return (\n        <CrossmintIdentityVerification\n            credentials={credentials}\n            onComplete={({ status }) => console.log(status)}\n        />\n    );\n}",
+        "language": "tsx"
```
luans-qa: Worth a note like the CrossmintEmbeddedCheckout entry above has. On Android RNCWebChromeClient.onPermissionRequest maps RESOURCE_VIDEO_CAPTURE → Manifest.permission.CAMERA and requests it at runtime, so it can only succeed if the host app declares CAMERA; iOS needs NSCameraUsageDescription. Deferring the code to the device-test protocol is fine, but a document-capture component with no camera setup in the reference docs is a support ticket. withCrossmintUI already injects GooglePay/device-signer native config, so it's the natural home.

packages/client/ui/react-native/src/components/identity-verification/crossmint-identity-verification.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+        // `args` is what the component forwards: the parsed payload, or nothing for the
+        // two events whose schema is empty.
+        test.each(LIFECYCLE)("forwards $event to $prop", ({ prop, event, data, args }) => {
+            const callback = vi.fn();
+            render(<CrossmintIdentityVerification credentials={CREDENTIALS} {...{ [prop]: callback }} />);
+
+            emit(event, data);
+
+            expect(callback).toHaveBeenCalledWith(...args);
+        });
```
luans-qa, line 91: The web suite pins the ref with "calls the callback from the latest render, not the one captured at subscribe time" (rerender with a fresh onComplete, assert stale was never called). The RN component ports the ref and its comment verbatim but nothing here covers it, it's the component's subtlest mechanism and the only one untested.

packages/client/ui/react-native/src/components/identity-verification/CrossmintIdentityVerification.tsx
AngelPaella: Fixed in 7f599db. Both `onError` and `onHttpError` now report non-retriable `widget-unavailable` through the shipped contract.

RN needs no main-frame guard, unlike the Flutter port of this path. `RNCWebViewClient` overrides only the legacy 4-arg `onReceivedError`, which Android fires for the main frame alone, and never the modern `WebResourceRequest` overload. It wraps `onReceivedHttpError` in `if (request.isForMainFrame())` too. So a failure inside Persona's iframe never reaches these handlers, and `WebViewError` carries no `isForMainFrame` field to guard on anyway.

Both use `retriable: false`, matching Flutter. The component never retries, so `true` would mislead the merchant.

Two new tests cover this. I deleted both handlers to check them: 2 failures, both new.

packages/client/ui/react-native/src/components/identity-verification/CrossmintIdentityVerification.tsx
AngelPaella: Fixed in 7f599db, which drops the local and two lines with it.

It is even more inert than you describe. `identityVerificationOutgoingEvents` is `{}`, so nothing ever sends (`RNWebViewTransport.ts:46,59`), and the component configures no `recovery`, so `reload` (`:129`) never fires. Neither stale-node path is reachable. I took the change for the smaller diff.

packages/client/ui/react-native/examples.json
AngelPaella: Added in 7f599db, naming `android.permission.CAMERA` in the Android manifest and `NSCameraUsageDescription` in Info.plist.

The docs obligation does separate from the code, and my deferral note covered only the code. `Activity.requestPermissions` fails on a permission the host app never declared, so the manifest requirement holds whatever Leg 1 finds about capture.

I ran the generator against this `examples.json`: the note renders as `> **Note:**` on `components.mdx`, and validation stays at zero warnings.

`withCrossmintUI` is the right home for the declaration. That goes with the Leg 1 code work, alongside `mediaCapturePermissionGrantType` and the same gap in `EmbeddedCheckoutV3WebView`.

packages/client/ui/react-native/src/components/identity-verification/crossmint-identity-verification.test.tsx
AngelPaella: Ported the web test in 7f599db.

I checked your read first. Swapping `latestProps.current.onComplete` for `props.onComplete` produced one failure, the new test. The `forwards 'kyc:completed' to 'onComplete'` row still passed, so the table could not have caught this: every other test renders once.

The suite goes from 7 to 10, this test plus two for the load-failure path.

packages/client/ui/react-native/scripts/generate-reference.mjs
AngelPaella: Your comment pushed me somewhere else: `getIdentityVerificationCredentials` reads better as a hook. It is the only camelCase non-hook export in either UI package, and merchants plumb the order by hand today:

```tsx
const { order } = useCrossmintCheckout();
const credentials = getIdentityVerificationCredentials(order);
```

A `useIdentityVerificationCredentials()` reading the checkout context collapses that to one line. Filed as ENG4-373. Out of scope here, since it adds public API to react-ui too and that one shipped in 4.4.0.

On adding it to this list, I tried it and hit two things.

It warns. Validation adds every `exports` name to `neededExamples`, so with no matching `examples.json` snippet the generator prints `⚠ [checkout] Missing example in examples.json for: "getIdentityVerificationCredentials"`. A snippet fixes that.

Where it lands is the real problem. `classifyExports` sorts by name alone (`*Provider`, `use*`, everything else), so a camelCase helper renders as a heading on `components.mdx` beside `CrossmintEmbeddedCheckout`. `descriptions` does not soften it either: only `buildProviders` reads that map, so `buildComponents` falls back to the JSDoc.

Left it out for now, which matches react-ui, where the same helper is exported and unlisted. Say the word if you would rather see it on the Components page than missing. If the hook lands it files to Hooks on its own and this goes away.

AngelPaella: i ended up landing useIdentityVerificationCredentials so i think this resolves it self

luans-qa: yeah leave it out for now, thats ok

luans-qa (approved):
lgtm

PR #2029, opened 2026-08-20
fix(quickstart-devkit): seed Solana test wallets from our own devnet keypair

luans-qa:
## The suite has been failing daily, and silently

[Run 32343465121](https://github.com/Crossmint/crossmint-sdk/actions/runs/32343465121) reports **success**. The real results:

| Browser | Failed | Flaky | Did not run | Passed |
|---|---|---|---|---|
| chromium | 2 | 2 | 4 | 12 |
| firefox | 2 | 2 | 4 | 12 |
| webkit | 2 | 0 | 4 | 14 |

The test step ran `pnpm exec playwright test ... || true`, so the exit code never reached the job, and the Slack notifier's `MATRIX_RESULT: ${{ needs.e2e-regression-tests.result }}` was always `success`.

## Root cause: bootstrap, not supply

Live devnet balances for the six per-browser wallets:

```
2jeHeFwLtLKDYgreboXqW7hzVAkiyMk2p3FigCt2DxBY  999833680 lamports   (chromium solana-email)
7nLg9UNWsENkLKMagNCJF8JdacgSEMd7pt7BpRbcYuvr          0 lamports
Fp8ibSCjcacDNzdXEEKdwBWFKR9P3FgEwTCaKuGCEYe2          0 lamports
8mN5fovn2jb6tSCie3498HqCM6rzhqH2C4yYwMEAwre2          0 lamports
Gky2i4FiTTXpw8Nr7n6MtYqLZzNr3E93iGW9ZF2DbPtm          0 lamports
58SqgBatpEDRyhrbhaw4CTANzW7ywH4rKYEcGUs4ursp          0 lamports
```

Five wallets have **never** been funded. They were created by the `TESTS_WALLET_EMAIL_SUFFIX: e2e-${{ matrix.browser }}` identity scheme and never received an initial airdrop, because `requestAirdrop` against `api.devnet.solana.com` now fails from every IP tried — CI runners *and* a residential connection (`SolanaJSONRPCError: ... Internal error`, and `429 ... the airdrop faucet has run dry`).

The one funded wallet holds 0.99983368 SOL and has spent 166,320 lamports (~0.000166 SOL) in its entire life. That confirms the premise already written in `wallet.ts`: a wallet needs seeding **once**, not per run. So the answer is a funding source we own, used for bootstrap only.

## Alternatives considered

| Option | Verdict |
|---|---|
| Crossmint faucet dispenses SOL | **Closed.** `faucet.dto.ts` validates `z.literal(CryptoCurrency.USDXM)`, `FAUCET_SUPPORTED_SOLANA_TOKENS = ["usdxm"]`, and there is no native SOL treasury. |
| Paid RPC via `SOLANA_DEVNET_RPC_URL` | **Weak.** The variable already exists and is unset, but providers proxy the same exhausted faucet for `requestAirdrop`. Wired into the docs anyway. |
| Gas sponsorship (`fees: { mode: "project" }`) | **Best long term, out of scope here.** `solana-gas-sponsorship.e2e.test.ts` in crossbit-main proves a 0-SOL wallet can transact, but it needs project fee config plus an app change. Worth a follow-up — it would delete this helper entirely. |
| Our own devnet keypair | **This PR.** No rate limits, no product change, ~1 SOL covers thousands of runs. |

## Changes

- `fundWalletWithSolAirdrop` transfers `FUNDER_TRANSFER_SOL` from `SOLANA_DEVNET_FUNDER_SECRET_KEY` when set, and keeps the `requestAirdrop` ladder as fallback. Absent the variable, behaviour is byte-for-byte today's.
- Secret wired into the workflow; documented in `e2e/README.md` and `.env.template`.
- `|| true` replaced with an explicit `tests_exit_code` so a red suite fails its job, preserving the "no results file" discovery check that ran after it.

## Action required before this helps

Add a **`SOLANA_DEVNET_FUNDER_SECRET_KEY`** repository secret: the 64-byte secret key as the JSON array `solana-keygen` writes (the contents of `~/.config/solana/id.json`), for a devnet keypair topped up via [faucet.solana.com](https://faucet.solana.com) — the captcha-gated web faucet still works where the RPC does not.

Until then the suite fails as it does today, only visibly.

## Not fixed here

`chromium / solana - email › transfers funds` failed with **1 SOL and 19.97 USDXM in the wallet** — `Transaction did not start - transfer button is still enabled` (`wallet.ts:299`), then `Transaction timeout: Success link did not appear within 2 minutes` (`wallet.ts:302`). That is a UI/transaction failure independent of funding and needs its own investigation. `firefox / evm - email › transfers funds` also failed.

## Test plan

- [x] Keypair round-trips through the JSON-array env encoding (`Keypair.fromSecretKey(Uint8Array.from(JSON.parse(...)))`).
- [x] The transfer instruction builds and signs against devnet (215-byte tx).
- [x] `requestAirdrop` confirmed failing from a non-CI IP, ruling out an IP-reputation fix.
- [x] Reworked run block verified for all three cases: tests pass → exit 0, tests fail → exit 1, no results file → error annotation + exit 1.
- [x] `tsc --strict` clean on the modified helper.
- [x] Backward compatible with the secret absent: the smoke suite (which shares this helper) passed 5/5 on this branch — `getFunderKeypair()` returns `null` and the old airdrop path runs unchanged.
- [ ] Add the secret, then `workflow_dispatch` the suite and confirm the five wallets get seeded.

### DIFF
--- a/.github/workflows/e2e-regression-tests.yml
+++ b/.github/workflows/e2e-regression-tests.yml
@@ -59,13 +59,20 @@ jobs:
                   # simultaneous OTP requests for the same email get rate-limited/blocked.
                   # Each browser gets its own persistent wallet, reused across runs.
                   TESTS_WALLET_EMAIL_SUFFIX: e2e-${{ matrix.browser }}
+                  # The public devnet faucet's requestAirdrop no longer works from the
+                  # runners, so Solana test wallets are seeded from our own keypair.
+                  SOLANA_DEVNET_FUNDER_SECRET_KEY: ${{ secrets.SOLANA_DEVNET_FUNDER_SECRET_KEY }}
               run: |
                   mkdir -p test-results
-                  pnpm exec playwright test --project=${{ matrix.browser }} || true
+                  set +e
+                  pnpm exec playwright test --project=${{ matrix.browser }}
+                  tests_exit_code=$?
+                  set -e
                   if [ ! -f test-results/playwright-results.json ]; then
                       echo "::error::No test results file generated — no tests were discovered. Check testMatch patterns in playwright.config.ts."
                       exit 1
                   fi
+                  exit $tests_exit_code
 
             - name: Generate test report
               if: always()
--- a/.github/workflows/smoke-tests.yml
+++ b/.github/workflows/smoke-tests.yml
@@ -59,9 +59,6 @@ jobs:
                   # Dedicated persistent wallet identity so smoke runs never share
                   # emails (and OTP requests) with concurrently running e2e jobs.
                   TESTS_WALLET_EMAIL_SUFFIX: smoke
-                  # The public devnet faucet's requestAirdrop no longer works from the
-                  # runners, so Solana test wallets are seeded from our own keypair.
-                  SOLANA_DEVNET_FUNDER_SECRET_KEY: ${{ secrets.SOLANA_DEVNET_FUNDER_SECRET_KEY }}
               run: |
                   pnpm exec playwright test --project=smoke
 
--- a/.github/workflows/smoke-tests.yml
+++ b/.github/workflows/smoke-tests.yml
@@ -59,6 +59,9 @@ jobs:
                   # Dedicated persistent wallet identity so smoke runs never share
                   # emails (and OTP requests) with concurrently running e2e jobs.
                   TESTS_WALLET_EMAIL_SUFFIX: smoke
+                  # The public devnet faucet's requestAirdrop no longer works from the
+                  # runners, so Solana test wallets are seeded from our own keypair.
+                  SOLANA_DEVNET_FUNDER_SECRET_KEY: ${{ secrets.SOLANA_DEVNET_FUNDER_SECRET_KEY }}
               run: |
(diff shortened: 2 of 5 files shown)


luans-qa:
Superseded by #2069.

The project behind the devkit test key now runs in `project` fee mode, so Crossmint sponsors transaction fees on every chain and no test wallet needs native currency. `fundWalletWithSolAirdrop` is deleted outright in `7c96ca32`, along with both call sites.

On-chain confirmation from a Solana transfer after the change: the fee payer is the relayer `34U6iAJmsmRUmodCbbqvSWKP46mjk49q7ykjdpW6dVeQ`, which pays all 10,040 lamports; the test wallet's SOL delta is `0`; and the `system: transfer 5544 lamports` reimbursement instruction that used to accompany every `spl-token: transferChecked` is gone. All five e2e configurations then pass 20 of 20 with every wallet at 0.00 native.

So the `SOLANA_DEVNET_FUNDER_SECRET_KEY` secret is not needed and the devnet faucet is off the critical path. Closing.

PR #2039, opened 2026-08-31
feat(wallets): accept a list of recovery signers on Solana and Stellar

panosinthezone:
## Description

Part 3/4 of splitting #2035 into a reviewable stack (stacked on #2038). This is the core feature: Solana and Stellar wallets can be created with up to 10 recovery signers; EVM stays single (backend DTO is still scalar).

Public types (`wallets/types.ts`):

```ts
type RecoveryCreateArg<C> = C extends SolanaChain | StellarChain
    ? RecoverySignerConfigFor<C> | Array<RecoverySignerConfigFor<C>>
    : RecoverySignerConfigFor<C>;
// WalletCreateArgs<C>.recovery: RecoveryCreateArg<C>  (still required, DeviceSignerConfig still excluded)
```

Wallet factory:
* Resolves every entry in the list (per-entry passkey creation and server signer derivation), sends arrays as `config.recovery` and scalars as the legacy `config.adminSigner` — never both.
* Client-side validation: rejects empty lists, lists longer than `MAX_RECOVERY_SIGNERS` (10), and lists on chains other than Solana/Stellar (`InvalidRecoveryConfigError` / `RecoverySignerLimitExceededError` / `RecoveryNotSupportedOnChainError`); wires `throwIfRecoverySignerApiError` into wallet creation.
* Response path prefers `config.recovery` and falls back to `[config.adminSigner]`, restoring caller-only data (server secrets, external-wallet callbacks, phone channel) per entry; existing-wallet validation matches input signers against the full existing list.

Wallet core: `wallet.recovery` now returns the full recovery signer list (`Array<RecoverySignerConfigForChain<C>>`) — the previously unannounced scalar getter is replaced per review, and there is no separate `recoverySigners` field. The first entry is sourced from the signer manager so it tracks signer adoption/secret-stripping, and the list is preserved through `SolanaWallet.from` / `StellarWallet.from` / `EVMWallet.from` rewrapping. A `toRecoverySignerList` helper normalizes scalar-or-list input and is exported.

React providers (`react-base` `CrossmintWalletBaseProvider`, `react-native` `CrossmintWalletProvider`) are updated in the same PR since they consume `WalletCreateArgs.recovery` and must handle the widened type to compile: email population, external-wallet readiness, passkey detection, and email/phone WebView initialization now iterate every recovery entry.

Signer *selection* for multi-admin wallets (transactions with N > 1 admins require an explicit `signer`) is intentionally left as a follow-up.

## Test plan

* `wallet-factory.test.ts`: Solana + Stellar list creation, per-entry passkey/server derivation, `recovery` vs `adminSigner` request field selection, empty/cap/unsupported-chain validation, response array parsing and legacy singular fallback, EVM scalar restriction.
* `solana.test.ts` / `stellar.test.ts`: chain rewrapping keeps every recovery signer.
* `wallet.test.ts`: `wallet.recovery[0]` follows the live recovery signer after `useSigner` adopts a fuller config.
* `pnpm --filter @crossmint/wallets-sdk test:vitest` (700 passed), `pnpm --filter @crossmint/client-sdk-react-base test:vitest` (10 passed), and the wallets/react-base/react-native builds pass.

## Package updates

* `@crossmint/wallets-sdk` (minor) — changeset `recovery-signer-list`.
* `@crossmint/client-sdk-react-base`, `@crossmint/client-sdk-react-native-ui` (patch) — changeset `recovery-list-providers`.

Link to Devin session: https://crossmint.devinenterprise.com/sessions/520c831e21b04f9f9d2fbb524107c536
Open in Devin Desktop: https://crossmint.devinenterprise.com/desktop/session/520c831e21b04f9f9d2fbb524107c536?variant=devin
Requested by: @panosinthezone

### DIFF
--- a/packages/wallets/src/utils/recovery.ts
+++ b/packages/wallets/src/utils/recovery.ts
@@ -1,0 +1,13 @@
+import type { Chain } from "../chains/chains";
+import type { RecoverySignerConfigForChain } from "../signers/types";
+import type { WalletCreateArgs } from "../wallets/types";
+
+/** Normalizes the single-or-list `recovery` wallet creation argument into a list. */
+export function toRecoverySignerList<C extends Chain>(
+    recovery?: WalletCreateArgs<C>["recovery"]
+): Array<RecoverySignerConfigForChain<C>> {
+    if (recovery == null) {
+        return [];
+    }
+    return (Array.isArray(recovery) ? recovery : [recovery]) as Array<RecoverySignerConfigForChain<C>>;
+}
--- a/packages/wallets/src/wallets/solana.ts
+++ b/packages/wallets/src/wallets/solana.ts
@@ -24,6 +24,7 @@ export class SolanaWallet extends Wallet<SolanaChain> {
                 options: Wallet.getOptions(wallet),
                 alias: wallet.alias,
                 recovery: Wallet.getRecovery(wallet),
+                recoverySigners: Wallet.getRecoverySigners(wallet),
                 apiRecoveryServerSignerAddress: Wallet.getApiRecoveryServerSignerAddress(wallet),
                 apiDelegatedServerSignerAddresses: Wallet.getApiDelegatedServerSignerAddresses(wallet),
                 signer: wallet.signer,
--- a/packages/wallets/src/wallets/solana.ts
+++ b/packages/wallets/src/wallets/solana.ts
@@ -24,7 +24,6 @@ export class SolanaWallet extends Wallet<SolanaChain> {
                 options: Wallet.getOptions(wallet),
                 alias: wallet.alias,
                 recovery: Wallet.getRecovery(wallet),
-                recoverySigners: Wallet.getRecoverySigners(wallet),
                 apiRecoveryServerSignerAddress: Wallet.getApiRecoveryServerSignerAddress(wallet),
                 apiDelegatedServerSignerAddresses: Wallet.getApiDelegatedServerSignerAddresses(wallet),
                 signer: wallet.signer,
--- a/packages/wallets/src/wallets/wallet-factory.ts
+++ b/packages/wallets/src/wallets/wallet-factory.ts
@@ -12,7 +12,16 @@ import type {
     Signer as SignerResponse,
     RegisterSignerParams,
 } from "../api";
-import { DEVICE_SIGNER_NOT_SUPPORTED_ERROR_CODE, WalletCreationError, WalletNotAvailableError } from "../utils/errors";
+import {
+    DEVICE_SIGNER_NOT_SUPPORTED_ERROR_CODE,
+    InvalidRecoveryConfigError,
+    MAX_RECOVERY_SIGNERS,
+    RecoveryNotSupportedOnChainError,
+    RecoverySignerLimitExceededError,
+    WalletCreationError,
+    WalletNotAvailableError,
+    throwIfRecoverySignerApiError,
+} from "../utils/errors";
 import { type Chain, validateChainForEnvironment } from "../chains/chains";
 import type {
     ExternalWalletRegistrationConfig,
@@ -21,7 +30,8 @@ import type {
     SignerConfigForChain,
 } from "../signers/types";
 import { Wallet } from "./wallet";
-import type { WalletArgsFor, WalletCreateArgs } from "./types";
+import type { RecoverySignerConfigFor, WalletArgsFor, WalletCreateArgs } from "./types";
+import { toRecoverySignerList } from "../utils/recovery";
 import { compareSignerConfigs, normalizeValueForComparison } from "../utils/signer-validation";
 import { getSignerLocator } from "../utils/signer-locator";
 import { deriveServerSignerDetails, deriveServerSignerCandidates } from "../signers/server";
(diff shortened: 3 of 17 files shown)


packages/wallets/src/utils/recovery.ts
```diff
@@ -0,0 +1,13 @@
+import type { Chain } from "../chains/chains";
+import type { RecoverySignerConfigForChain } from "../signers/types";
+import type { WalletCreateArgs } from "../wallets/types";
+
+/** Normalizes the single-or-list `recovery` wallet creation argument into a list. */
+export function toRecoverySignerList<C extends Chain>(
```
panosinthezone, line 6: @devinai Why do we have this weird method? Is it because EVM assumes a single signer object for now?

panosinthezone: @devinai What would it take to remove it right now? We are about to merge the EVM changes which includes a recovery taking a list so the shape would be the same for all.

panosinthezone: @alberto-crossmint, I am in favor of keeping this until Daniil lands the EVM changes. let me know if you think otherwise

packages/wallets/src/wallets/solana.ts
```diff
@@ -24,6 +24,7 @@ export class SolanaWallet extends Wallet<SolanaChain> {
                 options: Wallet.getOptions(wallet),
                 alias: wallet.alias,
                 recovery: Wallet.getRecovery(wallet),
+                recoverySigners: Wallet.getRecoverySigners(wallet),
```
panosinthezone: @devinai what is the difference between `recovery` and `recoverySigners`?

packages/wallets/src/wallets/wallet.ts
```diff
@@ -124,6 +128,7 @@ export class Wallet<C extends Chain> {
         } else if (recovery.type === "server" && isApiSourcedServerSignerConfig(recovery)) {
             apiRecoveryAddress = recovery.address;
         }
+        this.#recoverySigners = recoverySigners ?? [recovery];
```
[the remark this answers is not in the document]
panosinthezone: @devinai can you check this please

packages/wallets/src/wallets/wallet-factory.ts
```diff
@@ -32,10 +42,22 @@ const SIGNER_MISMATCH_ERROR =
     "When 'signers' is provided to a method that may fetch an existing wallet, each specified signer must exist in that wallet's configuration.";
 
 type SmartWalletConfig = {
+    /** @deprecated The API still returns the first admin signer here; `recovery` holds all of them. */
     adminSigner: RecoverySignerConfig | PasskeySignerConfig;
+    recovery?: Array<RecoverySignerConfig | PasskeySignerConfig>;
```
panosinthezone, line 47: @devinai why is this marked as optional?

panosinthezone: @devinai the api returns the recovery field. Can you confirm with the api response and update if required

panosinthezone: @alberto-crossmint same question as the above

packages/wallets/src/utils/recovery.ts
alberto-crossmint: Definitely! Steps 1 and 2 completion time is still uncertain and depends on security audit timelines on ZD side which we can't control. Keeping this normalization is beneficial. We can migrate this method to the factory entrypoint later

packages/wallets/src/wallets/wallet-factory.ts
alberto-crossmint: Hmm this one is about the RTO, right? I agree on keeping it optional until EVM lands. My lesson learnt here would be that these types should have been a discriminated union by a chainLayer generic or similar in the first place, but given the current state keeping it optional until EVM lands I think is perfectly fine 🙏

alberto-crossmint: Buuut pls check with @daniil-dovgal , I think that even when EVM is not fully live the changes that add recovery to the RTO on the backend side (even if it's just a single-field array containing the admin signer) is a non-breaking change that may avoid this complication in the SDK

alberto-crossmint: But non blocking! Your solutions looks perfectly fine to me in the meantime

alberto-crossmint (approved):

packages/wallets/src/wallets/solana.ts
panosinthezone: @devinai `recovery` is the proper name of the field where we would keep all the recovery/admin signers. Don't worry about breaking the already existing `recovery` since we didn't announce it. This means that `recoverySigners` can be dropped and `recovery` would function in the same way you implemented `recoverySigners`

packages/wallets/src/wallets/wallet-factory.ts
panosinthezone: Yeah, the EVM are not live. The evm type is signer | signer[] so it is weird

panosinthezone: Will keep optional

PR #2051, opened 2026-09-09
Wallets SDK: sign Solana approvals from the message the API supplies

alberto-crossmint:
The two ed25519 Solana signers rebuilt the payload they signed:

```ts
VersionedTransaction.deserialize(bytes).message.serialize()
```

`@solana/web3.js` cannot serialize a version-1 message, and the pinned 1.98.1 cannot read one. A version-1 approval failed with `Reached end of buffer unexpectedly`, which names neither Solana nor a version.

## Why this shape

The API already supplies the bytes to sign in `pendingApproval.message`, and Solana device signers already use it. Swift, Kotlin and Flutter all sign that field too. This SDK was the only one that re-derived the payload, and the re-derivation is the whole defect — so the fix makes it match the others rather than teaching it to encode version 1. That also avoids `@solana/kit`: roughly 56 packages peer-pin `@solana/web3.js@^1`, and an override cannot redirect a peer dependency.

Choosing the approval payload moves out of `wallet.ts` into `ChainAdapter.signApproval`, so the Solana rule lives in the Solana adapter and `wallet.ts` names neither a chain nor a signer type. Dispatch still keys off the reported `chainType`, preserving the cross-chain behaviour the integration suite pins.

## External wallets on version 1

`SolanaExternalWalletSignerConfig` takes an optional `onSignBytes(payload) => signature`, base58 both ways, and `signMessage` routes to it in place of rejecting. No wallet adapter can sign a version-1 transaction: web3.js deserializes the message but cannot serialize it, so `VersionedTransaction.sign` throws, and `onSign` must return a *signed* transaction. Version 0 and legacy still go to `onSign`, unchanged, and `onSign` stays required. A version-1 transaction with no `onSignBytes` fails with an error naming the callback.

## Version 0 signs the same bytes as before

`wireMessageBytes(wire)`, which is what the API encodes into `message`, is byte-identical to what the old code derived. Verified by execution against the backend builder before writing the fix, for a version-0 transaction.

## What this does NOT do

Browser wallet adapters still cannot approve a version-1 transaction; `onSignBytes` needs a directly held key. This does not migrate to `@solana/kit` and does not raise the `@solana/web3.js` pin.

## Test plan

`npx vitest run` in `packages/wallets`: **737 passed, 68 skipped, 0 failed**. In `packages/common/base`: **18 passed**.

- `npx tsc --noEmit -p tsconfig.json` clean in both. `@crossmint/common-sdk-base` must be built first — the wallets package resolves it to `dist`, not source.
- `biome check` clean on all changed files.
- An integration test approves a version-1 transaction through `wallet.approve` with a real keypair, and verifies the submitted signature against the API's message bytes with `nacl`.
- Mutation checks on the version-1 routing: forcing the test to always-false and to always-true each fails 2 tests.


packages/wallets/src/wallets/wallet.ts
```diff
@@ -1302,16 +1302,13 @@ export class Wallet<C extends Chain> {
...
-                        ? (transaction.onChain.transaction as string)
-                        : pendingApproval.message;
+                const wantsWholeTransaction = transaction.chainType === "solana" && signer.type === "external-wallet";
 
-                return signer.signTransaction(transactionToSign);
+                return signer.signTransaction(
+                    wantsWholeTransaction && "transaction" in transaction.onChain
+                        ? (transaction.onChain.transaction as string)
+                        : pendingApproval.message
+                );
```
alberto-crossmint: let's create `getSignablePayload` that contains a switch statement on `signer.type`. This is not good code

alberto-crossmint: Done, across three commits. The callback is now one line, and the comparison is behind a guard:

```ts
type SerializedTransactionOnChain = Extract<GetTransactionSuccessResponse["onChain"], { transaction: string }>;

function hasSerializedSolanaTransaction(
    transaction: GetTransactionSuccessResponse
): transaction is GetTransactionSuccessResponse & { onChain: SerializedTransactionOnChain } {
    return (
        transaction.chainType === "solana" &&
        "transaction" in transaction.onChain &&
        typeof transaction.onChain.transaction === "string"
    );
}
```

`Extract` was the right call — it keeps the real union members, so `lastValidBlockHeight` and `txId` survive the narrowing instead of being flattened into a synthesized shape. The `as string` cast is gone with it, because the guard now checks `typeof` rather than asserting.

Two things I checked rather than assumed, since both would have compiled either way:

**The switch is not exhaustiveness-checked.** I first wrote all seven signer types as explicit cases with no `default`, expecting TypeScript to force a decision when a type is added. `noImplicitReturns` is off in this package, so deleting the `api-key` case compiled clean and the function silently returned `undefined` — which would then be signed. It has a `default` now.

**`Extract` is not `never`.** A `never` there would also have compiled, and the guard would have been vacuous. `[SerializedTransactionOnChain] extends [never]` resolves to `"OK"`, so it really does match the two string-carrying members.

packages/wallets/src/signers/non-custodial/ncs-solana-signer.ts
```diff
@@ -17,10 +16,8 @@ export class SolanaNonCustodialSigner extends NonCustodialSigner {
         return await this.sign(base58.decode(message));
     }
 
-    async signTransaction(transaction: string): Promise<{ signature: string }> {
-        const transactionBytes = base58.decode(transaction);
-        const deserializedTransaction = VersionedTransaction.deserialize(transactionBytes);
-        return await this.sign(deserializedTransaction.message.serialize());
+    async signTransaction(message: string): Promise<{ signature: string }> {
+        return await this.signMessage(message);
```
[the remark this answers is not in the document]
alberto-crossmint: Deferring this, deliberately, and it is worth saying why rather than dismissing it.

The observation is accurate: `signTransaction` is reachable as `wallet.signer.signTransaction(...)`, and for these two signers the argument is now the approval message rather than a serialized transaction.

Two things make it the right trade here:

**The transaction-oriented contract was never uniform.** `DeviceSigner`, `PasskeySigner` and `EVMExternalWalletSigner` already implement `signTransaction` as `return this.signMessage(payload)` — they sign the bytes they are handed. The two Solana ed25519 signers were the outliers, so this makes them consistent with the rest of the SDK rather than departing from a shared rule.

**The alternative cannot fix the bug.** Keeping the transaction-oriented contract means keeping the deserialize-and-re-serialize step, and `@solana/web3.js` cannot serialize a version-1 message. There is no version of "preserve the old contract" that also lets a version-1 approval through.

The point underneath — that a method named `signTransaction` taking a message is a confusing public API — stands, and is tracked as a follow-up. The SDK can keep dealing in `VersionedTransaction` at its boundary as long as the signer can sign whatever message it is given.

alberto-crossmint:
## Smoke Tests is red, and it is not this PR

`build & test` and `lint` pass. `Run Smoke Tests` fails, and it fails the same way on `main`.

The failing case here:

```
✘  4 [smoke] › tests/smoke/specs/wallets/smoke.spec.ts:106:9 › Wallet Smoke › transfers funds @smoke (2.4m)
   Error: Transaction did not start - transfer button is still enabled
1 failed
3 passed (4.5m)
```

The most recent `main` run, 2026-09-08 (run 34218402880):

```
✘  4 [smoke] › tests/smoke/specs/wallets/smoke.spec.ts:106:9 › Wallet Smoke › transfers funds @smoke (2.4m)
   Error: Transaction did not start - transfer button is still enabled
1 failed
3 passed (4.5m)
```

Same test, same error string, same counts. Every smoke run in the last 100 workflow runs is red — on `main`, and on unrelated `devin/` and `codex/` branches — going back to at least 2026-09-03.

## Why this PR cannot be the cause

The failing wallet is `crossmint-evm-email-base-sepolia`, an EVM wallet. This PR changes Solana signing only, and the routing helper is a no-op off Solana:

```ts
function hasSerializedSolanaTransaction(transaction: GetTransactionSuccessResponse): ... {
    return (
        transaction.chainType === "solana" &&
        ...
    );
}
```

For an EVM transaction that returns false, so `getSignablePayload` returns `pendingApproval.message` — exactly what the previous code returned for EVM. The bytes an EVM signer receives are unchanged.

Not opening an issue for the smoke suite here, since it belongs to whoever owns that workflow, but it is worth someone's attention: it has been red long enough to stop being a signal.

.changeset/solana-version-1-signers.md
```diff
@@ -0,0 +1,5 @@
+---
+"@crossmint/wallets-sdk": patch
+---
+
+Sign Solana approvals from the message the API supplies, so a version-1 transaction can be approved. `SolanaNonCustodialSigner` and `SolanaServerSigner` previously rebuilt the payload with `VersionedTransaction.deserialize(...).message.serialize()`, which throws `Reached end of buffer unexpectedly` on a version-1 transaction. Both now sign the supplied bytes, and `wallet.approve` routes `pendingApproval.message` to every Solana signer except an external wallet, which still needs the whole transaction for its wallet adapter. The bytes signed are unchanged for version 0.
```
albertoelias-crossmint: so do external-wallet signers not work with v1?

packages/wallets/src/wallets/wallet.ts
```diff
@@ -86,6 +86,31 @@ type WalletContructorType<C extends Chain> = {
     signer?: SignerAdapter;
 };
 
+type SerializedTransactionOnChain = Extract<GetTransactionSuccessResponse["onChain"], { transaction: string }>;
```
albertoelias-crossmint: as this is solana specific, doesnt seem like wallet.ts is the place for this

albertoelias-crossmint (commented):
shouldnt the solution be to move the sdk away from web3js to the new sdk and also support signing v1 txs from external-wallet?

packages/wallets/src/wallets/wallet.ts
alberto-crossmint: This PR doesn't change that, the exception was already there, only that it was inlined.

Maybe let's delete `hasSerializedSolanaTransaction` and let's do

```typescript
switch (signer.type) {
        case "external-wallet":
           if(transaction.chainType === "solana" &&
          "transaction" in transaction.onChain &&
          typeof transaction.onChain.transaction === "string"){
            return transaction.onChain.transaction;
           }
        default:
            return approvalMessage;
    }
```

alberto-crossmint: Done in `ef37494d`. `hasSerializedSolanaTransaction` and `SerializedTransactionOnChain` are both gone, and the check is inline in the `case "external-wallet"` arm.

One difference from the snippet: I kept an explicit `return approvalMessage;` in the case, instead of falling through to `default`. Both compile here — the repo sets neither `noFallthroughCasesInSwitch` nor Biome's `noFallthroughSwitchClause` — but a silent fall-through breaks the moment either rule is turned on.

On your original point, to be straight about it: `wallet.ts` still tests `transaction.chainType === "solana"`, so the Solana knowledge shrank, it did not leave the file. Removing it properly means the signer declares which payload it needs, rather than `wallet.ts` asking what type each signer is. That is a wider change than this PR, so I left it.

.changeset/solana-version-1-signers.md
alberto-crossmint: They did not. `fb4a0a65` fixes that.

The blocker was never the key type — it was `@solana/web3.js`. Measured against the v1 fixture in this PR:

| call | 1.98.1 | 1.99.0 |
|---|---|---|
| `VersionedTransaction.deserialize` | throws `Transaction message version 1 deserialization is not supported` | OK |
| `message.serialize()` | — | throws `Serialization of version 1 transaction messages is not supported` |
| `tx.sign([keypair])` | — | throws the same |

1.99.0 added **read** support for v1, not write. `VersionedTransaction.sign()` calls `this.message.serialize()` to get the bytes to sign, so a version bump alone does not help. It only moves the error out of the SDK and into the caller's `onSign`.

So `SolanaExternalWalletSignerConfig` now takes an optional `onSignBytes(payload) => signature`, base58 both ways. `SolanaExternalWalletSigner` reads the version off the serialized transaction and calls `onSignBytes` **only** for version 1. Version 0 and legacy transactions keep the existing `onSign` adapter path, unchanged. `onSign` stays required, so this is additive.

The payload is the message sliced past the signature array. On version 0 that slice is byte-identical to `message.serialize()`, so it is the same Ed25519 preimage the adapter path signs — there is a test that asserts exactly this.

Two practical notes. A browser wallet adapter cannot supply `onSignBytes`; this is for a directly held key. And a v1 transaction with no `onSignBytes` configured now fails with an error naming the callback, in place of the web3.js message above.

No web3.js bump in this PR — the v1 path never calls web3.js.

packages/wallets/src/signers/solana-external-wallet.ts
```diff
@@ -7,14 +7,30 @@ import { ExternalWalletSigner } from "./external-wallet-signer";
...
-    async signMessage() {
-        return await Promise.reject(new Error("signMessage method not implemented for solana external wallet signer"));
+    /**
+     * Sign a payload directly with the external wallet key.
+     * Used for version-1 transactions, which no wallet adapter can sign.
+     * @param message - The payload to sign, base58 encoded
+     */
+    async signMessage(message: string) {
+        if (this.onSignBytes == null) {
+            throw new Error(
```
alberto-crossmint, line 20: We can't ensure at this point that it's a v1 transaction. It may be just arbitrary bytes signing.

alberto-crossmint: [Response approved by Human]

Right, corrected in `c459cfb5`. `signMessage` signs whatever payload it is given, so it cannot assert the payload is version 1. The error now states only what is true, matching the `onSign` one:

```
[SolanaExternalWalletSigner] No onSignBytes callback provided. Pass an onSignBytes callback when configuring the external wallet signer.
```

The version-1 decision lives in `solanaChainAdapter.signApproval`, which is what routes an approval here. The signer itself is now just "sign these bytes" and "sign this transaction", with no version knowledge.

packages/wallets/src/wallets/wallet.ts
alberto-crossmint: [Response approved by Human]

Moved. Choosing the approval payload is now `ChainAdapter.signApproval`, so the Solana rule lives in `chains/adapters/solana.ts`. `getSignablePayload` is gone and `wallet.ts` names neither a chain nor a signer type:

```ts
getApprovalAdapter(transaction.chainType, this.chainAdapter).signApproval(signer, transaction, pendingApproval.message)
```

Dispatch still keys off the reported `chainType` rather than the wallet's own chain. That preserves the behaviour `wallet.transactions.test.ts` already pins for a Stellar or EVM wallet whose response claims `solana`. It looks like a latent oddity to me, since a wallet is chain-scoped, but changing it is out of scope here.

.changeset/solana-version-1-signers.md
alberto-crossmint: [Response approved by Human]

They did not. They do now.

`SolanaExternalWalletSignerConfig` takes an optional `onSignBytes(payload) => signature`, base58 both ways, and `signMessage` routes to it in place of rejecting. Version-1 approvals go there because no wallet adapter can sign one. Measured on 1.99.0 against a real version-1 transaction:

```
deserialize          -> OK, version 1
message.serialize()  -> Serialization of version 1 transaction messages is not supported
tx.sign([keypair])   -> same
```

`onSign` must return a *signed* transaction, and signing is exactly what throws. A web3.js bump does not fix it, and `@solana/kit` does not either: `@solana/compat` converts web3.js to kit only, with no `toVersionedTransaction`.

The payload is `pendingApproval.message` itself, the same field every other Solana signer uses. Nothing is derived from the serialized transaction. Version 0 and legacy keep the `onSign` adapter path, and `onSign` stays required.

One limit worth stating: `onSignBytes` needs a directly held key. A browser wallet adapter cannot serve it.

packages/wallets/src/chains/adapters/solana.ts
```diff
@@ -1,9 +1,27 @@
...
 import type { AddSignerChain, AddSignerContext, ChainAdapter } from "../chain-adapter";
 import type { Chain } from "../chains";
 
+function isVersion1Message(approvalMessage: string): boolean {
+    let prefix: number | undefined;
+    try {
+        prefix = base58.decode(approvalMessage)[0];
+    } catch {
+        return false;
+    }
```
alberto-crossmint: If it fails it doesn't mean that it's a v0. We should avoid implicit behavior like this. If decoding fails it means there's data inconsistency and we want to fail

alberto-crossmint:
> shouldnt the solution be to move the sdk away from web3js to the new sdk and also support signing v1 txs from external-wallet?

That'd be a breaking change because objects are incompatible, so it would break many customer integrations. The hard part of this PR is the backwards compatibility

packages/wallets/src/chains/adapters/solana.ts
alberto-crossmint: [Response approved by Human]

Agreed, fixed in `60dc639a`. The `try/catch` is gone and `base58.decode` throws:

```ts
function isVersion1Message(approvalMessage: string): boolean {
    const prefix = base58.decode(approvalMessage)[0];
    return prefix != null && (prefix & 0x80) !== 0 && (prefix & 0x7f) === 1;
}
```

Worth naming why it was there, because the reason was bad: the integration fixtures used placeholders like `"msg-for-device"`, which are not base58, and I made production code absorb that instead of fixing the fixtures. They now carry base58 payloads through an `approvalMessage(label)` helper, which also makes the constraint visible to the next person editing them.

A new test pins the behaviour: an undecodable approval message throws, and neither `signMessage` nor `signTransaction` is called.

One thing to flag: the error that surfaces is base-x's own `Non-base58 character`, which does not say which field was bad. I left it unwrapped rather than invent an error type, but say the word if you would rather it carried the field name.

albertoelias-crossmint (approved):

PR #2052, opened 2026-09-09
Embedded checkout: read the Solana transaction as versioned

alberto-crossmint:
The embedded checkout's Solana handler parsed with the legacy `Transaction` class:

```ts
deserializedTransaction = Transaction.from(base58.decode(serializedTransaction));
```

That class reads pre-versioned transactions only. Measured against `@solana/web3.js` 1.98.1, the version this package pins:

```
version 0: THROWS Versioned messages must be deserialized with VersionedMessage.deserialize()
version 1: THROWS Reached end of buffer unexpectedly
```

The failure is caught and reported as `"Failed to deserialize transaction"`, so a buyer sees a generic error and the real reason stays in the console.

`VersionedTransaction.deserialize` reads legacy and version 0 alike — verified under the same pinned 1.98.1 — and Dynamic's signer accepts either object: `signAndSendTransaction<T extends Transaction | VersionedTransaction>`.

## The one behavioural change

For a **legacy** payload the wire bytes are unchanged, but the object handed to Dynamic is now a `VersionedTransaction` whose `message.version` is `"legacy"`, where before it was a legacy `Transaction`. Dynamic's type covers both. This is worth a reviewer's eye if any connector in the wild treats the two differently.

## What this does NOT do

Version 1 stays unsupported here, and deliberately so. This package pins `@solana/web3.js` 1.98.1, which cannot read a version-1 transaction; reading one needs 1.99.0-beta.0. Even then a third-party wallet signing it calls `MessageV1.serialize`, which throws by design. That is the same ecosystem constraint that blocks `SolanaExternalWalletSigner`, and it is not solvable in this PR.

I also did not determine which version the payments backend sends to this handler today. If it sends legacy, this is a forward fix; if it already sends version 0, this repairs a live break. Either way the handler should accept what it is given.

## Test plan

`npx vitest run` in `packages/client/ui/react-ui`: **40 passed, 0 failed** across 9 files, including 3 new cases.

- The new tests use checked-in base58 fixtures rather than building transactions in-test, because this package runs vitest under `jsdom` where web3.js's `SystemProgram.transfer` fails with `TypeError: b must be a Uint8Array`.
- Reverting the parser to `Transaction.from` fails both new transaction cases, so they do catch the defect rather than passing regardless.
- `biome check --diagnostic-level=error` clean across the repo; `tsc --noEmit` reports nothing for the changed files.

### DIFF
--- a/.changeset/embed-solana-versioned-transaction.md
+++ b/.changeset/embed-solana-versioned-transaction.md
@@ -1,0 +1,5 @@
+---
+"@crossmint/client-sdk-react-ui": patch
+---
+
+Parse the embedded checkout's Solana transaction with `VersionedTransaction.deserialize` instead of the legacy `Transaction.from`. The legacy parser rejects every versioned transaction — a version-0 payload failed with `Versioned messages must be deserialized with VersionedMessage.deserialize()`, surfacing to the buyer as "Failed to deserialize transaction". `VersionedTransaction.deserialize` accepts legacy and version-0 alike, and Dynamic's `signAndSendTransaction` accepts either object. Legacy payloads keep working, and now reach the wallet as a `VersionedTransaction` whose `message.version` is `"legacy"`.
--- a/packages/client/ui/react-ui/src/components/embed/v3/crypto/utils/handleSolanaTransaction.test.ts
+++ b/packages/client/ui/react-ui/src/components/embed/v3/crypto/utils/handleSolanaTransaction.test.ts
@@ -1,0 +1,63 @@
+import base58 from "bs58";
+import { describe, expect, it, vi } from "vitest";
+import { handleSolanaTransaction } from "./handleSolanaTransaction";
+
+const LEGACY_TRANSACTION =
+    "3md7BBV9wFjYGnMWcMNyAZcjca2HGfXWZkrU8vvho66z2sJMZFcx6HZdBiAddjo2kzgBv3uZoac3domBRjJJSXkbBvokxThZ5oh8T5ym9k6AXbRYtAAR6a3pECdz2pFCX2shi8b67DGHMtQVjLbdftyzbVvWeGJDUdj69qTnm6u6Lx2Z9iCXdGk4QSsRem8cHoAcpSq997N7BwMeauKofu9CA3Wa7YZYhdyoGk3LUF4B5U39sYKt6rpbWmxBAqT5sDh3ysRCy12ouPfuFcPFvubnf5GQRhmpB1yKD";
+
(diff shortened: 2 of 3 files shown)


alberto-crossmint:
## Answering the open question in the description: payments sends **legacy**, so this is a forward fix, not a live one

The description said I had not determined which version the payments backend sends to this handler, and that if it already sent version 0 the flow would be broken in production. **It sends legacy, so production is fine and no backport is needed.**

`generateSolanaSerializedTransaction.ts` builds with the legacy `Transaction` class (line 66), serialises with `{ requireAllSignatures: false, verifySignatures: false }`, and base58-encodes (lines 113-119). It is reached from `prepareOrderCryptoPayment.ts:484` and `:517` and stored as `cryptoProcessorDataSchema.serializedTransaction`. The only sibling producer, `generateSolanaNativePassthroughSerializedTransaction.ts`, ends in `throw new Error("Not implemented")`, so nothing else feeds that field.

The exact payload was round-tripped through both readers:

```
backend output: base58, decoded 215 bytes; first byte (shortvec sig count) = 1
Transaction.from                 -> OK, instructions: 1
VersionedTransaction.deserialize -> OK, message.version = legacy, class Message
                                    sign([keypair]) OK, re-serialized 215 bytes
```

Both parsers handle it, and the re-serialised byte count is identical.

## What that changes about this PR

Nothing in the diff, and nothing about whether it should merge. It changes only the urgency: this is not repairing a live break, it removes the failure that appears the first time payments emits a versioned transaction on this path.

It is also worth stating plainly that this path arguably **should** stay legacy on purpose. The buyer's own wallet signs here, through Dynamic, so the transaction has to be one that third-party wallets handle without reservation. Accepting versioned input is still the right behaviour for a handler — it should parse what it is given — but nobody should read this PR as a reason to start sending version 1 to the embedded checkout.

Credit where due: the round-trip evidence above came from a parallel investigation into the payments transaction path, now tracked as WAL-12263 through WAL-12269 under milestone M8.

alberto-crossmint:
CI note: `build & test` and `lint` pass. `Run Smoke Tests` fails, and it is not this PR — the same test fails identically on `main` (run 34218402880: `smoke.spec.ts:106 Wallet Smoke > transfers funds`, `Transaction did not start - transfer button is still enabled`, `1 failed / 3 passed`). Every smoke run in the last 100 workflow runs is red, on `main` and on unrelated branches, since at least 2026-09-03. The failing wallet is EVM on Base Sepolia; this PR touches only the Solana handler in the embedded checkout.

PR #2053, opened 2026-09-10
fix(wallets): preserve available balances and card breakdown

jcurbelo:
## Summary

`wallet.balances()` drops `available`, `locked`, and `accounts` from the balance API response, leaving apps with the total balance even when card spending power differs. Preserve those optional fields for the wallet's chain.

Extend the existing `TokenBalance` with shared generated chain balance fields using `Omit<..., "locator" | "amount" | "rawAmount">`. The existing base fields, optionality, chain identifiers, and total amounts remain unchanged. Tokens without an account breakdown, including plain MGUSD, retain their current shape.

Includes a patch changeset for `@crossmint/wallets-sdk`.

## Test plan

- [x] Build the wallets SDK and its six workspace dependencies.
- [x] Wallet test suite: 728 passed, 68 integration tests skipped.
- [x] Formatter regressions cover enriched USDC, plain MGUSD, additional requested tokens, and chain selection.
- [x] Type-check consumers of the built SDK: optional balance fields, card-provider narrowing, plain token balances, and compatibility across Stellar, Solana, and EVM.
- [x] Biome check on changed TypeScript files and `git diff --check`.

Live Rain staging verification was not run.

---

<a href="https://crossmint.devinenterprise.com/review/crossmint/crossmint-sdk/pull/2053" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>


packages/wallets/src/wallets/types.ts
```diff
@@ -281,7 +281,8 @@ export type TokenBalance<C extends Chain = Chain> = {
     amount: string;
     decimals?: number;
     rawAmount?: string;
-} & ChainExtras[ChainToExtrasKey<C>];
+} & Pick<GetBalanceSuccessResponse[number]["chains"][string], "available" | "locked" | "accounts"> &
```
albertoelias-crossmint: we should update the openapi json file instead of modifying this manually

jcurbelo: updated in [fe0405d8](https://github.com/Crossmint/crossmint-sdk/commit/fe0405d844a64588408383bee8fbc1d3f37f496b) - `TokenBalance` now derives from the generated token and chain types, keeping only the SDK-specific overrides. the formatter passes the API fields through too. the Rain fields are already in `openapi.json`, so no spec change was needed.

jcurbelo: simplified in [f01e662f](https://github.com/Crossmint/crossmint-sdk/commit/f01e662f29b99a69017473eb53c78bf9c2ffddca) - back to the first version, replacing only `Pick` with `Omit<..., "locator" | "amount" | "rawAmount">`. the base fields and optionality stay as before.

albertoelias-crossmint (approved):

PR #2056, opened 2026-09-10
fix(devkit-tests): give the smoke suite a new wallet on every run

luans-qa:
Fixes the smoke suite failure diagnosed in https://github.com/Crossmint/crossmint-sdk/pull/2049#issuecomment-5532680741.

## The failure

`Wallet Smoke › transfers funds` fails on `main`, and has since 2026-08-31. Latest run on `main` ([34218402880](https://github.com/Crossmint/crossmint-sdk/actions/runs/34218402880)): tests 1-3 pass, test 4 fails after 2.4m with

```
Error: Transaction did not start - transfer button is still enabled
```

That message comes from `tests/shared/utils/wallet.ts:299` and hides the real cause. The transfer is rejected by the backend.

## Why

`TESTS_WALLET_EMAIL_SUFFIX` was the fixed string `smoke`, so `getEmailForSigner` returned the same address on every run, and wallet creation is idempotent on that identity. One wallet therefore served every smoke run ever.

Each CI run starts in a clean browser with no device key, so `preAuthIfNeeded()` calls `recover()` and enrols another device signer on that wallet. `DeviceRecoveryService#resolveStaleDeviceSignerLocator` only prunes when the wallet has exactly one device signer, or exactly one with a matching device name, so none of the old ones were removed. The wallet reached 136 device signers. Paella-Labs/crossbit-main#29415 then added a combined signer cap (16 on staging, per Paella-Labs/crossbit-main#29636) and the backend started returning `400 SIGNER_LIMIT_EXCEEDED`.

The same wallet also made `wallet.signers()` issue one `getSignerState` call per delegated signer, which tripped the project rate limit.

## The change

Pass a run-scoped value through the existing suffix, which `globalConstants.ts` already documents as the way to move to a new identity. No new helper, no change to how the tests obtain a wallet.

A new EVM wallet needs no seeding. Both transfer tests already check the balance and call `fundWalletWithCrossmintFaucet` when it is short, and gas is sponsored on base-sepolia.

`github.run_id` is stable across re-run attempts, so a re-run keeps the wallet it created.

## Scope

The e2e suites keep their fixed suffix. Their Solana wallets need native SOL, and the devnet faucet is dry (`the airdrop faucet has run dry` appears in every recent e2e log); their Stellar wallets need XLM and this repo has no helper for it. Those wallets have to stay funded between runs, so they cannot rotate until funding is solved. Their EVM arms will eventually hit the same cap.

The e2e workflow also masks its own exit code with `|| true`, which is why it reports success while failing. #2029 covers that separately.

## Test plan

- [ ] Smoke workflow passes on this branch. The push triggers it, because the diff touches `apps/wallets/quickstart-devkit/**`, and the run uses this branch's workflow file.
- [ ] Job log shows `TESTS_WALLET_EMAIL_SUFFIX` resolved to `smoke-<run id>` and all 5 tests pass, including `transfers funds`.

### DIFF
--- a/.github/workflows/smoke-tests.yml
+++ b/.github/workflows/smoke-tests.yml
@@ -56,9 +56,12 @@ jobs:
                   MAILOSAUR_SERVER_ID: ${{ secrets.SERVER_ID_MAILOSAUR }}
                   MAILOSAUR_PHONE_NUMBER: ${{ secrets.PHONE_NUMBER_MAILOSAUR }}
                   PLAYWRIGHT_BASE_URL: "http://localhost:3000"
-                  # Dedicated persistent wallet identity so smoke runs never share
-                  # emails (and OTP requests) with concurrently running e2e jobs.
-                  TESTS_WALLET_EMAIL_SUFFIX: smoke
+                  # A new wallet identity per run, kept distinct from the e2e jobs so the
+                  # two never share an email (and its OTP requests). A reused wallet
+                  # collected one device signer per run, because each CI browser starts
+                  # with no device key, until the wallet reached the backend signer cap
+                  # and every transfer failed. run_id is stable across re-run attempts.
+                  TESTS_WALLET_EMAIL_SUFFIX: smoke-${{ github.run_id }}
               run: |
                   pnpm exec playwright test --project=smoke
 
--- a/.github/workflows/smoke-tests.yml
+++ b/.github/workflows/smoke-tests.yml
@@ -56,11 +56,8 @@ jobs:
                   MAILOSAUR_SERVER_ID: ${{ secrets.SERVER_ID_MAILOSAUR }}
                   MAILOSAUR_PHONE_NUMBER: ${{ secrets.PHONE_NUMBER_MAILOSAUR }}
                   PLAYWRIGHT_BASE_URL: "http://localhost:3000"
-                  # A new wallet identity per run, kept distinct from the e2e jobs so the
-                  # two never share an email (and its OTP requests). A reused wallet
-                  # collected one device signer per run, because each CI browser starts
-                  # with no device key, until the wallet reached the backend signer cap
-                  # and every transfer failed. run_id is stable across re-run attempts.
+                  # A new wallet per run; a reused one collects a device signer every run
+                  # until it reaches the backend signer cap and transfers start failing.
                   TESTS_WALLET_EMAIL_SUFFIX: smoke-${{ github.run_id }}
               run: |
                   pnpm exec playwright test --project=smoke
--- a/apps/wallets/quickstart-devkit/e2e/README.md
+++ b/apps/wallets/quickstart-devkit/e2e/README.md
@@ -12,7 +12,8 @@ Tests are configured to:
 - **Mailosaur errors**: Verify your API key and server ID are correct
 - **Timeout issues**: Check if the local dev server is running on port 3000
 - **OTP failures**: Tests run sequentially to avoid multiple OTP codes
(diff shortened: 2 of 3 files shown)


luans-qa:
## Verified on this branch

The push triggered the smoke workflow with this branch's workflow file. Run [34511818127](https://github.com/Crossmint/crossmint-sdk/actions/runs/34511818127) — **5 passed (3.1m)**, against 1 failed / 1 did not run / 3 passed on `main`.

```
🔑 Starting email OTP login for: test-email-smoke-34511818127@***.mailosaur.net
📋 Wallet address from data-testid: 0xc093185B63160e38C9C27E791c7C8cb12Ba9cDe0
💰 Funding wallet 0xc093185B63160e38C9C27E791c7C8cb12Ba9cDe0 on base-sepolia with 10 usdxm...
✅ Successfully funded wallet ... with 10 usdxm on base-sepolia
✅ Transfer of 0.001 completed successfully
  ✓  1 authenticates and creates wallet (21.0s)
  ✓  2 fetches wallet balances via SDK (37ms)
  ✓  3 fetches wallet activity (30.0s)
  ✓  4 transfers funds (30.6s)
  ✓  5 creates and approves a prepared transaction (40.6s)
  5 passed (3.1m)
```

The email carries the run id, the wallet is new (`0xc093…cDe0`, not the accumulated `0xe37b…0962`), the faucet seeded it, and both transfer tests passed. Gas needed no seeding, which confirms sponsorship covers a first-use wallet on base-sepolia.

### One observation, not a regression

The faucet runs twice per run, once in each transfer test, so a run consumes 20 USDXM rather than 10. `getWalletBalance` waits 10s for a `/balances` response, gets none after the first transfer, and falls back to a UI value that has not refreshed yet, so the second test funds again. This is existing behaviour and it also applied to the old wallet whenever it ran dry. Worth a separate look if faucet volume matters.

luans-qa:
## Controlled A/B: the wallet identity is the cause

The green run alone was weak evidence, so I ran a same-day control on `main` to isolate the variable.

| | Suffix | Wallet | Result |
|:--|:--|:--|:--|
| **Control** — `workflow_dispatch` on `main`, [run 34515415286](https://github.com/Crossmint/crossmint-sdk/actions/runs/34515415286) | `smoke` | `0xe37b6dd43f2939564dB7E1c0e83B45C113C30962` | `✘ 4 transfers funds (2.4m)` |
| **Treatment** — this branch, [run 34511818127](https://github.com/Crossmint/crossmint-sdk/actions/runs/34511818127) | `smoke-<run_id>` | `0xc093185B63160e38C9C27E791c7C8cb12Ba9cDe0` | **5 passed (3.1m)** |

Both ran within ~35 minutes of each other against the same staging backend. The only difference is which wallet the suite authenticates into.

**The failure is deterministic, not flake.** `main` has failed 6 consecutive times since Aug 31. Five of the six are identical: `Transaction did not start - transfer button is still enabled`, always at test 4 (`transfers funds`), always 1 failed / 1 did not run / 3 passed.

**Funding is not the explanation.** The control logged:

```
✅ Wallet already holds 19.72 USDXM, skipping faucet
```

and still failed. The old wallet is well funded and cannot transfer; the new wallet started at 0, took faucet money, and completed the transfer.

### What is not proven here

I did not observe the `SIGNER_LIMIT_EXCEEDED` response directly. Playwright does not forward the browser's network errors to stdout, so CI logs only carry the SDK's generic timeout message. The error code and the 136-signer count come from the investigation on #2049. Everything measured is consistent with the signer cap, but treat that specific attribution as inference.

albertoelias-crossmint (approved):

